PR #7, opened 2018-01-05
small beautifications

pannous:
I fully understand if you don't want these in the master branch.

aliases int, double, float, real, byte for i32, i64, ...

added 'and' operator for &&   
added 'or' operator for ||


src/program.ts
```diff
@@ -121,9 +121,12 @@ export class Program extends DiagnosticEmitter {
 
     this.types = new Map([
       ["i8", Type.i8],
+      ["byte", Type.i8],
       ["i16", Type.i16],
       ["i32", Type.i32],
+      ["int", Type.i32],
       ["i64", Type.i64],
+      ["double", Type.i64],
```
dcodeIO: in my understanding, a `double` always is a 64-bit float. isn't it?

src/program.ts
```diff
@@ -121,9 +121,12 @@ export class Program extends DiagnosticEmitter {
 
     this.types = new Map([
       ["i8", Type.i8],
+      ["byte", Type.i8],
```
dcodeIO: Actually, aliases like these were in the prototype compiler, along the lines of:

i8 = sbyte
u8 = byte
i16 = short
u16 = ushort
i32 = int
u32 = uint
f32 = float
f64 = double

That is, until this issue emerged: https://github.com/AssemblyScript/prototype/issues/104

src/tokenizer.ts
```diff
@@ -187,6 +189,7 @@ export namespace Token {
   export function fromKeyword(text: string): Token {
     switch (text) {
       case "abstract": return Token.ABSTRACT;
+      case "and": return Token.AND;
```
dcodeIO: Is it still TypeScript then?

src/program.ts
pannous: You're right, sorry for sloppy friday night commit ;)
Playing into the hands of AssemblyScript/prototype#104

src/tokenizer.ts
pannous: no, and it breaks a test

export function and(loLeft: u32 ...
   and,
   ~~~
 in i64-polyfill.ts(16,3)

Let me know if you ever consider creating a *new* language, I love the clarity of your code.

dcodeIO: The breakage is probably due to missing entries in `Token.isAlsoIdentifier`. Though, adding non-TS-compatible logical operators doesn't feel right to me.

dcodeIO:
I have thought about the proposed aliases a bit and am currently favoring to keep is simple without any aliases, at least on the compiler-side, for two reasons: 1) It keeps lookup maps small and efficient internally and 2) most devs will use the default types so AS types look familiar accross different projects, which might be a nice-to-have in open source projects so foreign sources are easier to understand.

Aside from that, it's also easy to add aliases on a per-project level if there's a need to:

```ts
type int = i32;
type uint = u32;
type float = f32;
type double = f64;
...
```

dcodeIO:
Closing this issue for now as it hasn't received any replies recently. Feel free to reopen it if necessary!

PR #28, opened 2018-02-13
Add the packaging metadata to build the assemblyscript snap

elopio:
This package will let you publish the latest assemblyscript in the Ubuntu store, and from there reach many users on all the supported Ubuntu versions, and Linux distributions. You just have to go to https://build.snapcraft.io and enable the automated continuous delivery.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,0 +1,18 @@
+name: assemblyscript
+version: git
+summary: A TypeScript to WebAssembly compiler.
+description: |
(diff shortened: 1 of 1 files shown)


elopio:
Ping, anybody around to review this?

dcodeIO:
Ok, thanks. I just turned this into strict confinement now because I couldn't spot anything special to configure on a first glimpse. Basically, the `asc` command just reads and writes from paths provided to it by the user, so nothing magic.

PR #39, opened 2018-03-11
Implement function types / indirect calls / trampolines

dcodeIO:


dcodeIO:
Still missing: Trampolines for indirect function calls. These are a bit tricky because the *actual*  function being called is not known at compile time and initializers or even the number of maximum arguments might vary between signature-compatible functions.

One straight-forward way to implement these could be to always call the trampoline instead of the function with one additional argument indicating the actual number of provided arguments, zeroing out other arguments. TBD.

dcodeIO:
With the generalized approach, one way to implement it for indirect calls would be to always tablify and call the trampoline there, with the limitation that compatible signatures must always have the same number of total arguments, no matter whether optional or not, which naturally comes from sharing a WASM function type anyway.

PR #41, opened 2018-03-16
array.includes method

PinkaminaDianePie:
Implemented `array.includes` method, improved spec compatibility for `array.indexOf`, added few missing tests.

`array.includes` was created as a new method, without using `indexOf` just because by spec they work slightly differently, especially for floats, so they can became different in the future.

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -5,6 +5,7 @@ under the licensing terms detailed in LICENSE.
 
 * Daniel Wirtz <dcode@dcode.io>
 * Max Graey <maxgraey@gmail.com>
+* Igor Sbitnev <PinkaminaDianePie@gmail.com>
 
 ================================================================================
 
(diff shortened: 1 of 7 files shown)


dcodeIO:
Looking good! Can you remove the dist files? Perhaps I should put that into a CONTRIBUTORS.md file eventually :)

PinkaminaDianePie:
Sure, i'll do it tonight. It should be in `CONTRIBUTORS.md` - i specially checked if `dist` is present in `.gitignore` and if you keep it in repository, so i commited it only after ensured that you keep them.

dcodeIO:
There are some initial [contributing guidelines](https://github.com/AssemblyScript/assemblyscript/blob/master/CONTRIBUTING.md) now and I tweaked the git files a bit to exclude the dist files by default (must be force-added now).

PinkaminaDianePie:
removed dist and rebased

dcodeIO:
This is fine to merge. Just please also add yourself to the [NOTICE](https://github.com/AssemblyScript/assemblyscript/blob/master/NOTICE) file because, you know, bureaucracy.

PinkaminaDianePie:
done

dcodeIO:
Thank you! :)

PR #42, opened 2018-03-18
fix --lib arg

FloorLamp:


dcodeIO:
Thanks, looking good except for some stylistic nits, that is surrounding code uses double quotes and semicolons. Please also add yourself to the [NOTICE](https://github.com/AssemblyScript/assemblyscript/blob/master/NOTICE#L9) file before I merge :)

dcodeIO:
Thank you! :)

PR #47, opened 2018-03-19
add convenience function to compile a source string

FloorLamp:


dcodeIO:
Just noticed that builtins need to be defined here. There might be [more](https://github.com/AssemblyScript/assemblyscript/blob/master/std/assembly/builtins.ts) a user expects here because there is no other way to get their functionality, but including them all doesn't feel quite right to me.

So I'm thinking: Could this also become a utility function exported by asc?

FloorLamp:
Makes sense, updated

FloorLamp:
Changed `compileString` to simply be a wrapper over `main`

dcodeIO:
> This branch has conflicts that must be resolved

These issues should go away when removing the dist files from the PR. These shouldn't be included ideally, and CI doesn't need them anyway :)

dcodeIO:
Thanks! :)

PR #48, opened 2018-03-19
minimal 'hello world' example

pannous:
If you allow I will also simplify wiki.

https://github.com/AssemblyScript/assemblyscript/wiki/Hello-World


examples/hello-assembly/test.sh
```diff
@@ -0,0 +1,9 @@
+asc -v || npm install --save-dev AssemblyScript/assemblyscript
+
+asc hello-assembly.ts -b test.wasm -t test.wast
+
+node -i -e "log_char = c => process.stdout.write(String.fromCodePoint(c));\
+  new WebAssembly.Instance(new WebAssembly.Module(require('fs').readFileSync('test.wasm')),{console:{log_char}})"
```
xtuc: While I understand this, it doesn't look simple. Could you please split it in multiple lines with a few comments?

dcodeIO:
What do you guys think of moving examples towards WebAssembly Studio fiddles on the long run? For example, here's one for the [Game of Life](https://github.com/AssemblyScript/assemblyscript/issues/37#issuecomment-371297351) example.

pannous:
while https://webassembly.studio will be nice for large projects, to get started a leaner IDE like
https://maxgraey.github.io/Assembleash/#AssemblyScript or
https://wasdk.github.io/WasmFiddle/?ttxwx
seems better suited for a hello world.

We can link to both Assembleash and webassembly.studio *and* keep the examples folder for people who prefer their terminal.

pannous:
@xtuc agreed, I over-optimized for brevity, not readability. fixed

xtuc:
> What do you guys think of moving examples towards WebAssembly Studio fiddles on the long run?

Yes that's a good idea, people would get a feeling by playing with it.

pannous:
WasmFiddle for AssemblyScript would be ideal IDE for starters IMHO

dcodeIO:
> WasmFiddle for AssemblyScript would be ideal IDE for starters IMHO

Maybe we should provide something like this then on GitHub pages? There used to be [this one](http://assemblyscript.org/try/) ([source](https://github.com/AssemblyScript/website/tree/master/try)) for the prototype for example.

examples/hello-string/wasmx
```diff
@@ -0,0 +1,110 @@
+#!/usr/bin/env node
```
xtuc: What is this file? Did you write it?

pannous: yes, in-lining the node code as in hello-world.sh would be a bit too much with all the imports

pannous: I use it for emcc generated code as well, thus the demangle and strange imports

pannous: makes no sense here, so I'll remove the clutter

pannous:
Prettier now. do we have custom settings are the default settings fine?

examples/hello-assembly/test.sh
```diff
@@ -0,0 +1,15 @@
...
+
+# compile Typescript to native WebAssembly 
+asc hello-assembly.ts -b test.wasm -t test.wast
+
+# load and run the binary in node.js
+node -i -e "\
+	binary = require('fs').readFileSync('test.wasm');\
+	module = new WebAssembly.Module(binary);\
+	imports={console:{log_char: c => process.stdout.write(String.fromCodePoint(c))}};\
+  instance= new WebAssembly.Instance(module,imports)"
```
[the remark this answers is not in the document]
pannous: removed hello-assembly altogether. two examples are enough

dcodeIO:
Sorry for coming back to this so late, but I am not quite content with the bash files necessary to run this. In general I think that the examples we have at this point have a more portable view on things and that the "Empty AssemblyScript project" template in WebAssembly Studio / the wiki pages might already cover a minimal setup sufficiently. Closing for now, but thanks for your efforts!

PR #51, opened 2018-03-21
Add a build check for PRs

dcodeIO:
Just experimenting with adding a stage that runs for PRs only. Could then check if dist files are present, for example.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,16 @@
 language: node_js
+notifications:
+  email: false
+stages:
+  - name: check-pr
+    if: type = pull_request
 jobs:
   include:
+
+  - stage: check-pr
+    node_js: lts/*
+    script: ./scripts/check-pr.sh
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -2,6 +2,15 @@ language: node_js
...
   include:
 
+  - stage: check-pr
+    if: type = pull_request
+    node_js: lts/*
+    script:
+    - echo $TRAVIS_BRANCH
+    - echo $TRAVIS_PULL_REQUEST_BRANCH
+    - git diff --name-only HEAD...$TRAVIS_BRANCH
+    env: Performs sanity checks before building pull requests
```
dcodeIO: Ok, doesn't work. Idea is that the stage runs for PRs only and that a git command is used to get a list of all files changed between the respective AssemblyScript base branch and the PR, whatever branch it comes from.

Someone around familiar with these kind of things? :)

PR #97, opened 2018-05-06
Implement optional trailing commas

alangpierce:
Trailing commas are allowed in these situations in JS/TS:
* Array literals
* Argument lists
* Parameter lists
* Enums
* Type parameter lists
* Type argument lists (Technically filed as https://github.com/Microsoft/TypeScript/issues/21984 )

They're also allowed in these cases not (yet) supported by AssemblyScript:
* Object literals
* Array destructures
* Object destructures

All of these cases had similar-looking code, which needed to be tweaked to
handle the possibility of a comma before the close-delimiter. Type arguments
were a little different because they take a backtracking approach.

I also fixed a missing case in the AST printer: `export function` wasn't being
printed right.

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -901,24 +901,23 @@ export class Parser extends DiagnosticEmitter {
     // at '<': TypeParameter (',' TypeParameter)* '>'
 
     var typeParameters = new Array<TypeParameterNode>();
-    if (!tn.skip(Token.GREATERTHAN)) {
-      while (!tn.skip(Token.GREATERTHAN)) {
-        let typeParameter = this.parseTypeParameter(tn);
-        if (!typeParameter) return null;
-        typeParameters.push(<TypeParameterNode>typeParameter);
-        if (!tn.skip(Token.COMMA)) {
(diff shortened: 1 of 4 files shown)


src/parser.ts
```diff
@@ -900,17 +902,21 @@ export class Parser extends DiagnosticEmitter {
 
     var typeParameters = new Array<TypeParameterNode>();
     if (!tn.skip(Token.GREATERTHAN)) {
-      do {
+      while (!tn.skip(Token.GREATERTHAN)) {
```
dcodeIO: Maybe you have an idea how the initial double-check for GREATERTHAN could be avoided? :)

alangpierce: I guess do/while handles that pretty nicely here. Will update.

alangpierce: Actually, I'll just do a length check for the "no type parameters" case so this looks like the other cases.

alangpierce: ✅

dcodeIO: Thanks, looking good :)

dcodeIO:
Thanks!

PR #107, opened 2018-05-12
Allow keywords in named import/export syntax

alangpierce:
Progress toward #98

See the imports and exports grammar from the spec:
https://www.ecma-international.org/ecma-262/8.0/index.html#sec-imports
https://www.ecma-international.org/ecma-262/8.0/index.html#sec-exports

In particular, ImportSpecifier and ExportSpecifier use IdentifierName to
identify the import/export name rather than the more restrictive ImportedBinding
(or just Identifier). That means any keyword is allowed.

There are (at least) three contexts to think about in JS identifier parsing:
* In most contexts, identifiers like variable names can be any name except a
  reserved word.
* Some keywords, like "as", are normally not keywords but have special
  keyword-like meaning in specific situations. Acorn and Babel call these
  "contextual keywords", and `tokenIsAlsoIdentifier` and `preferIdentifier` in
  AssemblyScript appear to handle that case.
* In some contexts, like object keys and named imports/exports, any name is
  allowed, even keywords.

This change adds that third case in order to handle imports/exports. I added a
`forceIdentifier` flag that tells the tokenizer that any identifier name should
be treated as a plain identifier, then used it in the import and export cases.

### DIFF
--- a/src/tokenizer.ts
+++ b/src/tokenizer.ts
@@ -528,7 +528,11 @@ export class Tokenizer extends DiagnosticEmitter {
     return this.token = this.unsafeNext(preferIdentifier);
   }
 
-  private unsafeNext(preferIdentifier: bool = false, maxTokenLength: i32 = i32.MAX_VALUE): Token {
+  private unsafeNext(
+    preferIdentifier: bool = false,
+    forceIdentifier: bool = false,
+    maxTokenLength: i32 = i32.MAX_VALUE
+  ): Token {
     var text = this.source.text;
     while (this.pos < this.end) {
       this.tokenPos = this.pos;
@@ -908,6 +912,7 @@ export class Tokenizer extends DiagnosticEmitter {
               let keywordToken = tokenFromKeyword(keywordText);
               if (
                 keywordToken != Token.INVALID &&
+                !forceIdentifier &&
                 !(preferIdentifier && tokenIsAlsoIdentifier(keywordToken))
               ) {
                 return keywordToken;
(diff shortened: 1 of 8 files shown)


dcodeIO:
Now that there are effectively three modes, what do you think of adding an enum that indicates how to handle identifiers? Something like

```ts
enum IdentifierHandling {
  DEFAULT,
  PREFER,
  ALWAYS
}
```

alangpierce:
Sure, sounds reasonable to me. (It wouldn't surprise me if it got more complicated than that in more obscure cases, but the three modes seems fine for now.)

src/tokenizer.ts
```diff
@@ -982,6 +998,26 @@ export class Tokenizer extends DiagnosticEmitter {
     }
   }
 
+  /**
+   * Skip any name token, whether or not it is a keyword.
+   */
+  skipIdentifierName(): bool {
```
dcodeIO: Could this become an `IdentifierHandling` option on `skip()`, maybe? :)

alangpierce: Make sense, yeah this copied function is a bit ugly. I guess it felt a bit weird given the special case in `skip`.

How about this: I get rid of the special case in `skip` and specify it using an optional param like you suggested. Then, to make the common case the most natural to write, I'll make `skipIdentifier` as a shorthand for `skip(Token.IDENTIFIER, IdentifierHandling.PREFER)` and keep `skipIdentifierName` as a shorthand for `skip(Token.IDENTIFIER, IdentifierHandling.ALWAYS)`. Let me know what you think.

dcodeIO:
Great, thank you! :)

PR #108, opened 2018-05-13
Add a more helpful error message when you haven't defined an allocator

alangpierce:
This tripped me up, and it looks like others have run into it as well, so it
seems like a good idea to have a custom error message.

I also added a system for writing tests that assert that certain error codes are
triggered so that I could test this.

### DIFF
--- a/src/builtins.ts
+++ b/src/builtins.ts
@@ -2718,7 +2718,11 @@ export function compileAllocate(
   var allocatePrototype = program.elementsLookup.get(allocateInternalName);
   if (!allocatePrototype) {
     program.error(
-      DiagnosticCode.Cannot_allocate_memory_unless_an_allocator_is_defined_Please_define_an_allocator_e_g_by_importing_allocator_arena_or_allocator_tlsf,
+      DiagnosticCode.Cannot_find_name_0,
+      reportNode.range, allocateInternalName
+    );
+    program.info(
+      DiagnosticCode.An_allocator_must_be_declared_to_allocate_memory_Try_importing_allocator_arena_or_allocator_tlsf,
       reportNode.range
     );
     return module.createUnreachable();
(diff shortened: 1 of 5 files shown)


src/builtins.ts
```diff
@@ -2718,8 +2718,8 @@ export function compileAllocate(
   var allocatePrototype = program.elementsLookup.get(allocateInternalName);
   if (!allocatePrototype) {
     program.error(
-      DiagnosticCode.Cannot_find_name_0,
-      reportNode.range, allocateInternalName
+      DiagnosticCode.Cannot_allocate_memory_unless_an_allocator_is_defined_Please_define_an_allocator_e_g_by_importing_allocator_arena_or_allocator_tlsf,
```
dcodeIO: What do you think of retaining the original error message for compatibility with tooling, but additionally emitting an info message with more information?

alangpierce: Makes sense. I guess it feels a bit weird to say that `allocate_memory` isn't defined when it's never explicitly referenced, and maybe this should be seen as fundamentally different from a "cannot find name" error (especially since it's already a special case in the compiler). But it looks like TypeScript gives a regular `'React' is not defined` error when using JSX without having imported React, which is pretty much the same situation, so interpreting it as the more general error makes sense as long as it's not confusing.

alangpierce:
Thanks for the thoughts, I'll update when I get a chance, and also make the test stuff a little smarter to detect the new message. I'm thinking we can parse the test code for all lines beginning with `// Expect error:`, and the trimmed string after the colon must be a substring of stderr. (No such line means it's expected to succeed, multiple such lines means that all must be a substring of stderr.) Does that sound good? Seems like there are a few ways of doing it.

alangpierce:
@dcodeIO ok, I rebased and changed it to keep the original message as-is, and made a new info message that explains this case.

For tests, you can write any number of lines starting with `// Expect error:`, and each one must appear in the error output.

dcodeIO:
Thank you! :)

PR #109, opened 2018-05-13
Added a fiddle link to the mandlebrot example

MidasLamb:
Just used the same structure as the Conways fiddle.


dcodeIO:
Thanks, I took the liberty [to fork](https://webassembly.studio/?f=pne1undq5g) your fiddle and remove some of the boilerplate from the original example that isn't strictly needed :)

PR #114, opened 2018-05-20
Allow trailing commas in import and export statements

alangpierce:
This uses the same strategy as #97 for two more cases in the parser.


dcodeIO:
Thanks!

PR #118, opened 2018-05-23
Add badge to display install size

styfle:
This adds a badge to the `README.md` that displays the npm install size


styfle:
I'm sorry about that, the server went down 😨 
Looks like we're getting a lot of traffic 📈 

It should work now 👍

dcodeIO:
Thanks, but after leaving this open for so long I finally came to the conclusion that install size is probably not relevant enough to have it on the README, sorry.

PR #157, opened 2018-07-04
fix: added missing dependency

ColinEberhardt:
It looks like it was removed in this PR https://github.com/AssemblyScript/assemblyscript/commit/5ca5df3dc7d17c23ae59f2323b415c598d7d69a0 - but it is still needed at runtime.

Found while testing https://github.com/ColinEberhardt/wasm-mandelbrot/pull/3

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,7 @@
   "dependencies": {
     "@protobufjs/utf8": "^1.1.0",
     "binaryen": "48.0.0-nightly.20180627",
+    "glob": "^7.1.2",
     "long": "^4.0.0"
   },
   "devDependencies": {


dcodeIO:
Oh, yeah, I missed one occurence of `require("glob")` in `listFilesNode` in `cli/asc.js`. Does it work if you replace that function with the following?

```js
  function listFilesNode(dirname) {
    var files;
    try {
      stats.readTime += measure(() => {
        files = fs.readdirSync(dirname).filter(file => /^(?!.*\.d\.ts$).*\.ts$/.test(file));
      });
      return files;
    } catch (e) {
      return [];
    }
  }
```

dcodeIO:
Thanks for reporting this! Should be fixed by the commit above that makes sure `glob` isn't required in production.

PR #232, opened 2018-09-02
Remove unnecessary 'queuedImports.splice()' calls

andy-hanson:
I noticed that `queuedImports` is never looked at again after this loop. Since `queuedImports.splice(i, 1)` is equivalent to `i++` if we don't care about the state of `queuedImports` after the loop, we can just convert this to an ordinary `for-of` loop.
As of this PR, `queuedImports` is only referenced twice: once to write to in `initializeImports`, and once to loop over here.

### DIFF
--- a/src/program.ts
+++ b/src/program.ts
@@ -490,18 +490,15 @@ export class Program extends DiagnosticEmitter {
     }
 
     // queued imports should be resolvable now through traversing exports and queued exports
-    for (let i = 0; i < queuedImports.length;) {
-      let queuedImport = queuedImports[i];
+    for (const queuedImport of queuedImports) {
       let declaration = queuedImport.declaration;
       if (declaration) { // named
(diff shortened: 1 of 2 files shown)


src/program.ts
```diff
@@ -490,40 +490,33 @@ export class Program extends DiagnosticEmitter {
     }
 
     // queued imports should be resolvable now through traversing exports and queued exports
-    for (let i = 0; i < queuedImports.length;) {
-      let queuedImport = queuedImports[i];
+    for (const queuedImport of queuedImports) {
```
dcodeIO: Looks good, but can we keep the plain for-loop here? Just to be sure, because for-of isn't yet supported when compiling to WASM (somewhat due to the lack of iterators), even though we don't compile the compiler itself to WASM yet.

andy-hanson:
This PR subsumed by #249 anyway.

PR #244, opened 2018-09-06
ModuleExport#identifier never used, so just use Element directly

andy-hanson:
Noticed that this class was only used for a single member; so it makes more sense to just use that member directly and remove the class. (And it looks like that would have been the only place where Program exposed AST nodes.)


dcodeIO:
I believe the reasoning here is that the identifier for an export can differ from the identifier of the element, for example in:

```ts
import { A as B } from "some/file";
export { B };
```

which makes the export use identifier B instead of the element's declared identifier A. To me it seems that `.identifier` should actually be read somewhere, which might be a bug.

dcodeIO:
Superseded by https://github.com/AssemblyScript/assemblyscript/pull/489, which updates the representation of module exports as well.

PR #246, opened 2018-09-06
Add an InternalName type for clarity

andy-hanson:
Adds `type InternalName = string;` to be used to mark names that are not simple names.
Note that this leads to a funny type annotation `externalName: InternalName;` -- maybe that should be renamed to `internalNameOfExport`.

### DIFF
--- a/src/ast.ts
+++ b/src/ast.ts
@@ -1520,7 +1520,7 @@ export class Source extends Node {
   /** Normalized path. */
   normalizedPath: string;
   /** Path used internally. */
-  internalPath: string;
+  internalPath: InternalName;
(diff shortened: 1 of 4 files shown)


dcodeIO:
I see how this might be helpful, but at some point all these names should be replaced with proper symbols anyway (unfortunately this is unlikely to help with clarity) to reduce the amount of string comparisons necessary, i.e. performance. Hence, not sure.

andy-hanson:
While I have your attention, could you explain more about internal names? While working on #249 I noticed that exports are looked up in a big global map by using a name qualified by the Source path, but why not just have a separate map per Source?

Issue #251, opened 2018-09-09
Can't call method on newly created instance


andy-hanson:
```ts
import "allocator/tlsf";

class C {
  m(): i32 {
    return 0;
  }
}

export function test(): i32 {
  return new C().m();
}
```

Expected: No error.
Actual: `ERROR AS100: Operation not supported.`

PR #254, opened 2018-09-09
Fix parsing of `new C().m()`

andy-hanson:
FIxes parsing for #251. Previously this parsed as `new (C().m)()`, but it should parse as `(new C()).m()`.
Does not fix the issue because `resolveExpression` still doesn't support `NodeKind.NEW`.


src/parser.ts
```diff
@@ -3266,9 +3245,13 @@ export class Parser extends DiagnosticEmitter {
       ) {
         let args = this.parseArguments(tn);
         if (!args) return null;
-        expr = Node.createCallExpression(expr, typeArguments, args, tn.range(startPos, tn.pos)); // is again callable
+        expr = (isNew ? Node.createNewExpression : Node.createCallExpression)(
```
dcodeIO: This'll compile to an indirect call with two table elements in WASM. If possible, we should probably avoid that.

src/parser.ts
```diff
@@ -2913,32 +2913,9 @@ export class Parser extends DiagnosticEmitter {
...
-            DiagnosticCode.Operation_not_supported,
-            tn.range()
-          );
-        }
-        return null;
-      } else {
-        operand = this.parseExpression(tn, precedence);
-        if (!operand) return null;
-      }
+      const operand = this.parseExpression(tn, precedence);
```
dcodeIO: This will result in a warning when compiling to WASM, because `operand` isn't a compile-time constant. Certainly not really JS-y, but at this point `const` is a bit special in AS.

andy-hanson:
Looks like the needed change would be different now since the parser is a bit different.

PR #260, opened 2018-09-11
Support definite assignment assertion `x!: i32`

andy-hanson:
Supports TypeScript's definite assignment assertion syntax `x!: i32;`, used to disable `--strictPropertyInitialization` errors.
Presumably at some point in the future this should trigger a null check when accessing the field in debug mode. See #234.


src/ast.ts
```diff
@@ -828,6 +829,7 @@ export abstract class Node {
     stmt.range = range;
     stmt.flags = flags;
     stmt.name = name; name.parent = stmt;
+    stmt.definiteAssignmentAssertion = definiteAssignmentAssertion;
```
dcodeIO: This could also become a part of `flags`, maybe as `CommonFlags.DEFINITE_ASSIGNMENT`, similar to other modifiers. The advantage of a flag is that it doesn't occupy additional space in the class layout. Wdyt?

dcodeIO:
Thank you! :)

PR #267, opened 2018-09-17
Added support for TypedArray interop between WASM and JS contexts

kyr0:
As discussed in https://github.com/AssemblyScript/assemblyscript/issues/263 and https://github.com/AssemblyScript/assemblyscript/issues/240, I added support for memory mapped interop of all standard TypedArray subtypes. I added tests for every subtype to ensure a safe operation. Type checks ensure the safe construction of TypedArray subtypes in the loader lib.


msbasanth:
Superb, is this change is already in... Whether I will get TypedArray interop in latest of AssemblyScript?

It would be great if you could share the sample usage...

kyr0:
@msbasanth 

See here:

Example code (using current impl. by @dcodeIO):

https://github.com/kyr0/assemblyscript-js-wasm-interop-example

PR #272, opened 2018-09-19
Adding freeString and getFunction functions to the loader

scorsi:
Can now free a allocated string (with freeString instead of memory.free).
Can now retrieve a function by its pointer (safe version from 0 to 4 arguments, unsafe availble for JavaScript only), useful for callbacks.


lib/loader/index.js
```diff
@@ -74,6 +80,12 @@ function instantiate(module, imports) {
     return parts.join("") + String.fromCharCode.apply(String, U16.subarray(dataOffset, dataOffset + dataRemain));
   }
 
+  /** Frees a string in the module's memory by its pointer. */
+  function freeString(ptr) {
+    checkMem();
```
dcodeIO: Doesn't need to checkMem here. That's only necessary when accessing one of the views I32, F32 etc.

lib/loader/index.js
```diff
@@ -130,6 +142,41 @@ function instantiate(module, imports) {
     memory_free(ptr);
   }
 
+  /** Gets a function from the module's memory by its pointer, for JavaScript only. */
+  function getFunction(ptr) {
+    return table.get(ptr);
+  }
+
+  /** Gets a function with 0 parameters from the module's memory by its pointer. */
+  function getFunction0(ptr) {
```
dcodeIO: Are these overloads necessary? Users could as well do checks like these on their own, if they have a need for it.

lib/loader/index.js
scorsi: Yeah you're right. I'll change this

lib/loader/index.js
scorsi: I think it's pretty useful because users don't then need to understand how wasm memory really works and where to find the functions inside it.
And all the getFunctionX reduce the boilerplate.
It's easier to get a callback with theses functions.

What do you think ?

dcodeIO: `getFunction` is fine, yeah, but I am not so sure about `getFunctionX` which seems like something that is likely to end up unused most of the time by most developers anyway while adding to package size.

scorsi: You're right, I will check a clever solution for doing this instead of `getFunction0`, `getFunction1`... May be only one which does all case ? But what about the typescript type ? What do you think of `getFunction(ptr: number, nbArg: number): (...params: any[]) => any` and calling with undefined/null will don't do the arguments checking ?

dcodeIO:
`getFunction` and `newFunction` have recently landed in the loader. Differs a bit in that these two APIs are able to work together now, even though wrapping for variable arguments is performed.

Not sure about `freeString` - we might not need it as it's really just an alias - so considering this PR solved.

PR #273, opened 2018-09-19
Add lib/loader/ to package.json

scorsi:
To allow using the loader as follow:

```javascript
const loader = require('assemblyscript/lib/loader');
```


dcodeIO:
One day this will be its own package, but it also doesn't hurt to ship the respective guaranteed compatible version with the main package I think. Thanks!

PR #316, opened 2018-11-07
Add DataView to standard library

### LINKED ISSUE #315: [Feature] Add DataView implementation
Currently using the following small work around in order to port a script that uses `DataView`, but would be very nice with native support:

```ts
class DataView {
  source: Uint8Array
  buffer: ArrayBuffer
  byteOffset: u32
  byteLength: u32

  constructor (source: Uint8Array) {
    this.source = source
    this.buffer = source.buffer
    this.byteOffset = source.byteOffset
    this.byteLength = source.byteLength
  }

  getUint32 (byteOffset: u32, littleEndian: boolean): u32 {
    return (this.source[byteOffset] << 24) & (this.source[byteOffset + 1] << 16) & (this.source[byteOffset + 2] << 8) & (this.source[byteOffset + 3])
  }

  getUint16 (byteOffset: u32, littleEndian: boolean): u32 {
    return (this.source[byteOffset] << 8) & (this.source[byteOffset + 1])
  }
}
```

If someone could point me in the right direction I would be happy to try and implement this ☺️

LinusU:
Opening this for early feedback, as soon as I know I'm on the right track it should be very easy for me to add all the required methods.

Some questions:

1. Should I just `throw new RangeError('Out of bounds access')` when trying to access outside the bounds?
1. Is there any way to overload the `get*` functions for when `littleEndian` is known at compile time, or will the compiler just figure it out anyway?

    e.g. something like this:

    ```ts
    function getUint32(offset: number, littleEndian: false) { ... }
    function getUint32(offset: number, littleEndian: true) { ... }
    function getUint32(offset: number, littleEndian: boolean) { ... }
    ```
1. Do I have to register `DataView` as a global variable somewhere?
1. Where do I put the tests? I'm assuming somewhere under `tests/`, but then I'm a bit lost ☺️ 

Thank you for your time, looking forward to working a lot more on this project, it rocks! 🙌 

fixes #315 

FIXME:

- [x] Don't commit `dist/*` files
- [x] Add me to `NOTICE` file

### DIFF
--- a/std/assembly/dataview.ts
+++ b/std/assembly/dataview.ts
@@ -1,0 +1,129 @@
+import { HEADER_SIZE } from "./internal/arraybuffer";
+
+export class DataView {
+  constructor(
+    readonly buffer: ArrayBuffer,
+    readonly byteOffset: i32 = 0,
+    readonly byteLength: i32 = i32.MIN_VALUE,
+  ) {
+    if (byteLength === i32.MIN_VALUE) byteLength = buffer.byteLength - byteOffset;
+
+    if (byteOffset < 0) throw new RangeError("byteOffset cannot be negative");
+    if (byteLength < 0) throw new RangeError("byteLength cannot be negative");
+    if (byteOffset + byteLength > buffer.byteLength) throw new RangeError("Length out of range of buffer");
+  }
+
+  @inline
+  getFloat32(byteOffset: i32, littleEndian: boolean = false): f32 {
+    var result: u32 = load<u32>(changetype<usize>(this.buffer) + this.byteOffset + byteOffset, HEADER_SIZE);
+    return reinterpret<f32>(littleEndian ? result : bswap<u32>(result));
+  }
+
+  @inline
+  getFloat64(byteOffset: i32, littleEndian: boolean = false): f64 {
+    var result: u64 = load<u64>(changetype<usize>(this.buffer) + this.byteOffset + byteOffset, HEADER_SIZE);
+    return reinterpret<f64>(littleEndian ? result : bswap<u64>(result));
+  }
+
+  @inline
+  getInt8(byteOffset: i32): i8 {
+    return load<i8>(changetype<usize>(this.buffer) + this.byteOffset + byteOffset, HEADER_SIZE);
+  }
+
+  @inline
+  getInt16(byteOffset: i32, littleEndian: boolean = false): i16 {
+    var result: i16 = load<i16>(changetype<usize>(this.buffer) + this.byteOffset + byteOffset, HEADER_SIZE);
+    return littleEndian ? result : bswap<i16>(result);
+  }
+
+  @inline
+  getInt32(byteOffset: i32, littleEndian: boolean = false): i32 {
+    var result: i32 = load<i32>(changetype<usize>(this.buffer) + this.byteOffset + byteOffset, HEADER_SIZE);
+    return littleEndian ? result : bswap<i32>(result);
+  }
+
+  @inline
+  getUint8(byteOffset: i32): u8 {
+    return load<u8>(changetype<usize>(this.buffer) + this.byteOffset + byteOffset, HEADER_SIZE);
+  }
+
+  @inline
+  getUint16(byteOffset: i32, littleEndian: boolean = false): u16 {
+    var result: u16 = load<u16>(changetype<usize>(this.buffer) + this.byteOffset + byteOffset, HEADER_SIZE);
+    return littleEndian ? result : bswap<u16>(result);
+  }
+
+  @inline
+  getUint32(byteOffset: i32, littleEndian: boolean = false): u32 {
+    var result: u32 = load<u32>(changetype<usize>(this.buffer) + this.byteOffset + byteOffset, HEADER_SIZE);
+    return littleEndian ? result : bswap<u32>(result);
+  }
+
+  @inline
+  setFloat32(byteOffset: i32, value: f32, littleEndian: boolean = false): void {
+    var input: f32 = littleEndian ? value : reinterpret<f32>(bswap<u32>(reinterpret<u32>(value)));
+    store<f32>(changetype<usize>(this.buffer) + this.byteOffset + byteOffset, input, HEADER_SIZE);
+  }
+
+  @inline
+  setFloat64(byteOffset: i32, value: f64, littleEndian: boolean = false): void {
+    var input: f64 = littleEndian ? value : reinterpret<f64>(bswap<u64>(reinterpret<u64>(value)));
+    store<f64>(changetype<usize>(this.buffer) + this.byteOffset + byteOffset, input, HEADER_SIZE);
+  }
+
+  @inline
+  setInt8(byteOffset: i32, value: i8): void {
+    store<i8>(changetype<usize>(this.buffer) + this.byteOffset + byteOffset, value, HEADER_SIZE);
+  }
+
+  @inline
+  setInt16(byteOffset: i32, value: i16, littleEndian: boolean = false): void {
+    store<i16>(changetype<usize>(this.buffer) + this.byteOffset + byteOffset, littleEndian ? value : bswap<i16>(value), HEADER_SIZE);
+  }
+
+  @inline
+  setInt32(byteOffset: i32, value: i32, littleEndian: boolean = false): void {
+    store<i32>(changetype<usize>(this.buffer) + this.byteOffset + byteOffset, littleEndian ? value : bswap<i32>(value), HEADER_SIZE);
+  }
+
+  @inline
+  setUint8(byteOffset: i32, value: u8): void {
+    store<u8>(changetype<usize>(this.buffer) + this.byteOffset + byteOffset, value, HEADER_SIZE);
+  }
+
+  @inline
+  setUint16(byteOffset: i32, value: u16, littleEndian: boolean = false): void {
+    store<u16>(changetype<usize>(this.buffer) + this.byteOffset + byteOffset, littleEndian ? value : bswap<u16>(value), HEADER_SIZE);
+  }
+
+  @inline
+  setUint32(byteOffset: i32, value: u32, littleEndian: boolean = false): void {
+    store<u32>(changetype<usize>(this.buffer) + this.byteOffset + byteOffset, littleEndian ? value : bswap<u32>(value), HEADER_SIZE);
+  }
+
+  /**
+   * Non-standard additions that makes sense in WebAssembly, but won't work in JS
+   */
+
+  @inline
+  getInt64(byteOffset: i32, littleEndian: boolean = false): i64 {
+    var result: i64 = load<i64>(changetype<usize>(this.buffer) + this.byteOffset + byteOffset, HEADER_SIZE);
+    return littleEndian ? result : bswap<i64>(result);
+  }
+
+  @inline
+  getUint64(byteOffset: i32, littleEndian: boolean = false): u64 {
+    var result: u64 = load<u64>(changetype<usize>(this.buffer) + this.byteOffset + byteOffset, HEADER_SIZE);
+    return littleEndian ? result : bswap<u64>(result);
+  }
+
+  @inline
+  setInt64(byteOffset: i32, value: i64, littleEndian: boolean = false): void {
+    store<i64>(changetype<usize>(this.buffer) + this.byteOffset + byteOffset, littleEndian ? value : bswap<i64>(value), HEADER_SIZE);
+  }
+
+  @inline
+  setUint64(byteOffset: i32, value: u64, littleEndian: boolean = false): void {
+    store<u64>(changetype<usize>(this.buffer) + this.byteOffset + byteOffset, littleEndian ? value : bswap<u64>(value), HEADER_SIZE);
+  }
+}
--- a/std/assembly/index.d.ts
+++ b/std/assembly/index.d.ts
@@ -380,6 +380,58 @@ declare class ArrayBuffer {
   slice(begin?: i32, end?: i32): ArrayBuffer;
 }
 
+/** The `DataView` view provides a low-level interface for reading and writing multiple number types in a binary `ArrayBuffer`, without having to care about the platform's endianness. */
+declare class DataView {
+  /** The `buffer` accessor property represents the `ArrayBuffer` or `SharedArrayBuffer` referenced by the `DataView` at construction time. */
+  readonly buffer: ArrayBuffer;
+  /** The `byteLength` accessor property represents the length (in bytes) of this view from the start of its `ArrayBuffer` or `SharedArrayBuffer`. */
+  readonly byteLength: i32;
+  /** The `byteOffset` accessor property represents the offset (in bytes) of this view from the start of its `ArrayBuffer` or `SharedArrayBuffer`. */
+  readonly byteOffset: i32;
+  /** Constructs a new `DataView` with the given properties */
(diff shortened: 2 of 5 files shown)


std/assembly/dataview.ts
```diff
@@ -0,0 +1,49 @@
...
+export class DataView {
+  readonly buffer: ArrayBuffer;
+  readonly byteOffset: i32;
+  readonly byteLength: i32;
+
+  constructor (buffer: ArrayBuffer, byteOffset: i32, byteLength: i32) {
+    this.buffer = buffer
+    this.byteOffset = byteOffset
+    this.byteLength = byteLength
+  }
```
MaxGraey: You can improve constructor in this way: 
```ts
export class DataView { 
   constructor(
     public readonly buffer: ArrayBuffer,
     public readonly byteOffset: i32 = 0,
     public readonly byteLength: i32 = i32.MAX_VALUE
   ) {
     this.byteLength = min(this.byteLength, byteLength);
   }
}
```

std/assembly/dataview.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  @inline
+  getUint32 (byteOffset: u32, littleEndian: boolean): u32 {
+    if (littleEndian) {
+      return (
+        (loadUnsafeWithOffset<u8, u8>(this.buffer, byteOffset + 0, this.byteOffset) << 24) &
```
MaxGraey: Just use single call:
```ts
loadUnsafeWithOffset<u32,u32>(this.buffer, byteOffset);
```

std/assembly/dataview.ts
LinusU: Neat!

Would you mind explaining this part?

```ts
this.byteLength = min(this.byteLength, byteLength);
```

It seems like `this.byteLength` and `byteLength` refers to the same variable in that case? and thus it shouldn't really do anything?

Or did you mean `buffer.byteLength`?

```ts
this.byteLength = min(buffer.byteLength, byteLength);
```

std/assembly/dataview.ts
LinusU: Won't the index then be aligned and thus read the wrong value? and aren't all three parameters requried? 🤔 

```ts
@inline export function loadUnsafeWithOffset<T,V>(buffer: ArrayBuffer, index: i32, byteOffset: i32): V {
  return <V>load<T>(changetype<usize>(buffer) + <usize>byteOffset + (<usize>index << alignof<T>()), HEADER_SIZE);
}
```

MaxGraey: For big endian the same. And use `bswap` built-in for reverse endian order

std/assembly/dataview.ts
MaxGraey: Yeah, you are right, I mean `this.byteLength = min(buffer.byteLength, byteLength);`

std/assembly/dataview.ts
MaxGraey: > Won't the index then be aligned and thus possibly read the wrong value? 
Good catch! Need checking `<usize>buffer + byteOffset` alignment first and fallback to per-byte read with wrong alignment

dcodeIO: Is a per-byte access actually faster than simply doing an unaligned access? Might make sense to not use `loadUnsafeWithOffset` here but `load` directly if not.

LinusU: Should I use `bswap` when reading out individual bytes as well, or will it be faster to have two paths that reads the bytes differently?

MaxGraey: No. for per-byte access this not make sense. Just read in reverse order

dcodeIO: All loads in WASM are little endian. Probably just `if (!littleEndian) value = bswap(...)`.

LinusU: Hmm, actually, we should be able to just pass `0` for the `index` parameter, and control the offset with the `byteOffset` parameter?

```ts
  @inline
  getUint32 (byteOffset: u32, littleEndian: boolean): u32 {
    const result = loadUnsafeWithOffset<u32, u32>(this.buffer, 0, byteOffset);
    return littleEndian ? result : bswap<u32>(result);
  }

Comment (dcodeIO) [reply std/assembly/dataview.ts]: Just use `load<i32>(changetype<usize>(this.buffer) + byteOffset, HEADER_SIZE)`. No need for `loadUnsafe` here as we are not operating on indexes but on offsets.

**Comment by MaxGraey** [comment]
> Is there any way to overload the get* functions for when littleEndian is known at compile time, or will the compiler just figure it out anyway?

Don't worry about that. Binaryen has `dae` (dead argument elimination) and constant propagation if argument knows in compile time during optimization

**Comment by LinusU** [comment]
I will rebase and order/squash the commits in a good way when I'm done, just pushing now so that you can see the changes more easily 👍

File: std/assembly/dataview.ts [hunk 3]
```diff
@@ -0,0 +1,103 @@
...
+  constructor(
+    readonly buffer: ArrayBuffer,
+    readonly byteOffset: i32 = 0,
+    readonly byteLength: i32 = i32.MAX_VALUE,
+  ) {
+    this.byteLength = min(buffer.byteLength, byteLength);
+  }
+
+  @inline
+  getFloat32 (byteOffset: i32, littleEndian: boolean): f32 {
```
MaxGraey: In spec default value for `littleEndian` is `false`.

std/assembly/dataview.ts
```diff
@@ -0,0 +1,103 @@
...
+    set<u8>(this.buffer, byteOffset, value, littleEndian);
+  }
+
+  @inline
+  setUint16(byteOffset: i32, value: u16, littleEndian: boolean = false): void {
+    set<u16>(this.buffer, byteOffset, value, littleEndian);
+  }
+
+  @inline
+  setUint32(byteOffset: i32, value: u32, littleEndian: boolean = false): void {
```
MaxGraey: Also I think we could add `setUint64`/`setInt64` and `getUint64`/`getInt64` although this is not supported in JS.

LinusU: Sure thing 👍

LinusU: Hmm, that will not work when compiling out to JavaScript using `tsc` though? 🤔 

Maybe that's a problem elsewhere as well though? 🤔

MaxGraey: Currently this not possible but in future this changes. I think just need mentioned about using this methods in `portable.md` file

std/assembly/dataview.ts
```diff
@@ -0,0 +1,127 @@
+import { HEADER_SIZE } from "./internal/arraybuffer";
+
+@inline
+function get<T>(buffer: ArrayBuffer, byteOffset: i32, littleEndian: boolean): T {
+  const result = load<T>(changetype<usize>(buffer) + byteOffset, HEADER_SIZE);
```
MaxGraey: in AS better use `var` instead `const` if this variable can't compute in compile time or store in data section

LinusU: Good to know, will fix 👍 

Do you mind explaining why? ☺️ it is also better than `let`?

MaxGraey: It better ask to @dcodeIO

MaxGraey:
About alignment. wasm can read/store unalignment memory without caveat:
https://github.com/WebAssembly/design/issues/30#issuecomment-98871405

std/assembly/dataview.ts
```diff
@@ -0,0 +1,127 @@
...
+    readonly buffer: ArrayBuffer,
+    readonly byteOffset: i32 = 0,
+    readonly byteLength: i32 = i32.MAX_VALUE,
+  ) {
+    this.byteLength = min(buffer.byteLength, byteLength);
+  }
+
+  @inline
+  getFloat32(byteOffset: i32, littleEndian: boolean = false): f32 {
+    return get<f32>(this.buffer, byteOffset, littleEndian);
```
MaxGraey: May be 
```ts
get<f32>(this.buffer, this.byteOffset + byteOffset, littleEndian)
```
?

std/assembly/dataview.ts
dcodeIO: As a rule of thumb, `const` currently behaves pretty much like `static const` in other languages, indicating storage in static memory. I'm considering changing that, though.

std/assembly/dataview.ts
```diff
@@ -1,49 +1,57 @@
...
-        (loadUnsafeWithOffset<u8, u8>(this.buffer, byteOffset + 0, this.byteOffset) << 0) &
-        (loadUnsafeWithOffset<u8, u8>(this.buffer, byteOffset + 1, this.byteOffset) << 8)
-      );
-    }
+  constructor(
+    readonly buffer: ArrayBuffer,
+    readonly byteOffset: i32 = 0,
+    readonly byteLength: i32 = i32.MAX_VALUE,
+  ) {
+    this.byteLength = min(buffer.byteLength, byteLength);
```
dcodeIO: Should also take care of negative values I think. What happens n JS if a negative value is provided?

MaxGraey:
```ts
if (byteOffset < 0 || byteOffset > byteLength) {
  throw new Error('Start offset is outside the bounds of the buffer');
}
if (byteLength < 0) throw new Error('Invalid DataView length');

this.byteLength = min(buffer.byteLength, byteLength);
```

LinusU: I've added checks that works exactly as in the browser 👌 (5358cff)

std/assembly/dataview.ts
```diff
@@ -16,9 +16,11 @@ export class DataView {
   constructor(
     readonly buffer: ArrayBuffer,
     readonly byteOffset: i32 = 0,
-    readonly byteLength: i32 = i32.MAX_VALUE,
+    readonly byteLength: i32 = buffer.byteLength - byteOffset,
```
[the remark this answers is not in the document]
LinusU:
```diff
diff --git a/std/assembly/dataview.ts b/std/assembly/dataview.ts
index f1a8ebe..ef5df88 100644
--- a/std/assembly/dataview.ts
+++ b/std/assembly/dataview.ts
@@ -16,8 +16,10 @@ export class DataView {
   constructor(
     readonly buffer: ArrayBuffer,
     readonly byteOffset: i32 = 0,
-    readonly byteLength: i32 = buffer.byteLength - byteOffset,
+    readonly byteLength: i32 = i32.MIN_VALUE,
   ) {
+    if (byteLength === i32.MIN_VALUE) byteLength = buffer.byteLength - byteOffset
+
     if (byteOffset < 0) throw new RangeError("byteOffset cannot be negative");
     if (byteLength < 0) throw new RangeError("byteLength cannot be negative");
     if (byteOffset + byteLength > buffer.byteLength) throw new RangeError("Length out of range of buffer");
```

Is this an acceptable compromise?

It won't give a proper error if someone passes in `i32.MIN_VALUE` as the `byteLength`, but I think we have to choose one i32 value to mean "nothing was passed"

(btw. creating a `DataView` with a `byteLength` of `0` is totally valid)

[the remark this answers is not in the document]
MaxGraey: What about this?
```ts
byteLength = min(buffer.byteLength - byteOffset, byteLength);
```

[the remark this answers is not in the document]
LinusU: I was thinking that it would be preferable to throw if a too long byteLength was passed, rather than silently truncating the length of the `DataView`. That is how the spec says as well:

https://www.ecma-international.org/ecma-262/6.0/#sec-dataview-buffer-byteoffset-bytelength

LinusU:
Hmm, without fb34590 I'm getting the following error, would love some insight 🤔 

```text
ERROR AS200: Conversion from type 'i32' to 'u16' requires an explicit cast.

   return littleEndian ? result : bswap<T>(result);
                                           ~~~~~~
 in ~lib/dataview.ts(6,42)

ERROR AS200: Conversion from type 'i32' to 'u16' requires an explicit cast.

   return littleEndian ? result : bswap<T>(result);
          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 in ~lib/dataview.ts(6,9)
```

std/assembly/dataview.ts
```diff
@@ -0,0 +1,131 @@
+import { HEADER_SIZE } from "./internal/arraybuffer";
+
+@inline
+function get<T>(buffer: ArrayBuffer, byteOffset: i32, littleEndian: boolean): T {
+  var result: T = load<T>(changetype<usize>(buffer) + byteOffset, HEADER_SIZE);
+  return littleEndian ? result : bswap<T>(result);
```
MaxGraey: Also `swap` not handle float points, so you need do this:

```ts
if (isFloat<T>()) {
  if (sizeof<T>() == 4) {
    let result = <u32>load<u32>(changetype<usize>(buffer) + byteOffset, HEADER_SIZE);
    return reinterpret<T>(littleEndian ? result : bswap<u32>(result));
  } else {
    let result = <u64>load<u64>(changetype<usize>(buffer) + byteOffset, HEADER_SIZE);
    return reinterpret<T>(littleEndian ? result : bswap<u64>(result));
  }
} else {
  let result: T = load<T>(changetype<usize>(buffer) + byteOffset, HEADER_SIZE);
  return littleEndian ? result : bswap<T>(result);
}
```

And for `set<T>` need to do similar

MaxGraey:
Also plz add tests to new file `tests/compiler/std/dataview.ts`

std/assembly/index.d.ts
```diff
@@ -380,6 +380,50 @@ declare class ArrayBuffer {
...
+  /** The `byteOffset` accessor property represents the offset (in bytes) of this view from the start of its `ArrayBuffer` or `SharedArrayBuffer`. */
+  readonly byteOffset: i32;
+  /** Constructs a new `DataView` with the given properties */
+  constructor(buffer: ArrayBuffer, byteOffset?: i32, byteLength?: i32);
+  /** The `getFloat32()` method gets a signed 32-bit float (float) at the specified byte offset from the start of the `DataView`. */
+  getFloat32(byteOffset: i32, littleEndian?: boolean): f32
+  /** The `getFloat64()` method gets a signed 64-bit float (double) at the specified byte offset from the start of the `DataView`. */
+  getFloat64(byteOffset: i32, littleEndian?: boolean): f64
+  /** The `getInt8()` method gets a signed 8-bit integer (byte) at the specified byte offset from the start of the `DataView`. */
+  getInt8(byteOffset: i32, littleEndian?: boolean): i8
```
MaxGraey: Also don't forget remove `littleEndian?: boolean` for i8/u8

LinusU:
I think that everything should be good to go now 🙌 

I've rebased on master and squashed everything into one clean commit 😎

MaxGraey:
Also need declare `declare class DataView` inside `pordable/index.d.ts`. Just copy-paste this from `assembly/index.d.ts`

LinusU:
Added entry to `portable/index.d.ts` and rebased on master again 👍

LinusU:
Yeah, they are currently only used for getting and setting integers, since floats require special handling. I'd be happy to either 1) remove the functions and just have the code at each callsite, 2) update them to also work for floats, so that they are used all the time or 3) rename them `getInteger`/`setInteger` or similar. Just let me know ☺️

dcodeIO:
Regarding 2) I think that these could have been wrapped in a `reinterpret` instruction to make them work with no other changes, but I'd personally favor 1) to save the additional inlining step and resulting block in untouched output.

LinusU:
Done 🚀

dcodeIO:
Thank you! :)

PR #349, opened 2018-11-27
%TypedArray%.prototype.map

jtenner:
So far this is the best I have for implementing map. 

This is my first pull request and I hope to contribute much more if this turns out well.

### DIFF
--- a/src/compiler.ts
+++ b/src/compiler.ts
@@ -2560,7 +2560,10 @@ export class Compiler extends DiagnosticEmitter {
 
         // f32 to int
         if (fromType.kind == TypeKind.F32) {
-          if (toType.is(TypeFlags.SIGNED)) {
+          if (toType == Type.bool) {
+            expr = module.createBinary(BinaryOp.NeF32, expr, module.createF32(0));
+            wrapMode = WrapMode.NONE;
+          } else if (toType.is(TypeFlags.SIGNED)) {
             if (toType.is(TypeFlags.LONG)) {
               expr = module.createUnary(UnaryOp.TruncF32ToI64, expr);
             } else {
@@ -2576,7 +2579,10 @@ export class Compiler extends DiagnosticEmitter {
 
         // f64 to int
         } else {
-          if (toType.is(TypeFlags.SIGNED)) {
+          if (toType == Type.bool) {
+            expr = module.createBinary(BinaryOp.NeF64, expr, module.createF64(0));
+            wrapMode = WrapMode.NONE;
+          } else if (toType.is(TypeFlags.SIGNED)) {
             if (toType.is(TypeFlags.LONG)) {
               expr = module.createUnary(UnaryOp.TruncF64ToI64, expr);
             } else {
@@ -2643,7 +2649,10 @@ export class Compiler extends DiagnosticEmitter {
       if (fromType.is(TypeFlags.LONG)) {
 
         // i64 to i32 or smaller
-        if (!toType.is(TypeFlags.LONG)) {
+        if (toType == Type.bool) {
+          expr = module.createBinary(BinaryOp.NeI64, expr, module.createI64(0));
+          wrapMode = WrapMode.NONE;
+        } else if (!toType.is(TypeFlags.LONG)) {
           expr = module.createUnary(UnaryOp.WrapI64, expr); // discards upper bits
         }
 
--- a/tests/compiler/bool.optimized.wat
+++ b/tests/compiler/bool.optimized.wat
@@ -31,9 +31,8 @@
    unreachable
   end
   get_global $bool/I
-  i32.wrap/i64
-  i32.const 0
-  i32.ne
+  i64.const 0
+  i64.ne
   i32.const 1
   i32.ne
   if
@@ -58,9 +57,8 @@
    unreachable
   end
   get_global $bool/U
-  i32.wrap/i64
-  i32.const 0
-  i32.ne
+  i64.const 0
+  i64.ne
   i32.const 1
   i32.ne
   if
@@ -72,9 +70,8 @@
    unreachable
   end
   get_global $bool/f
-  i32.trunc_u/f32
-  i32.const 0
-  i32.ne
+  f32.const 0
+  f32.ne
   i32.const 1
   i32.ne
   if
@@ -86,9 +83,8 @@
    unreachable
   end
   get_global $bool/F
-  i32.trunc_u/f64
-  i32.const 0
-  i32.ne
+  f64.const 0
+  f64.ne
   i32.const 1
   i32.ne
   if
--- a/tests/compiler/bool.untouched.wat
+++ b/tests/compiler/bool.untouched.wat
@@ -33,9 +33,8 @@
    unreachable
   end
   get_global $bool/I
-  i32.wrap/i64
-  i32.const 0
-  i32.ne
+  i64.const 0
+  i64.ne
   i32.const 1
   i32.eq
   i32.eqz
@@ -62,9 +61,8 @@
    unreachable
   end
   get_global $bool/U
-  i32.wrap/i64
-  i32.const 0
-  i32.ne
+  i64.const 0
+  i64.ne
   i32.const 1
   i32.eq
   i32.eqz
@@ -77,9 +75,8 @@
    unreachable
   end
   get_global $bool/f
-  i32.trunc_u/f32
-  i32.const 0
-  i32.ne
+  f32.const 0
+  f32.ne
   i32.const 1
   i32.eq
   i32.eqz
@@ -92,9 +89,8 @@
    unreachable
   end
   get_global $bool/F
-  i32.trunc_u/f64
-  i32.const 0
-  i32.ne
+  f64.const 0
+  f64.ne
   i32.const 1
   i32.eq
   i32.eqz
--- a/tests/compiler/portable-conversions.optimized.wat
+++ b/tests/compiler/portable-conversions.optimized.wat
@@ -1,82 +1,620 @@
 (module
+ (type $iiiiv (func (param i32 i32 i32 i32)))
  (type $v (func))
- (memory $0 0)
+ (import "env" "abort" (func $~lib/env/abort (param i32 i32 i32 i32)))
+ (memory $0 1)
+ (data (i32.const 8) "\17\00\00\00p\00o\00r\00t\00a\00b\00l\00e\00-\00c\00o\00n\00v\00e\00r\00s\00i\00o\00n\00s\00.\00t\00s")
  (table $0 1 anyfunc)
  (elem (i32.const 0) $null)
- (global $portable-conversions/f (mut f32) (f32.const 0))
- (global $portable-conversions/F (mut f64) (f64.const 0))
+ (global $portable-conversions/i (mut i32) (i32.const 1))
+ (global $portable-conversions/I (mut i64) (i64.const 1))
+ (global $portable-conversions/f (mut f32) (f32.const 1))
+ (global $portable-conversions/F (mut f64) (f64.const 1))
  (export "memory" (memory $0))
  (export "table" (table $0))
  (start $start)
- (func $start (; 0 ;) (type $v)
+ (func $start (; 1 ;) (type $v)
+  get_global $portable-conversions/i
+  i32.const 255
+  i32.and
+  i32.eqz
+  if
+   i32.const 0
+   i32.const 8
+   i32.const 6
+   i32.const 0
+   call $~lib/env/abort
+   unreachable
+  end
+  get_global $portable-conversions/I
+  i32.wrap/i64
+  i32.const 255
+  i32.and
+  i32.eqz
+  if
+   i32.const 0
+   i32.const 8
+   i32.const 7
+   i32.const 0
+   call $~lib/env/abort
+   unreachable
+  end
   get_global $portable-conversions/f
   i32.trunc_s/f32
-  drop
+  i32.const 255
+  i32.and
+  i32.eqz
+  if
+   i32.const 0
+   i32.const 8
+   i32.const 8
+   i32.const 0
+   call $~lib/env/abort
+   unreachable
+  end
   get_global $portable-conversions/F
   i32.trunc_s/f64
-  drop
+  i32.const 255
+  i32.and
+  i32.eqz
+  if
+   i32.const 0
+   i32.const 8
+   i32.const 9
+   i32.const 0
+   call $~lib/env/abort
+   unreachable
+  end
+  get_global $portable-conversions/i
+  i32.const 65535
+  i32.and
+  i32.eqz
+  if
+   i32.const 0
+   i32.const 8
+   i32.const 11
+   i32.const 0
+   call $~lib/env/abort
+   unreachable
+  end
+  get_global $portable-conversions/I
+  i32.wrap/i64
+  i32.const 65535
+  i32.and
+  i32.eqz
+  if
+   i32.const 0
+   i32.const 8
+   i32.const 12
+   i32.const 0
+   call $~lib/env/abort
+   unreachable
+  end
   get_global $portable-conversions/f
   i32.trunc_s/f32
-  drop
+  i32.const 65535
+  i32.and
+  i32.eqz
+  if
+   i32.const 0
+   i32.const 8
+   i32.const 13
+   i32.const 0
+   call $~lib/env/abort
+   unreachable
+  end
   get_global $portable-conversions/F
   i32.trunc_s/f64
-  drop
+  i32.const 65535
+  i32.and
+  i32.eqz
+  if
+   i32.const 0
+   i32.const 8
+   i32.const 14
+   i32.const 0
+   call $~lib/env/abort
+   unreachable
+  end
+  get_global $portable-conversions/i
+  i32.eqz
+  if
+   i32.const 0
+   i32.const 8
+   i32.const 16
+   i32.const 0
+   call $~lib/env/abort
+   unreachable
+  end
+  get_global $portable-conversions/I
+  i32.wrap/i64
+  i32.eqz
+  if
+   i32.const 0
+   i32.const 8
+   i32.const 17
+   i32.const 0
+   call $~lib/env/abort
+   unreachable
+  end
   get_global $portable-conversions/f
   i32.trunc_s/f32
-  drop
+  i32.eqz
+  if
+   i32.const 0
+   i32.const 8
+   i32.const 18
+   i32.const 0
+   call $~lib/env/abort
+   unreachable
+  end
   get_global $portable-conversions/F
   i32.trunc_s/f64
-  drop
+  i32.eqz
+  if
+   i32.const 0
+   i32.const 8
+   i32.const 19
+   i32.const 0
+   call $~lib/env/abort
+   unreachable
+  end
+  get_global $portable-conversions/i
+  i64.extend_s/i32
+  i64.eqz
+  if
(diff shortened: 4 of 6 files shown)


std/assembly/internal/typedarray.ts
```diff
@@ -123,4 +123,29 @@ export abstract class TypedArray<T,V> {
       return this;
     }
   }
+
+  map<W extends TypedArray<T, V>>(callbackfn: (value: T, index: i32, array: W) => T): TypedArray<T, V> {
```
dcodeIO: Wondering about `W extends XY` here. This is not supported by the compiler yet (though it might be able to parse it). Does `TypedArray#map` even have/need it?

jtenner: @dcodeIO according to the specification map passes the typedarray as the third argument.

```ts
    var cast = changetype<W>(this);

    for (let index = 0; index < length; ++index) {
      var val = loadUnsafeWithOffset<T,T>(buffer, index, byteOffset);
      storeUnsafeWithOffset<T, T>(
        newbuff,
        index,
        // it must be "cast" here (and I wasn't sure what to do about it
        callbackfn(val, index, cast),
        byteOffset,
      );
    }
```

std/assembly/internal/typedarray.ts
```diff
@@ -123,4 +123,29 @@ export abstract class TypedArray<T,V> {
...
+    var buffer = this.buffer;
+    var length = this.length;
+    var byteLength = this.byteLength;
+    var byteOffset = this.byteOffset;
+    var newbuff = allocateUnsafe(byteLength);
```
MaxGraey: Why just not create `var result = new TypedArray<T,V>(this.length)`?

jtenner: It's in the definition of the class

```ts
export abstract class TypedArray<T,V>
```

MaxGraey: yeah, now I see

dcodeIO: `TypedArray` is abstract because you'd lose the `subarray` implementation on the concrete class. Though I agree that this should be easier, somehow, because as it stands it is missing GC registration.

jtenner: Oh wait! I have an idea.

dcodeIO: Maybe a built-in helper `instantiate<T>(...constructorArgs)` that is able to instantiate anything, given the class as a type argument?

jtenner: @dcodeIO I can implement the method on each subclass using the `new` keyword. Let me submit changes.

jtenner:
Okay so this totally works. It feels a bit hacky, but the code is very clear and does the job. Is there a way to make this method more generic?

std/assembly/typedarray.ts
```diff
@@ -16,6 +16,14 @@ export class Uint8Array extends TypedArray<u8,u32> {
   subarray(begin: i32 = 0, end: i32 = 0x7fffffff): Uint8Array {
     return changetype<Uint8Array>(super.subarray(begin, end));
   }
+  map(callbackfn: (value: u8, index: i32, array: Uint8Array) => u8): Uint8Array {
+    var length = this.length;
+    var result = new Uint8Array(this.length);
```
MaxGraey: Only one thing - you need also copy `byteOffset`

jtenner:
How can I copy byteOffset if the field is readonly?

MaxGraey:
@dcodeIO for better design I think make sense use getter for `byteOffset` instead `readonly`. wdyt?

jtenner:
And for testing, how should I verify the map method works? Is there a place to modify the test suite?

MaxGraey:
https://github.com/AssemblyScript/assemblyscript/blob/master/tests/compiler/std/typedarray.ts

jtenner:
One last question, I'm looking through the spec and can't find the reason why we must copy the byteOffset to a new TypedArray. Is there a reason I'm missing?

MaxGraey:
see how implement `__get` and `__set`

dcodeIO:
Not sure about setting `byteOffset´, as you are dealing with an entirely new array with a new backing buffer. Modifying byteOffset there might actually lead to out of bounds loads and stores.

MaxGraey:
@dcodeIO yeah, you are right

jtenner:
Yeah this pull request isn't gonna be easy.

jtenner:
Looks like something is broken. Can someone help me identify the problem?

std/assembly/typedarray.ts
```diff
@@ -16,14 +24,22 @@ export class Uint8Array extends TypedArray<u8,u32> {
...
+  }
 }
 
 export class Uint8ClampedArray extends TypedArray<u8,u32> {
   static readonly BYTES_PER_ELEMENT: usize = sizeof<u8>();
 
   @inline @operator("[]=")
-  protected __set(index: i32, value: i32): void {
-    super.__set(index, max(min(value, 255), 0));
+  protected __unchecked_set(index: i32, value: i32): void {
```
MaxGraey: Why? This dublicate

jtenner:
Now it looks like I have a problem with the function callbacks.

```log
 Error: Compile error
  ERROR AS200: Conversion from type '(val: i8, index: i32) => i8' to '(value: i8, index: i32, array: Int8Array) => i8' requires an explicit cast.

   assert(isInt8ArrayEqual(testMapArray1.map(squarei8), [1, 4, 9]));
                                             ~~~~~~~~
   in std/typedarray.ts(370,42)
```

I went to check out how `Array.prototype.map` performs the map operation and I see the callback parameter looks like this.

```ts
  map<U>(callbackfn: (value: T, index: i32, array: Array<T>) => U): Array<U> {
```

Am I missing something?

MaxGraey:
Yes, currently signature of callback should fully declare. Change:
```ts
function squarei8(val: i8, index: i32): i8 { ... }
```
to
```ts
function squarei8(val: i8, index: i32, array: i8[]): i8 { ... }
```

jtenner:
Okay. It looks like I've gotten it to the point where the tests *look* like they are supposed to look. There's still some issues though.

std/assembly/typedarray.ts
```diff
@@ -8,6 +8,14 @@ export class Int8Array extends TypedArray<i8,i32> {
   subarray(begin: i32 = 0, end: i32 = 0x7fffffff): Int8Array {
     return changetype<Int8Array>(super.subarray(begin, end));
   }
+  map(callbackfn: (value: i8, index: i32, array: Int8Array) => i8): Int8Array {
+    var length = this.length;
+    var result = new Int8Array(this.length);
+    for (let i = 0; i < length; ++i) {
+      result.__unchecked_set(i, callbackfn(this.__unchecked_get(i), i, this));
```
dcodeIO: This can also be written as `unchecked(result[i] = callbackfn(this.__unchecked_get(i), i, this))`, which will ensure that it still works whenever indexed access becomes more optimized or changes otherwise.

MaxGraey:
Also you should run `npm run test:compiler -- --create` for update test snapshots

std/assembly/typedarray.ts
jtenner: Thanks dcode, but when I try to use it, it errors about the type because it doesn't have an index attribute.

image

jtenner:
I somehow broke the `bool` test.

MaxGraey:
It seems you need use `npm run build` first because `build` dist files outdated currently. So run:
```
npm run build
```
and
```
npm run test:compiler -- --create
```
later.
And on't forgot exclude dist forlder from commit.
@dcodeIO could you update dists?

std/assembly/typedarray.ts
dcodeIO: Might be a TS error that results from the .d.ts not having it, but should compile with AS.

jtenner:
Okay. What do you mean by disclude dist folder? Should I delete it and commit with empty dist folder?

std/assembly/typedarray.ts
jtenner: I tried compiling with AS and it fails when trying to run the tests.

MaxGraey:
https://github.com/AssemblyScript/assemblyscript/blob/master/CONTRIBUTING.md#submitting-pull-requests

jtenner:
@MaxGraey  I'm confused by what this means. Should I leave the dist files as they are, from when I cloned the repo?

dcodeIO:
When working on the compiler, make sure to run `npm run clean` to remove the dist files before running the tests. That makes sure that the sources are being run. Then, when committing, regardless of whether you decided to run `npm run build` for yourself, make sure to not commit any changes to `dist/*`. CI checks that, hence this PR fails.

jtenner:
Okay. So all the tests pass. Pushing now. Please check my work, because it modified the dist folder.

dcodeIO:
Modifying the dist folder will make CI fail automatically, unfortunately. Let me suggest that you `git reset --soft HASH` to the last commit from the master repo right before your first commit, and make a new condensed one from your changes with changes to the dist files excluded followed by an `git push -f`.

dcodeIO:
What you'd usually do in such a scenario is, for example

```
$> git reset --soft 7596d732849c26d9ee80de2d295647c235fd5d4b
```

that keeps your local changes but resets the tree to the state before you made your first commit, followed by a

```
$> git add .
$> git reset dist/*
```

to add your changes back, except those to the dist files. Then creating a new first commit on your branch through

```
$> git commit -m "New commit message"
```

and force-pushing it to your branch

```
$> git push -f
```

Hope that helps :)

jtenner:
Yeah thanks for the tip. It all seems to work now. Should I change the documentation, add some comments? It didn't seem like I needed any since it was all very obvious what the code was doing.

std/assembly/typedarray.ts
```diff
@@ -16,14 +24,22 @@ export class Uint8Array extends TypedArray<u8,u32> {
...
+  }
 }
 
 export class Uint8ClampedArray extends TypedArray<u8,u32> {
   static readonly BYTES_PER_ELEMENT: usize = sizeof<u8>();
 
   @inline @operator("[]=")
   protected __set(index: i32, value: i32): void {
-    super.__set(index, max(min(value, 255), 0));
+    super.__unchecked_set(index, max(min(value, 255), 0));
```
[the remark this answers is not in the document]
jtenner: changing this function back to `super.__set` causes this error to occur when testing `compiler/std/typedarray.ts`

```
Fatal: Module::addFunction: $~lib/internal/typedarray/TypedArray<u8,u32>#__set already exists
```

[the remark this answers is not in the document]
dcodeIO: Try renaming the function itself to `__setClamped` as a workaround.

std/assembly/typedarray.ts
```diff
@@ -16,14 +24,22 @@ export class Uint8Array extends TypedArray<u8,u32> {
   subarray(begin: i32 = 0, end: i32 = 0x7fffffff): Uint8Array {
     return changetype<Uint8Array>(super.subarray(begin, end));
   }
+  map(callbackfn: (value: u8, index: i32, array: Uint8Array) => u8): Uint8Array {
+    var length = this.length;
+    var result = new Uint8Array(this.length);
```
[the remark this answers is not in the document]
jtenner: I've fixed this locally. Thanks for the tip.

std/assembly/typedarray.ts
dcodeIO: For reference, the low-level way to write this is:

```ts
import { HEADER_SIZE } from "./internal/arraybuffer";
```

```ts
  map(callbackfn: (value: i8, index: i32, array: Int8Array) => i8): Int8Array {
    var length = this.length;
    var result = new Int8Array(length);
    var src = changetype<usize>(this.buffer) + HEADER_SIZE + this.byteOffset;
    var dst = changetype<usize>(result.buffer) + HEADER_SIZE;
    for (let i = 0; i < length; ++i) {
      store<i8>(dst + (i << alignof<i8>()), callbackfn(load<i8>(src + (i << alignof<i8>())), i, this));
      // can actually skip alignof<T> for i8/u8 because it's 0 anyway, but added it for clarity
    }
    return result;
  }
```

std/assembly/typedarray.ts
```diff
@@ -34,6 +50,14 @@ export class Uint8ClampedArray extends TypedArray<u8,u32> {
   subarray(begin: i32 = 0, end: i32 = 0x7fffffff): Uint8ClampedArray {
     return changetype<Uint8ClampedArray>(super.subarray(begin, end));
   }
```
MaxGraey: plz use one line separator between methods

std/assembly/typedarray.ts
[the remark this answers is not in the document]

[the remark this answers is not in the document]
jtenner: I'm so confused. The test was perfectly fine before the change.

std/assembly/typedarray.ts
jtenner: Alright. Is this documented somewhere? tslint didn't pick it up.

std/assembly/typedarray.ts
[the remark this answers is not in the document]
MaxGraey: may be you forget replace `result.__set` to `result.__setClamped` for `map` inside `Uint8ClampedArray`?

std/assembly/typedarray.ts
MaxGraey: just missing rule. But just observe rest sources and you found this using everywhere

std/assembly/typedarray.ts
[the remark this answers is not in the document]
jtenner: As it turns out, I don't think that matters because the values will already be clamped, and changing the function names simply does not resolve the issue.

image

image

It is trying to add the `__set` method to the table when it already exists apparently. Could this be a compiler error?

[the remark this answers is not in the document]
jtenner: I can bring this chat to slack if you want, but I'll be around to work on this again tomorrow. Thank you @MaxGraey and @dcodeIO  for your amazing help.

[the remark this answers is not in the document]
dcodeIO: Yes, this is most likely a compiler error that involves `super`.

dcodeIO:
The "Operation not supported" issue on `unchecked(... this[i])` should be fixed by the above commit. Should now now be possible to use that syntax instead of `__unchecked_get`.

jtenner:
image

We still get a bug when I tried to switch over to normal array access involving tslint. There is a signature problem.

jtenner:
Actually I think the typescript compiler throws the error from within tslint. It says specifically:

```
ERROR: Unsafe use of expression of type 'any'. [no-unsafe-any]
```

This is a **TypeScript** error. Not a tslint one.

jtenner:
image

Of course adding this signature doesn't work, and causes the build to completely freak out with syntax errors. This is gonna be good.

dcodeIO:
Well, while I hoped we could have avoided it, it appears that using `__unchecked_get` etc. is indeed the only viable option at this point :)

jtenner:
We still need to figure out what the problem with is with the unclamped array problem anyway.

```
Testing compiler/std/typedarray.ts

Fatal: Module::addFunction: $~lib/internal/typedarray/TypedArray<u8,u32>#__set already exists
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! assemblyscript@0.5.0 test:compiler: `node tests/compiler`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the assemblyscript@0.5.0 test:compiler script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     C:\Users\jtenner\AppData\Roaming\npm-cache\_logs\2018-11-29T14_51_25_586Z-debug.log
npm ERR! Test failed.  See above for more details.
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! assemblyscript@0.5.0 make: `npm run clean && npm test && npm run build && npm test`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the assemblyscript@0.5.0 make script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     C:\Users\jtenner\AppData\Roaming\npm-cache\_logs\2018-11-29T14_51_25_658Z-debug.log
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! assemblyscript@0.5.0 all: `npm run check && npm run make`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the assemblyscript@0.5.0 all script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     C:\Users\jtenner\AppData\Roaming\npm-cache\_logs\2018-11-29T14_51_25_718Z-debug.log
```

std/assembly/typedarray.ts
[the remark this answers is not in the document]
jtenner: Where can I look to help?

dcodeIO:
I think the problem is that through calling `__set` manually, that function becomes compiled already, while when a normal array access like `something[i]` is used, that only sees that the full accessor hasn't been compiled yet and compiles both the getter and the setter, which then results in the error due to the duplicate setter. Should be easy to fix, going to take a look :)

jtenner:
@dcodeIO 

Can't I just make a generic `map` function like this in `TypedArray<T, V>`?

```ts
export abstract class  TypedArray<T, V> {
  map(callbackfn: (value: T, index: i32, array: Int16Array) => T): this {
    var length = this.length;
    var result = instantiate<this>(length);
    for (let i = 0; i < length; ++i) {
      result.__unchecked_set(i, callbackfn(this.__unchecked_get(i), i, this));
    }
    return result;
  }
}
```

dcodeIO:
Problem is that `this´ is `TypedArray<T,V>` there, which is abstract. Nonetheless, it should work because `abstract` isn't checked currently, but would require an overload on each concrete class `changetype`ing from `super.map`'s return type to the concrete type, similar to `subarray`.

dcodeIO:
One solution could be to check whether a function makes use of the `this` type and compiling a suitable version of this function for each subclass, like one would expect. That'd make overloading and changetypeing unnecessary, but not sure how much work this would be.

jtenner:
I just realized something. `__unchecked_set(index: i32, value: V)` accepts type `V`. How do I convert `T` to `V`? Also, is this the reason why it needs to recompile the `__set` function?

dcodeIO:
The idea behind the unconvenient `T` and `V` notation is that an Int8Array, for example, handles the `i8` type `T` while that's actually a native `i32` type `V`. Doing so allows us to omit masking/sign-extension at some occasions. For any signed small integer `T`, `V` is `i32`, for any unsigned small integer, `V` is `u32`. Hence, conversion works by `<V>valueOfT`.

jtenner:
```ts
// in internal/typedarray.ts
  map(callbackfn: (value: T, index: i32) => V): TypedArray<T, V> {
    var length = this.length;
    var result = instantiate<TypedArray<T, V>>(length);

    for (let i = 0; i < length; ++i) {
      result.__unchecked_set(
        i,
        callbackfn(this.__unchecked_get(i), i),
      );
    }
    return result;
  }

// for Int8Array
  map(callbackfn: (value: i8, index: i32, array: Int8Array) => i8): Int8Array {
    return <Int8Array>super.map((input: i8, idx: i32) => <i32>callbackfn(input, idx, this));
  }
```

PR #361, opened 2018-12-06
Implement %TypedArray%.prototype.map

jtenner:
Thanks for all the help so far!

It uses `this` as the return type on each `map` implementation, along with the `instantiate<ReturnType>(length)` macro to create the return value, and also an `unchecked(assignmentExpression)` macro because bounds checks aren't necessary. (TypedArray length is not variable.)

Please let me know if I didn't comply with the proper coding standards, or I need to update my implementation (because you've committed something before this pull request.)

I've also included a few test optimizations involving creating a few generic tests with typed parameters to  increase code clarity and reduce the surface area for bugs.

### DIFF
--- a/std/assembly/array.ts
+++ b/std/assembly/array.ts
@@ -74,7 +74,7 @@ export class Array<T> {
 
   every(callbackfn: (element: T, index: i32, array: Array<T>) => bool): bool {
     var buffer = this.buffer_;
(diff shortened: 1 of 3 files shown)


dcodeIO:
Sorry, broke this like completely in the last commit. Rebase should also use a macro then :)

PR #370, opened 2018-12-10
implement array#flat

jtenner:
Array#flat isn't type safe, and with what little knowledge I have, all I can do is verify that `this` is an array using the `isArray<T>()` macro.

The tests pass in `tests/compiler/std/array.ts` using the implemented `flat` function.

Is it possible to use a macro and verify an object's type at compile time?

Something like...

```ts
isType<this, Array<Array<U>>>();
```

### DIFF
--- a/std/assembly/array.ts
+++ b/std/assembly/array.ts
@@ -585,6 +585,42 @@ export class Array<T> {
     return this.join();
   }
 
+  flat<U>(): Array<U> {
+    if (!isArray<T>()) {
+      throw new Error("Cannot call flat on array whose children are not an array type.");
+    }
+    var count: i32 = 0;
+    var i: i32 = 0;
+    var j: i32;
+    var length: i32 = this.length_;
+    var sublength: i32;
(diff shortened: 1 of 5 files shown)


dcodeIO:
@MaxGraey wdyt?

MaxGraey:
Hard to tell. Spec require also param `depth` which default is 1.

jtenner:
I believe that there may be a better way to implement `Array<T>#flat`. It may take some time for me to fix up this pull request. Thank you for the feedback.

jtenner:
Hey @dcode now that isArray(child) doesn't need a type parameter, could this be possible on the master branch?

jtenner:
This is now possible via type propagation.

```ts
assert(isArray(unchecked(this[0])));
```

dcode:
not the @dcode that you're looking for

jtenner:
This should be closed in favor of the new pull request :)

MaxGraey:
You could close this by yourself)

PR #381, opened 2018-12-18
loader: check for null pointer in getArray() and getString()

jedisct1:
When `NULL` pointers are returned, the `loader`'s module `getArray()` and `getString()` return arbitrary data sitting at offset `0`.

Even though `null` shouldn't be returned for error handling, this can happen in real code.

Throwing an exception instead prevents undefined behaviors.

### DIFF
--- a/lib/loader/index.js
+++ b/lib/loader/index.js
@@ -80,6 +80,11 @@ function postInstantiate(baseModule, instance) {
   }
   checkMem();
 
+  /** Raises an exception if a pointer is the NULL pointer. */
+  function checkNull(ptr) {
+    if (!ptr) throw Error("null pointer deref");
+  }
+
   /** Allocates a new string in the module's memory and returns its pointer. */
   function newString(str) {
     var dataLength = str.length;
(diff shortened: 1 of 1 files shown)


dcodeIO:
I think that instead of throwing, these functions should just return `null` because the underlying string or array reference type might be nullable. Wdyt?

jedisct1:
Calling `getString()` on a `null` pointer, no matter what the underlying type is, should throw an exception IMHO. Can't think of use case where that wouldn't be due to a bug.

dcodeIO:
This has been *somewhat* implemented by means of an [id check](https://github.com/AssemblyScript/assemblyscript/blob/2b4dce0c09d3b47c31d87e3b3f67488fd0d5762c/lib/loader/index.js#L138) meanwhile. In general I think that we should strive to avoid potentially redundant checks like the one proposed here so we don't find us in a situation where these add up, and instead document the behavior more thoroughly, so users are aware that if they need to check, they must check.

Closing this PR as part of [2020 vacuum](https://github.com/AssemblyScript/assemblyscript/issues/1277) because the loader has changed substantially meanwhile, and creating a new PR documenting the behavior.

PR #419, opened 2019-01-14
Async main and compileString

willemneal:
This makes the main function asynchronous allowing imported readFile/writeFile to be asynchronous.  Also added a test for compileString.

I'm not sure if this is the correct way to go about it since it will be breaking for any code that calls main directly, but I thought I'd give it a try to see how it could be done.


dcodeIO:
I think that something like this is what we'd ultimately want, yeah. One thing to keep in mind is that compilation must still remain deterministic, i.e. even when (here: just mentioning) loading multiple files at once asynchronously, make sure that compilation (not necessarily parsing) order remains the same. Ofc we'll also have to make sure that nothing breaks (looking at you, WebAssembly Studio) when merging such changes.

vgrichina:
@MaxGraey I think this would be very nice to have, cause e.g. would help http://studio.nearprotocol.com/ to load modules from npm directly vs having to copy them into project, etc.

dcodeIO:
Closing this PR as part of [2020 vacuum](https://github.com/AssemblyScript/assemblyscript/issues/1277) as it seems to be outdated and makes quite a few unrelated changes that add 10k lines of code. I agree however that `main` should become properly asynchronous ultimately, and that we should look into it.

PR #424, opened 2019-01-16
[WIP] Switch to symbol tables

dcodeIO:
This PR investigates what's necessary to make the switch from plain strings to proper symbols, starting at parser level. Currently, this'll slow down compilation quite a bit because every access to `.text` on AST nodes obtains the respective region within the underlying source as a starting point, in turn leading to a sitation where both can coexist for a while in order to get it done.

### DIFF
--- a/src/ast.ts
+++ b/src/ast.ts
@@ -8,7 +8,8 @@ import {
   PATH_DELIMITER,
   STATIC_DELIMITER,
   INSTANCE_DELIMITER,
-  LIBRARY_PREFIX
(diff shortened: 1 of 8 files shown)


dcodeIO:
Figured that this might be easier to implement incrementally by first getting rid of the problematic data structures like `Program#elementsLookup`, and switch to symbols once everything's prepared.

dcodeIO:
Closing as this one has diverged from the preparations for symbol tables within master.

PR #437, opened 2019-01-25
[WIP] lib/host

willemneal:
This is my attempt at solving the issue of including a set of imported functions.  For now they are just printing functions.

This builds on top of the refactored typescript loader.  I had originally planned to use a namespace for host that could be passed as an import object, but ran into some issues.  Currently the host is a class, but the one caveat is that inside instance methods `this` is not bound.  This is fine since the methods don't rely on this to access the loaded Instance, rather it's a captured in the closure and updated when a bootstrap method.  I think in the future the loader should allow for hooks into the post initialization step so that details like this are hidden.

One other note is that I placed `.d.ts` file in the build directory, this when you load it you can get the type information of the exports of the module.  Something that had previously been a headache since they are available at runtime, but not compile time which means either adding method definitions by hand or using and `any` type, which should be avoided.

I'm not crazy about the design so far, but definitely feel that this is something is being reinvented by each new developer.  Another thought is that during compilation the required imports are bundled and placed in a custom section of the binary.


lib/host/tests/index.ts
```diff
@@ -0,0 +1,12 @@
+import {Host} from "../src";
+import * as loader from "../../loader/src";
+import * as fs from "fs";
+import ASModule from "./build";
+import * as asc from "../../../cli/asc"
+type testMod = typeof ASModule;
+
+
+var wasm = fs.readFileSync(`${__dirname}/build/untouched.wasm`);
+let imports = (new Host()).getImports<testMod>();
```
willemneal: Now the imports object contains a special property `post`, which is a list of identity functions that are repeatably applied to the instance by the loader.

dcodeIO:
Btw. there is [trace](https://github.com/AssemblyScript/assemblyscript/blob/master/lib/loader/index.js#L39) that functions as a low level logging utility. It takes a string as the first parameter, the number of arguments to log as the second, and a number of value arguments.

lib/host/tests/index.ts
willemneal: This was removed.

willemneal:
In the light of the new bindings #562 and fact that the contributions of this was more to additions in the loader, I'll just make a PR for that separately.

Furthermore this changes the names of exported wasm functions to include their return type so that they would be automatically wrapped.

PR #477, opened 2019-02-08
Refactor unsafe

jtenner:
@dcodeIO seemed to think that we could redesign how `reallocateUnsafe()` works for `ArrayBuffer`.  This pull request aims to do the following things:

- Standardize `reallocateUnsafe` and `allocateUnsafe` to avoid zeroing of data
- Add `reallocateZeroedUnsafe` and `allocateZeroedUnsafe` to zero out the data
- Replace instances of unsafe methods to `reallocateZeroedUnsafe` and `allocateZeroedUnsafe` respectively for backwards compatibility


jtenner:
It's become painfully obvious that this pull request is not going to solve the problem (if it even was a problem in the first place.)

Closing in favor of #535 obviously.

PR #494, opened 2019-02-17
Shared memory support added to compiler

nidin:
**CHANGES**

- `sharedMemory` option added to the compiler
Since shared memory needs mandatory max memory size, `sharedMemory` option accept a positive integer. 

| value  | meaning |
| ------------- | ------------- |
|`0` (default)| non shared memory|
|`+ve integer > 0` | enable shared memory and set max shared memory size|

- Feature is behind the flag `--enable threads`

### DIFF
--- a/cli/asc.js
+++ b/cli/asc.js
@@ -422,6 +422,7 @@ exports.main = function main(argv, options, callback) {
   assemblyscript.setNoTreeShaking(compilerOptions, args.noTreeShaking);
   assemblyscript.setNoAssert(compilerOptions, args.noAssert);
   assemblyscript.setImportMemory(compilerOptions, args.importMemory);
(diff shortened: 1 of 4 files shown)


dcodeIO:
LGTM. Almost certain that modules are going to explode due to too few memory picked ofc, but should be fine to iterate on that later.

One thing that comes to my mind here is that we should try to become as independent of linear memory as possible in the future, i.e. by eventually using the GC spec to its full extend by placing everything in GC managed containers (thinking about ArrayBuffer and whatnot), so memory can be used by the user exclusively. That's one of the reasons why implementing GC in linear memory now seems like a waste of time btw.

dcodeIO:
Thanks! :)

PR #504, opened 2019-02-21
Add isFunction and isNullable builtins

### LINKED ISSUE #502: [Enhancement] Support isFunction<T>()
I'm developing an assertion library, and it would be helpful to be able to assert if a type is a function.

A simple use case is the `toThrow()` expectation function.

```ts
  @inline
  public toThrow(message: string = ""): void {
    // assert(isFunction<T>(), "toThrow expectation must be called on a function type.");
    var throws: bool = !tryCall(this.actual!);
    reportActualString<string>(throws ? "throws" : "not throws");
    reportExpectedString("throws", this._not);
    assert(this._not ^ throws, message);
    clearExpected();
  }
```

dcodeIO:
This PR adds two new builtins to statically evaluate type conditions that would be otherwise impossible to check for.

fixes #502

### DIFF
--- a/src/builtins.ts
+++ b/src/builtins.ts
@@ -76,6 +76,8 @@ export namespace BuiltinSymbols {
   export const isReference = "~lib/builtins/isReference";
   export const isString = "~lib/builtins/isString";
   export const isArray = "~lib/builtins/isArray";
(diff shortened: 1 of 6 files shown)


jtenner:
I can't believe how small this pull request is.

PR #519, opened 2019-02-28
Enable source maps out of the box

dcodeIO:
Since the compiler relies on assertions heavily while it isn't that rare to hit one, and having to deal with issues regularly that show useless stacktraces referring to dist files only, I figured that it makes sense to ship source map support by default.

Afaik this still doesn't solve the same problem when dealing with stack traces from WebAssembly Studio, because source maps aren't picked up there for some reason. Any ideas why this isn't working?

### DIFF
--- a/bin/asc
+++ b/bin/asc
@@ -1,3 +1,4 @@
 #!/usr/bin/env node
+require("source-map-support").install();
 const asc = module.exports = require("../cli/asc.js");
 if (/\basc$/.test(process.argv[1])) process.exitCode = asc.main(process.argv.slice(2));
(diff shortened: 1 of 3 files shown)


vgrichina:
> Afaik this still doesn't solve the same problem when dealing with stack traces from WebAssembly Studio, because source maps aren't picked up there for some reason. Any ideas why this isn't working?

Likely because WebAssembly Studio hacks together solution to host website client-side (from RAM) inside of iframe, but the code which actually loads source map (dev tools) isn't hooked to load from RAM.

dcodeIO:
Now that you mention it, yeah, might be that this is a result of outputing the stack trace in a custom output window or something like that. Need to check if this also affects stack traces printed to browser console.

PR #525, opened 2019-03-02
General cleanup

dcodeIO:
This branch consolidates some general clean up that makes sense to do after a variety of other PRs have landed.


dcodeIO:
Ok, this reduces the overly large (because lazy) builtins file by round about 1600 lines. There might be more oportunities to shrink it, but I'm already hitting diminishing returns so I think it doesn't matter much.

PR #530, opened 2019-03-06
Implement TypedArray#forEach

jtenner:
- [x] add test for forEach
  - [x] test values against forEachValues array
  - [x] test index against call count
  - [x] test self parameter against cached pointer
  - [x] test final callback count
- [x] update d.ts file
- [x] implement macro FOREACH in `internal/typedarray.ts`
- [x] implement prototype method on each class in `typedarray.ts`


dcodeIO:
Thanks! :)

PR #532, opened 2019-03-08
TypedArray#reverse

jtenner:
- [x] add function to types file
- [x] create a test with generic type inputs
- [x] create a generic macro with type inputs
- [x] add the macro to each class that extends `TypedArray<T>`
- [x] account for byteOffset in ArrayBuffer
- [x] use `this` type where applicable
This should be all set for review.

### DIFF
--- a/std/assembly/internal/typedarray.ts
+++ b/std/assembly/internal/typedarray.ts
@@ -244,3 +244,14 @@ export function FOREACH<TArray extends TypedArray<T>, T>(
     callbackfn(LOAD<T>(buffer, i, byteOffset), i, array);
   }
 }
+
+@inline
+export function REVERSE<TArray extends TypedArray<T>, T>(array: TArray): TArray {
+  var buffer = array.buffer;
+  for (let front = 0, back = array.length - 1; front < back; ++front, --back) {
+    let temp = LOAD<T>(buffer, front);
+    STORE<T>(buffer, front, LOAD<T>(buffer, back));
+    STORE<T>(buffer, back, temp);
+  }
+  return array;
+}
--- a/std/assembly/internal/typedarray.ts
+++ b/std/assembly/internal/typedarray.ts
@@ -248,10 +248,12 @@ export function FOREACH<TArray extends TypedArray<T>, T>(
 @inline
 export function REVERSE<TArray extends TypedArray<T>, T>(array: TArray): TArray {
   var buffer = array.buffer;
+  var byteOffset = array.byteOffset;
+
   for (let front = 0, back = array.length - 1; front < back; ++front, --back) {
-    let temp = LOAD<T>(buffer, front);
-    STORE<T>(buffer, front, LOAD<T>(buffer, back));
-    STORE<T>(buffer, back, temp);
+    let temp = LOAD<T>(buffer, front, byteOffset);
+    STORE<T>(buffer, front, LOAD<T>(buffer, back, byteOffset), byteOffset);
+    STORE<T>(buffer, back, temp, byteOffset);
   }
   return array;
 }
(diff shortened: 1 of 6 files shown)


std/assembly/internal/typedarray.ts
```diff
@@ -244,3 +244,14 @@ export function FOREACH<TArray extends TypedArray<T>, T>(
...
   }
 }
+
+@inline
+export function REVERSE<TArray extends TypedArray<T>, T>(array: TArray): TArray {
+  var buffer = array.buffer;
+  for (let front = 0, back = array.length - 1; front < back; ++front, --back) {
+    let temp = LOAD<T>(buffer, front);
+    STORE<T>(buffer, front, LOAD<T>(buffer, back));
+    STORE<T>(buffer, back, temp);
```
MaxGraey: `array.byteOffset`?

jtenner: I did forget about that! Whoops.

tests/compiler/std/typedarray.ts
```diff
@@ -441,3 +441,33 @@ testArrayForEach<Int64Array, i64>();
...
+  for (i = 0; i < 9; i++) {
+    array[i] = <T>values[i];
+  }
+
+  array.reverse();
+
+  for (i = 0; i < 9; i++) {
+    assert(array[i] == <T>values[8 - i], "TypedArray reverse value mismatch");
+  }
+}
```
MaxGraey: plz add tests with offsets

jtenner: Please see the added test code. Thanks!

std/assembly/internal/typedarray.ts
jtenner: This should be resolved. I added the `byteOffset` parameter to each function call.

std/assembly/typedarray.ts
```diff
@@ -68,6 +69,10 @@ export class Int8Array extends TypedArray<i8> {
   forEach(callbackfn: (value: i8, index: i32, self: Int8Array) => void): void {
     FOREACH<Int8Array, i8>(this, callbackfn);
   }
+
+  reverse(): Int8Array {
```
MaxGraey: `reverse(): this`?

jtenner: I have avoided using `this` because I did not understand what the `this` type was for in AssemblyScript. I seem to remember some kind of issue with using it before.

MaxGraey: It mean same as in TS. It's just alias for instance's type.

Yeah. I would normally use the `this` type in TypeScript coding. For some reason, my tooling doesn't like this.

jtenner: Looks like it compiles though. I will start using this feature now.

MaxGraey:
```ts
reverse(): this {
    return REVERSE<this, i32>(this);
  }
```

jtenner: Please see the latest commit. This should be resolved.

MaxGraey: ok, great!

dcodeIO: Some context on `this` type: It's actually the exact same type as the class it is used in and doesn't automatially cast up to subclasses. So, using it in let's say `TypedArray<T>` would not work as expected, because a function inside of a subclass like `Uint8Array` would return `TypedArray<T>` still.

jtenner:
Alright. All outstanding problems listed so far are addressed.  Any other problems?

MaxGraey:
LGTM

jtenner:
Tomorrow I'll create another pull request to do a major cleanup of the previously written functions. A lot of the definitions could use the `this` type.

dcodeIO:
Thank you! :)

PR #535, opened 2019-03-08
Rework runtime

dcodeIO:
As mentioned a few times, I'm relatively unhappy with our internal runtime. More precisely, the current runtime stacked stuff on top of other stuff at random places in order to get things going, starting at memory allocation level, putting ArrayBuffers, Strings and other classes on top, introducing cumbersome macro helpers, and then there's GC.

So this PR aims at designing something better, a common runtime that can be used for all sorts of tasks, using a common header for all objects, providing common helpers for internal use, eliminating unnecessary helpers, simplifying buffer layout and so on.

See `std/assembly/runtime.ts` for the proposed implementation.

Differences are:

* All objects have a common hidden `HEADER` with their class id (for `instanceof` etc.) and their payload size (for easy realloc and length/byteLength computation).
* The `LOAD` and `STORE` macros are gone. Strings and ArrayBuffers don't have their own header anymore and start with their data immediately, making it possible to just `load` and `store` from/to them. Actually one can, theoretically, even cast between a String and an ArrayBuffer now.
* The `xyzUnsafe` helpers become mostly redundant and most of them will be removed (still need to think about copyUnsafe etc.).
* Objects can be `reallocate`ed always and `discard`ed as long as not `register`ed yet.
* Downside: Whenever an object is meant to become managed by GC, i.e. when returned from a standard library function, it must be `register`ed or it will leak. Upside: One can work with scratch objects and throw them away if necessary.
* The [GC interface](https://github.com/AssemblyScript/assemblyscript/tree/runtime/std/assembly/collector) has been extended to eventually support reference counting GCs as well
* The [MM interface](https://github.com/AssemblyScript/assemblyscript/tree/runtime/std/assembly/allocator) has been formalized
* Various issues in standard library code have been fixed in the process


std/assembly/string.ts
```diff
@@ -14,222 +13,248 @@ import {
...
-
-    var thisLen: isize = this.length;
-    var otherLen: isize = other.length;
-    var outLen: usize = thisLen + otherLen;
-    if (outLen == 0) return changetype<String>("");
-    var out = allocateUnsafe(outLen);
-    copyUnsafe(out, 0, this, 0, thisLen);
-    copyUnsafe(out, thisLen, other, 0, otherLen);
-    return out;
+    var thisSize: isize = this.length << 1;
```
MaxGraey: why `this.length << 1`?

dcodeIO:
The more I work on this I feel like redoing the entire standard library from scratch would be a better approach. Already stumbled over 20 places where stuff wasn't properly linked with GC and whatnot.

std/assembly/internal/sort.ts
```diff
@@ -42,7 +33,7 @@ export function COMPARATOR<T>(): (a: T, b: T) => i32 {
       if (!alen && !blen) return 0;
       if (!alen) return -1;
       if (!blen) return  1;
-      return compareUnsafe(<string>a, 0, <string>b, 0, <usize>min(alen, blen));
+      return String.cmp(<string>a, 0, <string>b, 0, <usize>min(alen, blen));
```
MaxGraey: Hmm, wdyt add `String.ord(a: string, b: string): i32`?

willemneal:
This seems like a good place to add interfaces/abstract methods.  Since we are adding new fields across the board, adding virtual table like data structure would be a good idea.  An interface could be a special class where each method is an anonymous function.  Then when a class implements an interface, a class variable holds the location in the function table of the first anonymous function.  Then references to interface methods become an indirect call using the pointer + method number.  This will also become trivial once there are function references.

dcodeIO:
@willemneal Laying the foundation for virtual lookup tables is one goal here, yeah, but don't hold your breath that it'll make it into this PR already :)

dcodeIO:
Still TODO:
- Checked builtin array set using a similar check like builtin array get
- Rewrite ITCM to the new GC interface and update the remaining gc tests
- Review entire stdlib and validate that runtime hooks are correct/not missing
- Update stdlib code from using shifted indexes (`dataStart + <usize>i << alignof``) to plain pointers (`dataStart + offset`) if number of instructions/loads/whatev can be reduced this way
- Think about `instanceof` (see https://github.com/AssemblyScript/assemblyscript/pull/535#issuecomment-473626389)

dcodeIO:
Turns out built-in array access suffers from the same issue as worked around in dataviews

```ts
import "allocator/arena";

var arr = new Array<i32>(2);
arr[i32.MIN_VALUE]; // not error
```

caused by shifting the index early in order to compare to byteLength. Unfortunate...

dcodeIO:
One solution could be to do only unchecked array access with a builtin and keep checked access as overloads. I think the final result would be about the same as only unchecked accesses can benefit from constant offset propagation anyway. Also, checked access is considerably larger code-wise and it makes sense to make a function call to the overload there and let the optimizer decide.

dcodeIO:
Another thing: I'm also a bit unhappy with `unchecked(...)` cluttering stdlib code. An alternative could be that any `@unsafe` function does unchecked accesses by default, and possibly also renaming `unchecked(...)` to `unsafe(...)` to unify the concept.

dcodeIO:
With all the assertions in place and running the dummy GC to check that hooks are present where I'd expect them, one obvious issue this triggers is that an array like `new Array<Ref>(10)` has a non-nullable `T` yet contains ten times `null` on instantiation. The obvious fix here is to not populate a new array like this with `null`s but with `new Ref()`s, which makes all of this type-safe, but might confuse JS devs. Also makes `new Array<Ref>(10)` inefficient if it's instantly populated with elements as all the temporary Refs remain unused and need to be collected. Plus, a type `Ref` used as a non-nullable array element type must have a default constructor and otherwise errors. Any thoughts?

jtenner:
I actually hit this problem when coding `as2d`, and it took a very long time (something like 2 hours) to figure out the array value was `null` because no `null` exception was thrown. 

One of the possible fixes might include changing the `index.d.ts` file to include a type signature that accurately represents reality.

https://github.com/AssemblyScript/assemblyscript/blob/master/std/assembly/index.d.ts#L1174

Should probably be `T | null` here. This would have signaled to me that the value could be `null`. Combined with postfix `!` assertions this should be just fine.

Using a default constructor like the example presented that accepts parameters could be devastatingly slow for reference types. Please do not do this.

dcodeIO:
Hmm, actually there are more issues like this down the road. like:

```ts
var arr: Ref[] = [];
arr.length = 2; // needs filling
```

dcodeIO:
Another idea might be to deviate from JS semantics and make the array constructor behave like the argument was more like a capacity hint for the initial allocation, but still have it be `.length = 0` afterwards. In such a case, nothing would have to be instantiated, and doing sequential assignment to the array would work as expected.

```ts
var arr = new Array<Ref>(10); // allocates a backing buffer to hold 10 elements
assert(arr.length == 0); // but length remains 0
arr[0] = ...; // sets and increases length = 1
...
```

Wouldn't cover something like this ofc, which is a lot more uncommon though:

```ts
var arr = new Array<Ref>(10);
arr.length = 5; // needs filling, could also just emit an error
```

jtenner:
Playing with people's expectations of how things should work is bad news.  We should admit that nothing we code into the std library is going to hold the developer's hand all the way through. It becomes a problem of balance. Every time you tighten the grip on solving the null array value problem, the more obscure the problems that slip through your fingers. There are no good solutions for this, otherwise, languages like `C#` would have solved it.

I think the reason why `AssemblyScript` will be successful is *because* it currently limits itself to conform to JavaScript as much as possible. This kind of limitation makes things *more* accessible. Not less.

I vote to make the smallest number of assumptions possible.

dcodeIO:
I get your point there, yeah. The safest bet with the least assumptions would be to simply disallow `new Array<Ref>(10)` in safe code, but allow it in `@unsafe` contexts - maybe.

MaxGraey:
But in js/ts `new Array<Ref>(10)` already misaligned with AS behaviour. In js this 10 `undefined` values, in AS this 10 `null` values, so 10 `Ref()` may be not is bad idea. Need to think over

dcodeIO:
```ts
// this errors, as it would create a holey array
var refArray = new Array<Ref>(10);

// this errors if length is less than 5 currently as it would create holes:
refArray.length = 5;

// if length is already 5, this is allowed as it doesn't create holes:
refArray[5] = new Ref(); // now length is 6

// a non-standard helper is present to give capacity hints, so the following works:
var refArray = Array.create<Ref>(10); // length is initialized with 0
refArray[0] = new Ref(); // now length is 1
refArray[1] = new Ref(); // now length is 2 etc.
```

std/assembly/dataview.ts
```diff
@@ -39,7 +42,7 @@ export class DataView {
     if (
       i32(byteOffset < 0) |
       i32(byteOffset + 4 > this.dataLength)
-    ) throw new Error("Offset out of bounds");
+    ) throw new RangeError(E_INDEXOUTOFRANGE);
```
MaxGraey: how about make special instance? and use it as `throw new IOOFError()`?

dcodeIO:
Another change I've made to be more type-safe is that non-null assertions now involve a runtime check. Compiling without assertions also disables the check.

std/assembly/dataview.ts
dcodeIO: Not sure about the benefits, as checking for `RangeError` usually suffices and is in line with the spec. Suggesting to think about this again once exceptions are actually supported :)

dcodeIO:
Hmm, I already had a bad feeling about the RETAIN/RELEASE macros and now I see why. On global assignment, a tracing GC can't do anything with them while a reference counting GC needs the ref only (there's actually no parent). On a first glimpse it looks like the APIs required for tracing/rc are too different to use a common abstraction. That's a bummer...

dcodeIO:
Complexity is steadily growing over my head, so instead of going overboard trying to implement a new GC, I think that I'll try to converge this PR to a baseline that can be used to do this in a follow-up. Regarding my last comment, I now made the parent on RETAIN/RELEASE optional if, and only if, a reference on a global (here: root) is retained/released. At least that looks more appealing to me than having two different retain/release mechanisms. My hope is that the unnecessary checks for null parents that a tracing GC would do will eliminate after optimizations, but even if that's not the case yet, it's just one if and optimization can be looked into later.

dcodeIO:
Just to point out the alternative: Do it like previously, with just the hooks to implement by a potential collector, and have something like this at the relevant parts in stdlib:

```ts
if (isManaged<T>()) {
  if (isDefined(__ref_link)) __ref_link(changetype<usize>(element), changetype<usize>(this));
  if (isDefined(__ref_retain)) __ref_retain(changetype<usize>(element));
}
```

That'd be less convenient and would make this more explicit and uglier, but the integration would be... well, more direct, straight-forward I guess.

dcodeIO:
This should pass all tests now, **but** GC tests are still incomplete and reference counting support is still missing handling of locals (or other zero-rc objects like in `new Ref().doSomething();`) due to WASM not providing access to the execution stack. Theoretically, a reference counting GC could keep a list of all registrations already and visit them, checking for rc=0, at specific points in time, like when the start function has finished executing or when returning control to the module after a function export was called.

* [Memory manager interface](https://github.com/AssemblyScript/assemblyscript/tree/runtime/std/assembly/allocator)
* [Garbage collector interface](https://github.com/AssemblyScript/assemblyscript/tree/runtime/std/assembly/collector)

singpolyma-shopify:
> All objects have a common hidden HEADER with their class id (for instanceof etc.) and their payload size (for easy realloc and length/byteLength computation).

Does this include user defined classes?  That would prove very problematic for our use case, where we currently pass struct-like data to assemblyscript and rust from outside, and so it is *very* helpful that user defined classes have the same layout as a struct.

dcodeIO:
Yes, this includes _managed_ user defined classes as well. Decorating a class as `@unmanaged` makes it behave like a struct, without a header, not garbage collected etc.

dcodeIO:
Hmm, this is new

```
threads [WebAssembly.Module(): Compiling wasm function #1:<?> failed: Atomics are allowed only in AsmJs modules @+718]
```

dcodeIO:
Fascinating. Seems only the push build uses a different version of node-v8, something that has simd but throws a pretty old error on threads. Is this Travis CI slowly dissolving?

dcodeIO:
With that fixed, I'm wondering it `__ref_unlink` could be removed. Iirc I added that to support both tracing and arc with the same hooks, but as that's not necessary anymore, it seems unnecessary now. Unfortunately I don't know enough GCs to be sure, though.

leodasvacas:
@dcodeIO At The Graph we write our own classes but also std classes such as `ArrayBuffer` directly to the linear memory and pass the pointer to AS. This has worked so far, but the introduction of GC and the effect of headers sound like it may interfere, and I can't mark the std classes as `@unmanaged`. Could there be a compiler cli flag to compile all classes as if they are `@unmanaged`?

dcodeIO:
Well, on managed objects, the header is necessary for language features like `instanceof` and virtual overloads, regardless of the presence of GC. Layout is really simple, though, and it's straight-forward to map with a C struct:

```
╒════════════════════ Managed header (32-bit) ══════════════════╕
   3                   2                   1
 1 0 9 8 7 6 5 4 3 2 1 0 9 8 7 6 5 4 3 2 1 0 9 8 7 6 5 4 3 2 1 0  bits
├─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┤
│   .classId         unique id of the respective class          │
├───────────────────────────────────────────────────────────────┤
│   .payloadLength   size of the object                         │
├───────────────────────────────────────────────────────────────┤
│   .reserved1       reserved space for GC                      │ 
├───────────────────────────────────────────────────────────────┤
│   .reserved2       reserved space for GC                      │
└───────────────────────────────────────────────────────────────┘ 
```

The reserved fields are present only if GC is, so that's 8 bytes without a GC and 16 bytes with a GC.

dcodeIO:
Last commit finalizes GC integration into the compiler. Similar to how it was before, when a tracing GC is present, the class id is equivalent to the iteration function (that calls a callback for each managed reference field and traverses. see also https://github.com/WebAssembly/binaryen/issues/1957). No GC or a reference counting GC doesn't need this and just increments a counter.

Also turned out that iterate functions can make use of the `call_indirect` builtin to avoid dealing with the `argc` helper, which lead me to a simple optimization: When `call_indirect` is called with a constant function index, the called function is actually known statically and a direct call can be emitted. This leads to iterate functions being mostly direct calls, except the first call when marking to enter the traversal and the call of the callback (itcm marks there). Actually, the callback is also known statically (it's always the same argument), so these calls could be converted to direct calls as well after argument reduction, but that's rather an optimization for Binaryen. TODO: Figure out if that already happens, otherwise propose.

What remains is handling potential edge cases and cleaning up :)

leodasvacas:
@dcodeIO I see that's very helpful, I can see it working. What determines the presence of GC? Is that done at compile time?

dcodeIO:
Yes, GC is included at compile time. Thinking about a suitable mechanism to set sensible defaults currently, for example include `allocator/tlsf` and `collector/itcm` by default, and add a new compiler option `--noRuntime` for the old behavior.

dcodeIO:
I am currently prototyping documentation / instructions for the new runtime functionality in the wiki:

* https://github.com/AssemblyScript/assemblyscript/wiki/Runtime

Please let me know what you'd like to be covered there :)

dcodeIO:
To elaborate on the possibility of reference counting a bit: If I'd want to implement this now, I'd most likely do something like this: Whenever a managed object is allocated, I'd create a new temporary local holding a reference to it (remember: we don't have sequential access to the execution stack) and initialize the reference with RC=1. For example

```ts
function myFunc(s: string): string {
  return s + "b";
}
```

turns into

```ts
function myFunc(s: string): string {
  var temp0: string:
  return (temp0 = register(s + "b")); // RC=1
}
```

so the intermediate reference of the string concatenation can be tracked to find out if it remained unused when the function exits. Unfortunately, this check cannot be performed without accounting for `return`s because the reference would have been collected before the returned element reaches the caller. Thus, whenever a reference is `return`ed from a function, its RC must be incremented by 1 before cleanup steps run. making sure it lives long enough to reach the caller. One thing to have in mind: Considering the design of the RC prototype,, this would result in the reference being considered a possible root of a reference cycle further down the road, which happens when a reference with RC>=2 is released.

Now, when accounting for returns, our function looks something like this:

```ts
function myFunc(s: string): string {
  var ret;
  l_ret: {
    let temp0: string;
    ret = retain(temp0 = register(s + "b")); // RC=1 on register, RC=2 on retain
    release(temp0); // RC=1, now considered a possible root of a cycle
    break l_ret;
  }
  return ret;
}
```

During compilation, the lifetime of variables would have to be tracked and whenever a variable is free'd (and holds a reference) its RC is decremented - essentially a concept of scopes with an outer `l_ret` block so `release` calls can be inserted before returning.

Makes me realize once more that the concepts of a tracing and a reference counting GC are different enough to require two distinct approaches to compiling them, which again lets me want to drop the other and just keep one. Considering that we don't even have a working ARC implementation yet, it seems risky to go this route because there could be new blockers along the road that I haven't thought of yet.

Writing this down because @MaxGraey is in favor of ARC and asked about it when noticing that this branch is still using tracing. What would you guys do?

MaxGraey:
@dcode How about use hybrid GC + [Region-based memory management](https://en.wikipedia.org/wiki/Region-based_memory_management)? This approach is alternative to stack frame mem managment. Also: http://www.cs.utexas.edu/users/speedway/DaCapo/papers/immix-pldi-2008.pdf

dcodeIO:
Hard to tell. This seems to add a third concept into the mix and I'm already overwhelmed by the other two to be honest. This whole PR feels like I'm in an infinite loop of trying one thing, throwing it away and starting all over again.

dcode:
@MaxGraey

MaxGraey:
I suggest leave most working (existing) solution for current PR. And start experimenting with alternatives in different branch or separate playground later. I worry current pure mark & sweep approach is not perfect especially for realtime applications due to long pausing time.

MaxGraey:
@dcode Sorry to bother you again. Next time just ignore this curiosity =)

dcode:
It's no problem. It doesn't happen too often, so I just have fun with it.

dcodeIO:
I worry about that too, yeah. ITCM is a relatively basic GC that I don't expect good performance of, but it's the most practical to start with at this point in time. It also still requires the developer to manually call `gc.collect()` to reclaim memory, which isn't ideal for sure but at least provides a way to do it, at a user-defined point in time, i.e. when the application is considered idle.

dcodeIO:
As suggested by @kripken on the [call_indirect optimization issue](https://github.com/WebAssembly/binaryen/issues/1973) I created over at Binaryen, the last commits now bakes GC functionality into the compiler statically, eliminating the use of indirect calls and the function table.

This adds one more interface to implement to tracing GCs, `__ref_mark`, and utilizes the `__gc_mark_roots` and `__gc_mark_members` builtins to perform traversal, calling the mark implementation.

This makes it a bit harder to grasp as a GC implementer now, but considering that its ultimately me doing this it's probably fine. Also adds additional compiler-generated code to a module in the form of a big switch in `__gc_mark_members`, which in turn allows to host all generated traversal code there instead of making one function per class for it. Should be fine :)

dcodeIO:
Alright, dynamic `instanceof` is implemented now, so this works:

```ts
class Animal {]
class Cat extends Animal {}

var whichAnimal: Animal = new Cat(); // pretend the actual kind of animal is unknown
if (whichAnimal instanceof Cat) {
  ...
}
```

Externally, there is also a helper that can be exported together with runtime internals:

```ts
// entry.ts
export { runtime };
``` 

```js
// binding.js
myModule.runtime.instanceOf(someRef, STRING_ID);
```

though it's still non-trivial to obtain `STRING_ID` (the internal class id of the `String` class). Going to think about a proper way, like when `export { String }`, make it have a static property `__id` or something.

MaxGraey:
Well done!
btw `instanceof` is operator so `instanceOf` function inside runtime may be make sense wrap like: ```
```ts
@global
@decorator("instanceof")
function instanceOf(...) {
   ...
}
```
wdyt?

dcodeIO:
I see, `runtime.instanceof = function() {}` actually works in JS and is callable, one just can't `function instanceof() {}` ... hmm.

dcodeIO:
Regarding overriding `instanceof` as an operator: I think that'd be unsafe as it undermines static type checking. Not sure.

dcodeIO:
Last commit made me think. To properly "use" the runtime, one would have to `export { runtime, gc }` from their program

```wat
  (export "memory" (memory 0))
  (export "table" (table 0))
  (export "runtime.instanceof" (func $runtime.instanceof))
  (export "runtime.allocate" (func $runtime.allocate))
  (export "runtime.discard" (func $runtime.discard))
  (export "runtime.register" (func $runtime.register))
  (export "runtime.newString" (func $runtime.newString))
  (export "runtime.newArrayBuffer" (func $runtime.newArrayBuffer))
  (export "runtime.newArray" (func $runtime.newArray))
  (export "gc.implemented" (global 6))
  (export "gc.collect" (func $gc.collect))
  (export "gc.retain" (func $gc.retain))
  (export "gc.release" (func $gc.release))
  (export ".setargc" (func $.setargc))
  (export ".capabilities" (global 9))
```

compiling to about 7KB of runtime code. I see this leading to the misconception that AS modules are huge, having to educate everyone about `--runtime none` etc. Problem is, that without exporting runtime, interchanging strings with the host becomes pretty much impossible.

dcodeIO:
This now

* Combines runtime/gc into just runtime
* Moves `runtime.reallocate` to util because it's of limited external use
* Comments out `memcpy`, relying solely on `memmove` under the assumption that copy operations are aligned

making it 3.9KB with and 2.8KB without assertions. Still odd that assertions alone account for 1.1KB (most of them seem to be in TLSF), but overall those aren't bad numbers for having a proper memory manager, a simple garbage collector and the runtime helpers that the loader needs.

dcodeIO:
Notes on how the externally visible runtime interface might look like: https://github.com/AssemblyScript/assemblyscript/wiki/Runtime#runtime-functions

MaxGraey:
looks good

dcodeIO:
One thing that bothers me is the `$.newArray` interface. I'd prefer if it would take a backing buffer instead, wrapping it, but still need to figure out how that'd look when used by the compiler internally.

leodasvacas:
@dcodeIO Those functions are going to be convenient, thanks! With them we might consider using the gc.

leodasvacas:
@dcodeIO Ah well noted, we use those. But then again we don't need the gc.

dcodeIO:
Maybe one way to expose this API would be an additional argument to `$.retain` that can either be omitted to retain the object externally or be provided to retain a managed object on another one, like in the case of `Array<Ref>` retain all `Ref` instances on the array, so these become collected if the array does. I guess this stands and falls with documentation.

dcodeIO:
Another take on this could be to include more type information with a module, like 32-bit flags per managed object type representing

```ts
/** Runtime flags. */
const enum RuntimeFlags { // keep in sync with src/program.ts
  NONE = 0,
  /** Value alignment of 1 byte. */
  VALUE_ALIGN_0 = 1 << 0,
  /** Value alignment of 2 bytes. */
  VALUE_ALIGN_1 = 1 << 1,
  /** Value alignment of 4 bytes. */
  VALUE_ALIGN_2 = 1 << 2,
  /** Value alignment of 8 bytes. */
  VALUE_ALIGN_3 = 1 << 3,
  /** Value alignment of 16 bytes. */
  VALUE_ALIGN_4 = 1 << 4,
  /** Value type is nullable. */
  VALUE_NULLABLE = 1 << 5,
  /** Value type is managed. */
  VALUE_MANAGED = 1 << 6,
  /** Key alignment of 1 byte. */
  KEY_ALIGN_0 = 1 << 7,
  /** Key alignment of 2 bytes. */
  KEY_ALIGN_1 = 1 << 8,
  /** Key alignment of 4 bytes. */
  KEY_ALIGN_2 = 1 << 9,
  /** Key alignment of 8 bytes. */
  KEY_ALIGN_3 = 1 << 10,
  /** Key alignment of 16 bytes. */
  KEY_ALIGN_4 = 1 << 11,
  /** Key type is nullable. */
  KEY_NULLABLE = 1 << 12,
  /** Key type is managed. */
  KEY_MANAGED = 1 << 13
}
```

So there'd be runtime information available for types like `Array<Ref>` that would eliminate some parameters to `$.newArray` and also includes information whether elements must be linked with the array. Ultimately this goes into the direction of reflection but is less heavy than `Reflect`, hmm.

dcodeIO:
Still need to wire this to the compiler, but once that's done the `$.newArray` interface will look like

* $.**newArray**(id: `u32`, buffer?: `usize`): `usize`

obtaining the relevant information by looking up the runtime flags for `id`, like alignment, whether elements are managed and so on. Also has the advantage that `__runtime_flags` validates that it's indeed an array being created.

Combining this with `$.newArrayBuffer` that is pre-populated with the backing data then yields a proper allocated, registered and linked array with externally created data.

dcodeIO:
So, I've been thinking/unproductive about this for a while now because I'm not satisfied with it yet. More specifically I think that `RuntimeFlags` goes into the right direction, but that it might make sense to generalize this even more to provide not just type info for managed objects but for pretty much every type used in a program. This, however, can be extended upon up to the point where full reflection information is present, like runtime descriptions of all the classes that'd ultimately allow us to do more dynamic stuff, but I can't see the full picture yet, i.e. where this'd become useful practically. Maybe serialization/deserialization, hard to tell.

jtenner:
Reflection would be very useful for testing libraries. Definitely!

dcodeIO:
With RTTI in memory and implementations in stdlib, this brings us to 3.7KB (2.6KB w/o assertions) for an -O3 empty module with exposed runtime.

RTTI is still pretty basic and currently holds mostly information necessary for `$.newArray`, but I now made it so that the definition is shared between stdlib and the compiler so extending should be straight-forward.

dcodeIO:
One thing that isn't well-though-out yet is how to integrate properly with the loader now. For example, if an `abort`, `trace` or anything else that involves something with a runtime header happens during instantiation, it is not yet known whether a full or half header is present, so logging a string correctly at this stage becomes guesswork. One solution to this problem could be to do it like C and avoid using the `(start ..)` entry point of WASM altogether, instead always exporting a function to *start* a module explicitly when instantiation is complete. Related: https://github.com/AssemblyScript/assemblyscript/issues/566

dcodeIO:
One way to deal with the issue mentioned in the last comment would be to make the header fixed size btw, but that means 8 bytes unused overhead where no GC is used. I *think* that'd be OK because using anything of stdlib without a GC will most likely leak anyway (like backing buffers that become resized automatically) and not using a GC should be deprecated better sooner than later. Wdyt?

dcodeIO:
Still somewhat frustrated with the runtime stuff, so I decided to use the new tools in this branch to implement something that doesn't suck: Logical `&&` and `||` now become simplified to a simple `(if ...)` if all we are interested in is trueish/falseish, and there's some initial null checking in flows. From [the test](https://github.com/AssemblyScript/assemblyscript/blob/da4a7751fd0a3174004e1cb275251b5a84e3a6f7/tests/compiler/possibly-null.ts) this seems to work well so far, even though the checks don't account for every possible scenario yet and there are some situations, like propagating throughout logical expressions, that don't fork flows yet.

dcodeIO:
Last commit integrates purerc and tlsf into one package (currently just named asrt for AssemblyScript Runtime), now with a single runtime header.

In the process I actually stumbled upon a couple strange things in TLSF (for instance `get tail` and `set tail` loaded from strange offsets, and tail merging seemed weird) that could have caused corruption when memory is grown (maybe related to https://github.com/AssemblyScript/assemblyscript/issues/548). There's still a fixme on SL layout that I need to check and I haven't run a single test yet.

Also turns out that the common header (block overhead in TLSF) is exactly 16 bytes in WASM32, which is great. Doesn't hold true for WASM64, but considering that this isn't anywhere on the horizon this is fine with me.

dcodeIO:
Still debugging TLSF. There's something very odd when picking fl/sl on allocations and determining maximum block size. Takes a while because I need to build a visual memory debugger for this to know for sure.

MaxGraey:
I remember someone already built visual debugger and as I remember tlsf was very fragmented

dcodeIO:
What I'm after specifically is something that shows fl/sl/heads visually, with a way to allocate and free by clicking, on top of a debug build of asrt that logs stuff :)

jtenner:
Is there any way to query all the allocated blocks? I have a few ideas.

dcodeIO:
No, TLSF doesn't keep lists of "used" blocks, just free blocks. The way this works is that it picks a suitable block from free blocks, possibly splitting a larger one (leaving the remainder as a free block), and giving the allocated block to the user, forgetting about it. Once that's free'd again, it's inserted back into free lists and possibly merged with adjacent free blocks. I think the visualizer I'm working on will explain the concept quite well :)

dcodeIO:
To try it out, run `npm run build && npm run server` in `tests/runtime`. Not really polished yet, though, but helped me to find some issues.

dcodeIO:
Ok I think the allocator should be fairly robust now, but feel free to play with the visualizer and break it. Let me know if you somehow manage to :)

Now thinking how to visualize RC. Maybe similar to the fuzz stuff on the purerc repo, with buttons to allocate objects and arrays of various sizes and dragging/dropping them or pushing retain/release buttons - but not sure if that's worth it and I should instead wire stuff into the compiler right away.

dcodeIO:
Idea now is that some sort of runtime must be present in every program in the future, be it our TLSF/PureRC runtime or something custom. Difference to before is that not the allocators and collectors are pluggable, but the runtime is pluggable _and_ will be fixed to ARC (no tracing anymore because the cognitive and code overhead simply isn't worth it imo). It'll still be possible to have a stub runtime without a GC and just the arena allocator, of course.

dcodeIO:
As you might have guessed, rethinking/rewriting this stuff over and over again turned out to be somewhat exhausting on my end. Current state is that I'm looking at completely red code that came from removing the old memory allocators, garbage collectors and previous runtime efforts, while trying to wrap my head around how reference counting (here: automatic insertion of runtime calls like retain/release) can integrate with stdlib which sometimes needs to use these interfaces manually by means of `changetype`. Not being able to test anything before everything is at least somewhat in place makes this quite a challenge.

dcodeIO:
Quick update: Code isn't red anymore, integrates with updated stdlib and uses the new runtime API, but still lacks some of the RC integration when it comes to locals. Essentially, whenever a reference is provided as a function argument, it must be retained pre-call and released post-call (possibly triggering free), plus I still have to investigate retain/release calls on locals when assigning references between them. From what I've learned so far, much of the work on Swift, which uses ARC, evolves around static elimination of retain/release calls where refcount can be statically proven to remain unchanged, because each increment and decrement may involve a cache miss. From my current perspective, getting something working out the door first has priority ofc.

dcodeIO:
So I've come to the conclusion that whenever I try to be smart about avoiding unnecessary retain/release calls, this tends to result in situations that can't be solved without further code analysis down the road, which we don't do but usually ask Binaryen to do after transforming to SSA form and similar. Currently looks like the compiler itself must be super paranoid about everything reference counting at first and then rely on optimization passes anyway. Then doing it the paranoid way leads to the return value problem again, where since we don't have a stack there must be some mechanism to make sure that these don't become released before the caller possibly retains them. I think Obj-C calls this autorelease pools or something, need to investigate.

dcodeIO:
Turns out I might actually be able to be smarter than expected about this, completely eliminating retains/releases of locals and avoiding the need for autorelease pools, by using special temp locals with an AUTORELEASE property. Chances are that I'm overlooking something again, of course, but looks promising:

```ts
class Ref {}

// Returning a reference first retains it on return and then
// is tracked by the caller's surrounding scope:

function returnRef(): Ref {
  return /* __retain( */ changetype<Ref>(0) /* ) */;
}

export function testReturnRef(): void {
  /* TEMP = */ returnRef();
  // __release(TEMP)
}

// Taking a reference doesn't insert anything. If there was an
// allocation, it would be handled by the caller.

function takeRef(ref: Ref): void {
}

export function testTakeRef(): void {
  takeRef(changetype<Ref>(0));
}

// Allocating a reference keeps track of it in the surrounding
// scope and releases it on exit.

export function newRef(): void {
  /* TEMP = */ new Ref();
  // __release(TEMP)
}

// Assigning a reference to a global retains it, releasing the old value.

var glo: Ref;

export function assignGlobal(): void {
  glo = /* __retainRelease( */ changetype<Ref>(0) /* , glo) */;
}

// Assigning a reference to a field retains it, releasing the old value.

class Target { fld: Ref; }

export function assignField(): void {
  changetype<Target>(0).fld = /* __retainRelease( */ changetype<Ref>(0) /* , fld) */;
}
```

Remains the question: What am I missing this time? :)

dcodeIO:
Closing in favor of https://github.com/AssemblyScript/assemblyscript/pull/592

PR #612, opened 2019-05-28
Move OptionDescription to utils/options so types agree

willemneal:
I found this when using the options parser in as-pect and `asc.exports` disagreed with `util.parse`'s expected argument type `Config`.

### DIFF
--- a/cli/asc.d.ts
+++ b/cli/asc.d.ts
@@ -1,4 +1,4 @@
-import { OptionDescription }from "./util/options";
+import { OptionDescription } from "./util/options";
 export { OptionDescription };
 
 /** Whether this is a webpack bundle or not. */
--- a/cli/asc.d.ts
+++ b/cli/asc.d.ts
@@ -1,5 +1,5 @@
-import {OptionDescription}from "./util/options";
-export {OptionDescription};
+import { OptionDescription }from "./util/options";
+export { OptionDescription };
 
 /** Whether this is a webpack bundle or not. */
 export const isBundle: boolean;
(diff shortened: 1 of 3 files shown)


cli/asc.d.ts
```diff
@@ -1,3 +1,6 @@
+import {OptionDescription}from "./util/options";
+export {OptionDescription};
```
dcodeIO: style nit: 

```ts
import { OptionDescription } from "./util/options";
export { OptionDescription };
```

dcodeIO:
Looks good, with my minor comment resolved :)

cli/asc.d.ts
```diff
@@ -1,5 +1,5 @@
-import {OptionDescription}from "./util/options";
-export {OptionDescription};
+import { OptionDescription }from "./util/options";
```
dcodeIO: missing a whitespace :)

cli/util/options.d.ts
```diff
@@ -3,7 +3,7 @@ export interface OptionDescription {
   /** Textual description. */
   description?: string | string[],
   /** Data type. One of (b)oolean [default], (i)nteger, (f)loat or (s)tring. Uppercase means multiple values. */
-  type?: "b" | "i" | "f" | "s", "I", "F", "S",
+  type?: "b" | "i" | "f" | "s" | "I" | "F" | "S",
```
dcodeIO: nice find! did that even work before?

willemneal: Nut sure it was ever used.  Since it was just the type definition file and no one tried to use the type with a typescript project.

PR #636, opened 2019-06-06
Add RTrace support and Types

jtenner:
Include `rtrace` with the `assemblyscript` package.

image

- [x] modifies `package.json` to include files
- [x] adds `lib/rtrace/index.d.ts` types for tooling
- [x] adds code documentation (to the best of my ability)

All feedback is welcome.

### DIFF
--- a/lib/rtrace/index.d.ts
+++ b/lib/rtrace/index.d.ts
@@ -1,80 +1,82 @@
 /**
- * The RTrace interface represents a collection of properties that help develoeprs describe the
- * reference counts and state of the Web Assembly module.
- */
-export interface RTrace {
-
-  /**
-   * The current allocation count.
-   */
-  allocCount: number;
-
-  /**
-   * The current free count.
-   */
-  freeCount: number;
-
-  /**
-   * The current increment count.
-   */
-  incrementCount: number;
-
-  /**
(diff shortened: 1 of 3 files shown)


lib/rtrace/index.d.ts
```diff
@@ -0,0 +1,81 @@
+/**
+ * The RTrace interface represents a collection of properties that help develoeprs describe the
+ * reference counts and state of the Web Assembly module. It is used by the AssemblyScript loader
```
dcodeIO: Isn't really used by the loader, but more of an extra package to include as an import, independently of whether the loader is used or not :)

lib/rtrace/index.d.ts
```diff
@@ -0,0 +1,81 @@
+/**
+ * The RTrace interface represents a collection of properties that help develoeprs describe the
```
dcodeIO: "developers"

jtenner: All set for review :)

lib/rtrace/index.d.ts
jtenner: The line was simply removed. All set for review.

dcodeIO:
Thanks! :)

PR #637, opened 2019-06-07
Added paths option to resolve stdlib imports in IDE's

willemneal:
Currently IDE's can't resolve stdlib imports that aren't relative.  This uses the ts path compiler option to try to resolve non-relative paths with the stdlib. 

This also fixes two typing errors that IDE found.

### DIFF
--- a/std/assembly/rt/index.d.ts
+++ b/std/assembly/rt/index.d.ts
@@ -10,3 +10,4 @@ declare function __visit(ref: usize, cookie: i32): void;
 declare function __visit_globals(cookie: u32): void;
 declare function __visit_members(ref: usize, cookie: u32): void;
 declare function __allocArray(length: i32, alignLog2: usize, id: u32, data?: usize): usize;
+declare const ASC_RTRACE: boolean;
--- a/std/assembly/rt/pure.ts
+++ b/std/assembly/rt/pure.ts
@@ -15,7 +15,7 @@ import { onincrement, ondecrement, onfree, onalloc } from "./rtrace";
 // B: buffered
 
 // @ts-ignore: decorator
-@inline const BUFFERED_MASK: u32 = 1 << (sizeof<u32>() * 8 - 1);
+@inline const BUFFERED_MASK: u32 = 1 << ((sizeof<u32>() * 8) - 1);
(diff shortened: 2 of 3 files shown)


std/assembly/rt/pure.ts
```diff
@@ -15,7 +15,7 @@ import { onincrement, ondecrement, onfree, onalloc } from "./rtrace";
 // B: buffered
 
 // @ts-ignore: decorator
-@inline const BUFFERED_MASK: u32 = 1 << (sizeof<u32>() * 8 - 1);
+@inline const BUFFERED_MASK: u32 = 1 << ((sizeof<u32>() * 8) - 1);
```
MaxGraey: this unnecessary. Multiply have bigger precedence than subtraction

std/assembly/rt/index.d.ts
```diff
@@ -10,3 +10,4 @@ declare function __visit(ref: usize, cookie: i32): void;
 declare function __visit_globals(cookie: u32): void;
 declare function __visit_members(ref: usize, cookie: u32): void;
 declare function __allocArray(length: i32, alignLog2: usize, id: u32, data?: usize): usize;
+declare const ASC_RTRACE: boolean;
```
MaxGraey: this declaration better include [here](https://github.com/AssemblyScript/assemblyscript/blob/master/std/assembly/index.d.ts#L60) I guess

dcodeIO: note that this is actually not present at all without `--use ASC_RTRACE=1`, and meant to be used as `isDefined(ASC_RTRACE)`.

std/assembly/rt/pure.ts
willemneal: I know but the syntax confused my IDE.

Screen Shot 2019-06-07 at 1 06 00 PM

std/assembly/rt/index.d.ts
willemneal: Which is my I thought it best to put here since this is the folder where it is referenced and so IDE's will use it.  And it seemed like an internal definition so I didn't want to put it in the higher level file.

dcodeIO:
Thanks!

PR #644, opened 2019-06-09
Rework loop compilation / flows

dcodeIO:
As reported in https://github.com/AssemblyScript/assemblyscript/issues/635, flows did not properly recognize termination if multiple distinct termination conditions were present in different arms.

This PR attempts to solve this by restructuring flow flags to include a general `TERMINATES` flag that is set whenever a flow `RETURNS`, `THROWS` or `CONTINUES`.

Additionally, code generation for `for`, `while`, `do` and `switch` has been made a bit smarter in that it now recognize parts of generated code that are unnecessary.


dcodeIO:
One thing not particularly beautiful here is the [unreachable spam](https://github.com/AssemblyScript/assemblyscript/pull/644/files#diff-934585a85a9df7679a250f2315d98aa2R197) that happens somewhere between generation and output. My guess is that Binaryen adds these in an attempt to handle its custom unreachable type, making sure that for example blocks of that type become properly terminated, but going a bit overboard with it, like sometimes also terminating unreachable itself with an unreachable. @kripken maybe?

kripken:
@dcodeIO Some things in binaryen emit extra unreachables (like the relooper). But the optimizer should be able to clean up up very well. Have you seen a case of optimized code that looks bad?

dcodeIO:
Thanks @kripken , just created a [little issue](https://github.com/WebAssembly/binaryen/issues/2169) with more details. Not really an issue of course, just happened to be a bit confusing because AssemblyScript uses unoptimized StackIR for its tests.

Issue #650, opened 2019-06-10
Array of bools is not allowed in new runtime


willemneal:
```
ERROR AS200: Conversion from type 'i32' to 'bool' requires an explicit cast.

     return element; // no need to retain -> is moved
            ~~~~~~~
 in ~lib/array.ts(290,11)
```

dcodeIO:
Ah, issue here is on line

```ts
var element = load<T>(this.dataStart + (<usize>(--length) << alignof<T>()));
```

Here, the type for `element` is inferred, starting by assuming `i32`, and `load<bool>` will try to load to the target type. Currently, the fix is

```ts
var element: T = load<T>(...)`
```

but there really should be something better. One thought I had was to have a contextual flag for `RETAIN_TYPE` as an alternative, that the `load` built-in could use to detect such cases.

dcodeIO:
Should have been fixed by the commit linked above. Feel free to reopen if necessary!

PR #658, opened 2019-06-12
Use a special 'auto' type in type inference

dcodeIO:
This PR replaces the previous `compileExpressionRetainType` functionality, that assumed `i32` in inference, with a special `Type.auto` alias of `Type.i32` that can be used by subsequent steps to determine whether inference is taking place.

Previously, using just `i32` as the contextual type could lead to issues like https://github.com/AssemblyScript/assemblyscript/issues/650 where the `load<T>` built-in didn't know whether it is being called as `var a = <i32>load<bool>` or `var a = load<bool>`, in turn leading to an inferred type of `i32` instead of `bool`. With the change, the `load<T>` built-in can simply check for `contextualType == Type.auto`.

Also reverts contextual flags for statements because there was essentially only one that can as well be handled as a parameter instead, so computing inherited flags in every single statement sub-function isn't necessary.

### DIFF
--- a/src/builtins.ts
+++ b/src/builtins.ts
@@ -5,7 +5,7 @@
 
  import {
   Compiler,
-  ContextualFlags,
+  Constraints,
   RuntimeFeatures,
   flatten
 } from "./compiler";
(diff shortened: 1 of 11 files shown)


dcodeIO:
It also turned out that there's something wrong with `makeAutorelease` and friends in that it loses information about nonNull states. Going to look into this in a future PR.

PR #659, opened 2019-06-12
[Feature] add offset overloads

jtenner:
This is a really simple `index.d.ts` change. I left the general case intact just in case. This improves the developer experience by adding suggested field names in intellisense.

Documentation should be consistent with the wording used by the original function definition because it was copy pasted.

### DIFF
--- a/std/assembly/index.d.ts
+++ b/std/assembly/index.d.ts
@@ -110,6 +110,10 @@ declare const __heap_base: usize;
 declare function sizeof<T>(): usize;
 /** Determines the alignment (log2) of the specified underlying core type. Compiles to a constant. */
 declare function alignof<T>(): usize;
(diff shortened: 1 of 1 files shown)


MaxGraey:
May be just one line?
```ts
declare function offsetof<T>(fieldName?: keyof T | string): usize;
```
without other overloads

jtenner:
I was going back and forth between that and what I settled on because having multiple implementations made the inline description easier to read. For instance, if the parameter is withheld, it shows a shorter description of what to expect.

image

Personal taste. If @dcodeIO and @MaxGraey (you) decide that you would prefer the single line implementation, I have no problem with changing it because both solutions are valid.

MaxGraey:
@jtenner Got it. Makes sense

dcodeIO:
Thanks!

PR #662, opened 2019-06-13
Investigate custom try-catch support

dcodeIO:
This PR aims at investigating ways to at least support basic try-catch without finally. Idea is that whenever an error is `throw`n, it will be remembered in a helper global. Now, if there is a reachable catch block in the same function, the scope will be cleaned and control flow branches there, or if there's not, the function will be terminated forcefully, possibly with a return value of whatever is zero.

Each callsite then evaluates the flow of the called function to determine whether the call possibly throws, and if so inserts a check whether an error is present right after the call. This continues until a reachable catch block is found in a calling function that handles the error, moving the error into the catch variable and forgetting about it, or, if it's never caught, leading to an `uncaught error` abort at top-level.

The implementation is somewhat fragile currently because the compiler already has to handle a shit ton of cases everywhere, misses a way to catch externally and leaks memory for reasons I still have to investigate, but I thought I share what I have so far.


dcodeIO:
Turns out that the issues here are very similar to the runtime's problems wrapping stuff in retain/release, just that it's now wrapping expressions, incl. the runtime's expressions which already try to use clever workarounds, in interrupts that must (auto)release. I think this is broken beyond repair.

dcodeIO:
Closing this PR as part of [2020 vacuum](https://github.com/AssemblyScript/assemblyscript/issues/1277) as it is outdated. With exceptions support landing in Binaryen and becoming more mature, and exceptions not being mentioned for months making this seem not particularly urgent, my expectation is that we'll go straight for the feature instead of rolling our own.

PR #677, opened 2019-06-19
Allow internal abort to be disabled

dcodeIO:
As mentioned in https://github.com/AssemblyScript/assemblyscript/issues/676, it is not possible to completely disable the use of the internal `abort` function, making it necessary to at least provide an alternative. This PR changes this by just emitting an `unreachable` instruction if the abort function has been disabled explicitly through `--use abort=`.


dcodeIO:
Docs: https://docs.assemblyscript.org/details/debugging#overriding-abort

PR #681, opened 2019-06-20
Improve precomputation of static instanceof

dcodeIO:
This allows code like

```ts
if (expr instanceof ArrayBuffer) {
  ...
} else {
  ...
}
```

to statically eliminate the `else` branch again in the common case that `expr` has no side-effects (leading to just a constant instead of a block also dropping `expr`). This was supported before but stopped working due to being more cautious to preserve potential side-effects here.

Note that this only affects the case where static evaluation can take place, like in a downcast with known `T` of `expr`.


dcodeIO:
Last commit tries to solve this in a more general way by applying a `vacuum` pass prior to performing `precompute`. This removes any unnecessary `drop`s and other noise without having to deal with this during codegen.

dcodeIO:
One additional optimization I came across while looking into this was be to attempt solving `assert`s statically where possible, not emitting any actual code. Makes some of our tests largely evaporate, like the WASI test that checks static offsets, and would allow to have static assertions in runtime code, like TLSF, that don't end up as actual code then. Not sure yet whether this deserves another builtin, like `static_assert` or if it's ok to make `assert` adapt to it. Future PR.

jtenner:
Wow. That makes things really interesting for people who use the `assert()` function. Nice observation.

PR #684, opened 2019-06-20
Utilize new inlining utility to potentially inline any call

dcodeIO:
As reported in https://github.com/MaxGraey/bignum.wasm/issues/20 the compiler didn't yet attempt to inline all possible call sites (while emitting an AS224 warning), but this should be possible now by using the utility introduced with the runtime branch. Essentially, whenever a call is inlined, the locals used in operands are blocked right-to-left so other temps don't conflict.


dcodeIO:
One side-product of this is that the `tests/std/pointer` experiment can now inline everything.

PR #703, opened 2019-07-02
Cleanup resolver API

dcodeIO:
Started this a while ago but became distracted by other things happening. There's some cleanup with mostly naming changes in here for the resolver API, so the compiler should be updated to use a similar `ctxXY` naming scheme eventually I think, but that's minor I guess. Idea is to go through it once before attempting to rework the remaining issues where it can't resolve any expression.


dcodeIO:
Merging to get this out of the way, continuing from here.

PR #710, opened 2019-07-08
Implement --noUnsafe compiler option

dcodeIO:
This is an initial implementation of the general idea of an "unsafe" compiler option as described in https://github.com/AssemblyScript/assemblyscript/issues/709, but a bit simpler.

* `--noUnsafe`
  Disallows the use of unsafe features in user code. Does not affect library files and external modules

This is fully backwards compatible in that the default is not changed, yet an embedder may now decide to disallow unsafe features in user code in order to protect from undefined behavior, like corruption of runtime structures by rogue stores and similar.

Instead of making stdlib files contagious as suggested in the original issue linked above, this PR refactors source kinds into

* `USER_ENTRY`
  normal entry file where exports become module exports
* `USER`
  normal file imported from user code
* `LIBRARY_ENTRY`
  library entry file where exports become globals
* `LIBRARY`
  normal library file

to keep all options open, e.g. if it ever becomes necessary to specify the file kind in asc. The runtime entry file is somewhat special in that it is both a library file and a user entry file (where exports become module exports), so it resorts to `USER_ENTRY`, since the RT entry doesn't use unsafe code anyway but just re-exports.


dcodeIO:
Should now [handle most, if not all](https://github.com/AssemblyScript/assemblyscript/blob/e46bfef51d5cba2266db68bafc56edbb83a471c4/tests/compiler/unsafe.ts), of the possibilities where something unsafe is used and must be checked. If you notice something I've missed, please let me know :)

dcodeIO:
Last commit makes the implementation fully backwards compatible and much simpler, in that disallowing unsafe features is now opt-in. Updated the initial post accordingly.

dcodeIO:
Since no objections so far, and the new error handling being useful in an upcoming PR as it seems, I'm going to merge. Can still iterate upon the current state if there are additional requirements.

PR #713, opened 2019-07-10
1. Asc command supported absolute path.

LiaoPeng:
Got the below errors, when executed command asc /Workspace/assemblyscript/tests/compiler/for.ts -t for.wast

The reason is that the asc command works improperly when using absolute path.

```typescript
 asc /Workspace/assemblyscript/tests/compiler/for.ts -t for.wast
ERROR: Entry file '/Workspace/assemblyscript/tests/compiler/for.ts' not found.
    at Object.main (/Users/peng/Workspace/assemblyscript/cli/asc.js:391:25)
    at Object.<anonymous> (/Users/peng/Workspace/assemblyscript/bin/asc:21:26)
    at Module._compile (internal/modules/cjs/loader.js:774:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:785:10)
    at Module.load (internal/modules/cjs/loader.js:641:32)
    at Function.Module._load (internal/modules/cjs/loader.js:556:12)
    at Function.Module.runMain (internal/modules/cjs/loader.js:837:10)
    at internal/main/run_main_module.js:17:11
```

### DIFF
--- a/cli/asc.js
+++ b/cli/asc.js
@@ -380,6 +380,8 @@ exports.main = function main(argv, options, callback) {
     const filename = argv[i];
 
     let sourcePath = String(filename).replace(/\\/g, "/").replace(/(\.ts|\/)$/, "");
+    // Setting the path to relative path
+    sourcePath = sourcePath.indexOf(baseDir) == 0 ? sourcePath.slice(baseDir.length) : sourcePath;
 
     // Try entryPath.ts, then entryPath/index.ts
     let sourceText = readFile(sourcePath + ".ts", baseDir);
--- a/cli/asc.js
+++ b/cli/asc.js
@@ -381,7 +381,7 @@ exports.main = function main(argv, options, callback) {
 
     let sourcePath = String(filename).replace(/\\/g, "/").replace(/(\.ts|\/)$/, "");
     // Setting the path to relative path
-    sourcePath = sourcePath.indexOf(baseDir) == 0 ? sourcePath.slice(baseDir.length) : sourcePath;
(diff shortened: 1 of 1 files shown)


cli/asc.js
```diff
@@ -380,6 +380,8 @@ exports.main = function main(argv, options, callback) {
     const filename = argv[i];
 
     let sourcePath = String(filename).replace(/\\/g, "/").replace(/(\.ts|\/)$/, "");
+    // Setting the path to relative path
+    sourcePath = sourcePath.indexOf(baseDir) == 0 ? sourcePath.slice(baseDir.length) : sourcePath;
```
dcodeIO: This somehow feels like half a fix for paths inside `baseDir` only. Or does it work for other absolute paths outside `baseDir` properly already?

LiaoPeng: I missed this test case.  Now I fixed it. 

What do you think the way to change implementation of [`readFileNode`](https://github.com/LiaoPeng/assemblyscript/blob/ab94853a23dbf33451f64c6ca631cc6cfad6eb3d/cli/asc.js#L727) like this?

```
function readFileNode(filename, baseDir) {
    try {
      let text;
      stats.readCount++;
      stats.readTime += measure(() => {
        let filePath = path.isAbsolute(filename) ? filename :  path.join(baseDir, filename);
        text = fs.readFileSync(filePath , { encoding: "utf8" });
      });
      return text;
    } catch (e) {
      return null;
    }
  }
```

dcodeIO:
Thanks!

PR #716, opened 2019-07-17
Fix missing argument when specifying a --lib directory

dcodeIO:
As discovered in https://github.com/AssemblyScript/assemblyscript/issues/714 the respective call to `listFiles` in asc when a directory was provided with `--lib` missed the necessary second parameter, leading to the described issue.

### DIFF
--- a/cli/asc.js
+++ b/cli/asc.js
@@ -259,7 +259,7 @@ exports.main = function main(argv, options, callback) {
         libFiles = [ path.basename(libDir) ];
         libDir = path.dirname(libDir);
       } else {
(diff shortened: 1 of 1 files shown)


dcodeIO:
@jtenner: Can you confirm that this fixes the problem?

jtenner:
Should be about an hour. I'll confirm soon!

jtenner:
Confirmed. This fixes the problem. 

I removed the top level import statement in the test, and `Buffer` was added as a global.

Thank you!

dcodeIO:
Thanks for the confirmation! :)

PR #721, opened 2019-07-19
[Fix] Static class callback fixes #720

### LINKED ISSUE #720: [Bug] Static class method cannot be used as callback
```
ERROR AS218: Cannot access method 'compare' without calling it as it requires 'this' to be set.

     actual.sort(Buffer.compare);
                 ~~~~~~~~~~~~~~
 in tests/buffer.spec.ts(60,16)
```

Thoughts?

EDIT: Found here: https://github.com/AssemblyScript/node/pull/7

jtenner:
This is my first pass at fixing #720.

Thoughts?

### DIFF
--- a/src/compiler.ts
+++ b/src/compiler.ts
@@ -7989,17 +7989,26 @@ export class Compiler extends DiagnosticEmitter {
         );
       }
       case ElementKind.FUNCTION_PROTOTYPE: {
-        let prototypeInstance = <FunctionPrototype>target;
+        let prototype = <FunctionPrototype>target;
 
-        if (prototypeInstance.is(CommonFlags.STATIC)) {
-          let instance = this.compileFunctionUsingTypeArguments(prototypeInstance, [])!;
-          this.currentType = instance.type;
-          return module.i32(this.ensureFunctionTableEntry(instance));
+        if (prototype.is(CommonFlags.STATIC)) {
+          let instance = this.compileFunctionUsingTypeArguments(
+            prototype,
+            [],
+            makeMap<string,Type>(),
+            propertyAccess,
+          );
+          if (instance == null) {
+            return module.unreachable();
+          } else {
+            this.currentType = instance.type;
+            return module.i32(this.ensureFunctionTableEntry(instance));
+          }
         }
 
         this.error(
           DiagnosticCode.Cannot_access_method_0_without_calling_it_as_it_requires_this_to_be_set,
-          propertyAccess.range, prototypeInstance.name
+          propertyAccess.range, prototype.name
         );
         return module.unreachable();
       }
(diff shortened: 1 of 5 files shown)


src/compiler.ts
```diff
@@ -7989,9 +7989,17 @@ export class Compiler extends DiagnosticEmitter {
         );
       }
       case ElementKind.FUNCTION_PROTOTYPE: {
+        let prototypeInstance = <FunctionPrototype>target;
+
+        if (prototypeInstance.is(CommonFlags.STATIC)) {
+          let instance = this.compileFunctionUsingTypeArguments(prototypeInstance, [])!;
```
dcodeIO: This can return `null`, in which case an `unreachable` should be emitted. Also, providing `alternativeReportNode = propertyAccess` will refer to a better code location in the error case, as it will otherwise report on the original function, which misses a clue where the error actually happened.

src/compiler.ts
```diff
@@ -7989,9 +7989,17 @@ export class Compiler extends DiagnosticEmitter {
         );
       }
       case ElementKind.FUNCTION_PROTOTYPE: {
+        let prototypeInstance = <FunctionPrototype>target;
```
dcodeIO: Typically named just "prototype" in other parts of the compiler.

dcodeIO:
LGTM with the comments addressed :)

jtenner:
This is definitely my first time trying to work with the compiler. Thanks for the information.

src/compiler.ts
jtenner: Should we emit a compiler error here instead of `unreachable()` when the return is `null`? I'm not sure what returning null actually means.

Also the property node fix makes perfect sense. I'll get around to this soon.

dcodeIO: What happens in situations like these is that some part of the code finds that an operation is invalid, so it emits a diagnostic and returns `null`. That's the case here as well: An error has already been emitted if `compileFunctionUsingTypeArguments` returns `null`.

dcodeIO: If you are interested in seeing what happens in such a case, make the function being used generic. That'd result in a diagnostic in the resolve step of `compileFunctionUsingTypeArguments`, which returns `null`, making `compileFunctionUsingTypeArguments` return null, leading to an `unreachable` here.

jtenner: That makes perfect sense. Thank you for taking the time to explain this.

src/compiler.ts
jtenner: All set!

src/compiler.ts
jtenner: All set!

dcodeIO:
Thanks! :)

PR #730, opened 2019-07-26
[Implement] idof<T>() for function references

jtenner:
All set for review. Am I missing something?

Thanks for your help again!

### DIFF
--- a/src/builtins.ts
+++ b/src/builtins.ts
@@ -3592,7 +3592,12 @@ export function compileCall(
     case BuiltinSymbols.idof: {
       let type = evaluateConstantType(compiler, typeArguments, operands, reportNode);
       compiler.currentType = Type.u32;
-      if (!type) return module.unreachable();
+      if (!type || checkArgsOptional(operands, 0, 0, reportNode, compiler)) return module.unreachable();
+
+      if (type.is(TypeFlags.REFERENCE) && type.signatureReference !== null) {
+        return module.i32(type.signatureReference.id);
+      }
+
       let classReference = type.classReference;
       if (!classReference || classReference.hasDecorator(DecoratorFlags.UNMANAGED)) {
         compiler.error(
--- a/src/builtins.ts
+++ b/src/builtins.ts
@@ -3592,7 +3592,7 @@ export function compileCall(
     case BuiltinSymbols.idof: {
       let type = evaluateConstantType(compiler, typeArguments, operands, reportNode);
       compiler.currentType = Type.u32;
-      if (!type || checkArgsOptional(operands, 0, 0, reportNode, compiler)) return module.unreachable();
+      if (!type) return module.unreachable();
 
       if (type.is(TypeFlags.REFERENCE) && type.signatureReference !== null) {
         return module.i32(type.signatureReference.id);
--- a/src/program.ts
+++ b/src/program.ts
@@ -336,8 +336,8 @@ export class Program extends DiagnosticEmitter {
   typeClasses: Map<TypeKind,Class> = new Map();
   /** Managed classes contained in the program, by id. */
   managedClasses: Map<i32,Class> = new Map();
-  signatureTypes: Signature[] = new Array<Signature>(0);
-  signatureID: i32 = 0;
+  /** A set of unique function signatures contained in the program, by id. */
+  uniqueSignatures: Signature[] = new Array<Signature>(0);
 
   // standard references
 
@@ -405,7 +405,8 @@ export class Program extends DiagnosticEmitter {
 
   /** Next class id. */
   nextClassId: u32 = 0;
-
+  /** Next signature id. */
+  nextSignatureId: i32 = 0;
   /** Constructs a new program, optionally inheriting parser diagnostics. */
   constructor(
     /** Shared array of diagnostic messages (emitted so far). */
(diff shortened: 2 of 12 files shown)


src/builtins.ts
```diff
@@ -3592,7 +3592,12 @@ export function compileCall(
     case BuiltinSymbols.idof: {
       let type = evaluateConstantType(compiler, typeArguments, operands, reportNode);
       compiler.currentType = Type.u32;
-      if (!type) return module.unreachable();
+      if (!type || checkArgsOptional(operands, 0, 0, reportNode, compiler)) return module.unreachable();
```
dcodeIO: The arguments check should already be covered by `evaluateConstantType` above

src/program.ts
```diff
@@ -336,6 +336,8 @@ export class Program extends DiagnosticEmitter {
   typeClasses: Map<TypeKind,Class> = new Map();
   /** Managed classes contained in the program, by id. */
   managedClasses: Map<i32,Class> = new Map();
+  signatureTypes: Signature[] = new Array<Signature>(0);
+  signatureID: i32 = 0;
```
dcodeIO: Perhaps `nextSignatureId` and move it right below `nextClassId` for consistency.

src/program.ts
```diff
@@ -336,6 +336,8 @@ export class Program extends DiagnosticEmitter {
   typeClasses: Map<TypeKind,Class> = new Map();
   /** Managed classes contained in the program, by id. */
   managedClasses: Map<i32,Class> = new Map();
+  signatureTypes: Signature[] = new Array<Signature>(0);
```
dcodeIO: Perhaps `uniqueSignatures`.

src/types.ts
```diff
@@ -573,20 +574,37 @@ export class Signature {
...
+    this.program = program;
     this.hasRest = false;
     this.type = Type.u32.asFunction(this);
+
+    let compare: Signature;
+    let signatureTypes = program.signatureTypes;
+    let length = signatureTypes.length;
+    for (let i = 0; i < length; i++) {
+      compare = signatureTypes[i];
+      if (this.isAssignableTo(compare)) {
```
dcodeIO: While this check works currently, the naming suggests that there can be multiple valid signatures. So I'm wondering if we should have an `equals` as well that we can use here, that `isAssignableTo` would be an alias of currently.

src/builtins.ts
jtenner: Okay. I'll remove the additional check.

src/program.ts
jtenner: Okay. Easy enough.

src/program.ts
jtenner: Okay.

src/types.ts
jtenner: Would it make sense to create an `equals<T>(value: T)` that returns `instanceof Signature && this.isAssignableTo(value)`?

dcodeIO: The `equals` I am proposing is a method `Signature#equals` that is then used here, to make this line of code more clear. For instance, in the future a signature with a trailing optional argument might be considered "assignable to" a signature without that trailing optional argument, but what we want here is actual uniqueness I think, hence `equals` is a better word. So I'm proposing:

1. Rename current `Signature#isAssignableTo` to `Signature#equals`
2. Introduce a new `Signature#isAssignableTo` that calls `equals` (for now)
3. Use `isAssignableTo` where something must be "assignable to" and `equals` where something must be "equal"

jtenner: Okay. That sounds reasonable. I'll work on this today.

jtenner: Should be all set!

src/builtins.ts
jtenner: This was removed

src/program.ts
jtenner: All set.

src/types.ts
```diff
@@ -573,20 +574,37 @@ export class Signature {
...
+  /** The program that created this signature. */
+  program: Program;
 
   /** Constructs a new signature. */
   constructor(
+    program: Program,
     parameterTypes: Type[] | null = null,
     returnType: Type | null = null,
-    thisType: Type | null = null
+    thisType: Type | null = null,
```
dcodeIO: nit: trailing comma

src/types.ts
```diff
@@ -573,20 +574,37 @@ export class Signature {
...
     this.thisType = thisType;
+    this.program = program;
     this.hasRest = false;
     this.type = Type.u32.asFunction(this);
+
+    let compare: Signature;
+    let signatureTypes = program.uniqueSignatures;
+    let length = signatureTypes.length;
+    for (let i = 0; i < length; i++) {
+      compare = signatureTypes[i];
```
dcodeIO: nit: `let compare =`, and remove the var statement above

src/types.ts
jtenner: Ah. This is a personal style choice. Sorry about that.

src/types.ts
jtenner: Not a problem. This was left over from an initial implementation that required the Signature to stick around after the loop. Easy fix. Thanks for the heads up.

src/types.ts
jtenner: All set.

src/types.ts
jtenner: Should be all set.

jtenner:
closed in favor of #739

PR #735, opened 2019-07-29
[Implement] ReturnType

### LINKED ISSUE #728: [help] Get return types?
I have the following functions on index.ts file:
```typescript
export function str(): string {
  return "string";
}

export function n32(): i32 {
  return 1
}


export function r32(): f32 {
  return 1.2;
}

export function aStr(): Array<String> {
  return ["a", "b"];
}

export function aN(): Array<i32> {
  return [1, 2];
}

export function aR(): Array<f32> {
  return [1.1, 2.2];
}

export function truth(): bool {
  return true;
}

```

and I'd like to be able to tell, from my code on Go, what is the return type of the called function. My function on Go looks like this:
```go
func runCode(funcName string) {
	raw, err := ioutil.ReadFile("optimized.wasm")
	if err != nil {
		fmt.Println(err)
	}

	r := new(Resolver)
	vm, err := exec.NewVirtualMachine(raw, exec.VMConfig{}, r, nil)
	if err != nil {
		return
	}

	entryID, ok := vm.GetFunctionExport(funcName)
	if !ok {
		fmt.Println("function not found")
	}

	ret, err := vm.Run(entryID)
	if err != nil {
		vm.PrintStackTrace()
		fmt.Println(err)
	}

	fmt.Println("return:", ret)
}
```

The problem I'm facing is that `ret` is sometimes a memory position, sometimes an int32, sometimes a big number (float on IEEE 754 representation). Is there any way to dynamically tell what am I going to receive, so I can properly read it from memory?

jtenner:
fixes #728

Questions: 

1. How can we test this better?
2. Any way to optimize this? (I copied from `valueof<T>` and adapted it)
3. Can we make the error code better?

### DIFF
--- a/src/resolver.ts
+++ b/src/resolver.ts
@@ -273,6 +273,7 @@ export class Resolver extends DiagnosticEmitter {
           case CommonSymbols.native: return this.resolveBuiltinNativeType(node, ctxElement, ctxTypes, reportMode);
           case CommonSymbols.indexof: return this.resolveBuiltinIndexofType(node, ctxElement, ctxTypes, reportMode);
           case CommonSymbols.valueof: return this.resolveBuiltinValueofType(node, ctxElement, ctxTypes, reportMode);
+          case CommonSymbols.ReturnType: return this.resolveBuiltinReturnTypeType(node, ctxElement, ctxTypes, reportMode);
         }
       }
 
@@ -548,6 +549,41 @@ export class Resolver extends DiagnosticEmitter {
       );
     }
     return null;
+  }
+
+  private resolveBuiltinReturnTypeType(
+    /** The type to resolve. */
+    node: NamedTypeNode,
+    /** Contextual element. */
+    ctxElement: Element,
+    /** Contextual types, i.e. `T`. */
+    ctxTypes: Map<string,Type> | null = null,
+    /** How to proceed with eventualy diagnostics. */
+    reportMode: ReportMode = ReportMode.REPORT
+  ): Type | null {
+    var typeArgumentNodes = node.typeArguments;
+    if (!(typeArgumentNodes && typeArgumentNodes.length == 1)) {
+      if (reportMode == ReportMode.REPORT) {
+        this.error(
+          DiagnosticCode.Expected_0_type_arguments_but_got_1,
+          node.range, "1", (typeArgumentNodes ? typeArgumentNodes.length : 1).toString(10)
+        );
+      }
+      return null;
+    }
+    var typeArgument = this.resolveType(typeArgumentNodes[0], ctxElement, ctxTypes, reportMode);
+    if (!typeArgument) return null;
+    var signatureReference = typeArgument.signatureReference;
+    if (!signatureReference) {
+      if (reportMode == ReportMode.REPORT) {
+        this.error(
+          DiagnosticCode.Operation_not_supported,
+          typeArgumentNodes[0].range, typeArgument.toString()
+        );
+      }
+      return null;
+    }
+    return signatureReference.returnType;
   }
 
   /** Resolves a type name to the program element it refers to. */
--- a/src/resolver.ts
+++ b/src/resolver.ts
@@ -565,7 +565,7 @@ export class Resolver extends DiagnosticEmitter {
     if (!(typeArgumentNodes && typeArgumentNodes.length == 1)) {
       if (reportMode == ReportMode.REPORT) {
         this.error(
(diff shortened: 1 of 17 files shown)


dcodeIO:
Makes me wonder if we should rename

* `native<T>` to `NativeType<T>`
* `indexof<T>` to `KeyType<T>`
* `valueof<T>` to `ValueType<T>`

and maybe introduce

* `ArgumentCount<T>()`
* `ArgumentType<T>(index: i32)`
* `ThisType<T>`, maybe

As for the error, perhaps

> TS2757: Type '{0}' has no call signatures.

jtenner:
Yeah. I can work on those separately. They are all within my ability.

Are there any other improvements I could make on this feature beside the error description?

jtenner:
Also if we implement parameter count, won't that make emit() safer and faster with compile time constants?

src/resolver.ts
```diff
@@ -550,6 +551,41 @@ export class Resolver extends DiagnosticEmitter {
...
+      }
+      return null;
+    }
+    var typeArgument = this.resolveType(typeArgumentNodes[0], ctxElement, ctxTypes, reportMode);
+    if (!typeArgument) return null;
+    var signatureReference = typeArgument.signatureReference;
+    if (!signatureReference) {
+      if (reportMode == ReportMode.REPORT) {
+        this.error(
+          DiagnosticCode.Index_signature_is_missing_in_type_0,
```
dcodeIO: Index signature is for `something[xy]` overloads. Might be that the necessary error code is not in diagnosticMessages.json yet. If so, the process is to add it there and run `scripts/build-diagnostics` to update.

jtenner: Thanks for the tip. What should the first error be? The same thing?

I've submitted a commit that has both errors exactly the same.

src/resolver.ts
```diff
@@ -565,7 +565,7 @@ export class Resolver extends DiagnosticEmitter {
     if (!(typeArgumentNodes && typeArgumentNodes.length == 1)) {
       if (reportMode == ReportMode.REPORT) {
         this.error(
-          DiagnosticCode.Expected_0_type_arguments_but_got_1,
+          DiagnosticCode.Type_0_has_no_call_signatures,
```
dcodeIO: Previous error message was correct in that it reports missing type arguments on the builtin itself.

jtenner: Got it. Switched back.

jtenner:
Okay. I know I've submitted a lot of pull requests recently. I'm very flexible and can work on each one individually until everything is perfect for each one. Can you help me clean them off one by one?

dcodeIO:
Of course, these are really useful PRs. The one thing I'm unsure about, as mentioned, is whether we should rename existing type-builtins to match the `UpperCase` naming scheme used by TS. In particular, type-builtins might use the uppercase TS scheme, while code-builtins (like `parameterCount`) would use the lowercase Wasm scheme, like so:

**Type builtins**

* `NativeType<T>` (was native)
* `KeyType<T>` (was indexof)
* `ValueType<T>` (was valueof)
* `ReturnType<T>`
* `ParameterType<T,index>` (with index being an integer literal, might need special code)
* `ThisType<T>`

**Code builtins**

* `parameterCount<T>`
* `parameterName<T,index>(orIndex?: i32)` (either index argument)

With all the renaming going on, it might make sense to join all of this to a single breaking PR instead of breaking just some stuff while waiting for additional PRs. Alternatively, this PR could do the necessary renaming and we merge it first, before adding other new builtins. What would you prefer?

jtenner:
I don't think we need to break anything at all. We can simply add some aliases for each type that was mis-named and support both. For instance, it would be nice not to have to break the entire `@as-pect/core` library on a whim because you felt like the name was wrong.

As for combining all these pull requests into a single request, if you prefer it that way, let's go ahead and close these in favor of a single request with a bunch of builtins, types, and aliases.

Here are the ones I propose we get out of the way first:

- `ReturnType<T>`
- `NativeType<T>`
- `KeyType<T>` added as an alias instead
- `ValueType<T>` added as an alias instead
- `parameterCount<T>()` definitely useful

Some of the ones we should probably think about in seperate prs with some guidance:

- `parameterName()`:  I'm not this is useful. Got an example?
- `ThisType`: cannot be used meaningfully because prototype functions can't be used as expressions. We might also need `typeof` keyword support to make this worth while.
- `ParameterType<T, index>` as a `type` with a numeric literal should be possible but would require a lot more effort, but would be very useful

jtenner:
closed in favor of #739

PR #736, opened 2019-07-30
[Implement] isVoid<T>()

jtenner:
For combined use with `ReturnType`.

```ts
if (isFunction<T>()) {
  if (isVoid<ReturnType<T>>())  {
    call_indirect(func, ...props)
  } else {
    let value: ReturnType<T> = call_indirect(func, ...props);
    // do something with value
  }
}
```


jtenner:
closed in favor of #739

PR #737, opened 2019-07-30
[Implement] ParameterCount

jtenner:
Usage:

```ts
switch(ParameterCount<CallbackType>()) {
  case 1: call_indirect(func, arg0); break;
  case 2: call_indirect(func, arg0, arg1); break;
  case 3: call_indirect(func, arg0, arg1, arg2); break; // etc
}
```

Useful for `EventEmitter.prototype.emit()` and `EventEmitter.on()`.


jtenner:
closed in favor of #739

PR #738, opened 2019-07-30
[Implement] ThisType<T>

jtenner:
Problems:

1. No `typeof` keyword.
2. Inferring `this` from a function expression via property access (unless static) results in `unreachable()`
3. No testing yet because there is no way to generate a type that actually is valid for assertions anyway


jtenner:
I've given this up. I have a list of things I'm trying to already make better and I don't think I'll be able to grok it anyway

PR #745, opened 2019-08-05
Asyncify Support

willemneal:
This is an experiment to use the new [async/await pass for binaryen](https://github.com/WebAssembly/binaryen/blob/master/src/passes/Asyncify.cpp).  Essentially four functions allow you to unwind and rewind the stack allowing for "pauses" a function by storing the stack in memory.

By default all functions will be instrumented to track the stack.  So the async keyword can be used to distinguish functions to be instrumented.

This will also update the loader since the paused functions can return to the host runtime, there will need to add a run loop.

### DIFF
--- a/lib/bysyncify/as-pect.config.js
+++ b/lib/bysyncify/as-pect.config.js
@@ -1,0 +1,70 @@
+module.exports = {
+  /**
+   * A set of globs passed to the glob package that qualify typescript files for testing.
+   */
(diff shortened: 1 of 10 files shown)


dcodeIO:
Closing this PR as part of [2020 vacuum](https://github.com/AssemblyScript/assemblyscript/issues/1277) as it states that it's rather an experiment. Toyed around with a `Promise` implementation meanwhile as well, but figured that a feasible implementation requires at least closures.

PR #752, opened 2019-08-08
Fix types generated by dts script

willemneal:
Also add test for importing the compiler.

### DIFF
--- a/scripts/build-dts.js
+++ b/scripts/build-dts.js
@@ -419,6 +419,113 @@ declare type u32 = number;
 declare type usize = number;
 declare type f32 = number;
 declare type f64 = number;
+declare module "assemblyscript/std/assembly/shared/feature"{
+    // This file is shared with the compiler and must remain portable
+
+    /** Indicates specific features to activate. */
+    export const enum Feature {
+        /** No additional features. */
+        NONE = 0,
+        /** Sign extension operations. */
+        SIGN_EXTENSION = 1 << 0, // see: https://github.com/WebAssembly/sign-extension-ops
+        /** Mutable global imports and exports. */
+        MUTABLE_GLOBAL = 1 << 1, // see: https://github.com/WebAssembly/mutable-global
+        /** Bulk memory operations. */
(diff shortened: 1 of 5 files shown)


willemneal:
@dcodeIO @MaxGraey This fixes typing issues that I need fixed for other work.  Currently you can't compile an external project that depends on the types of the compiler.  If you don't like the example I can get rid of it.

scripts/build-dts.js
```diff
@@ -408,17 +408,116 @@
...
-declare type i32 = number;
-declare type isize = number;
-declare type u8 = number;
-declare type u16 = number;
-declare type u32 = number;
-declare type usize = number;
-declare type f32 = number;
-declare type f64 = number;
+const prelude = `
+declare module "assemblyscript/std/assembly/shared/feature"{
```
dcodeIO: Wondering what's necessary to make the build step actually pick these up, instead of copy & pasting here.

dcodeIO:
Would prefer separating the fixes from the example, yeah.

scripts/build-dts.js
willemneal: It's because for the sake of simplicity I extended "./node_modules/assemblyscript/std/portable.json".  I'll revert this change and just manually extend it.

willemneal:
@dcodeIO I also added a test `test:importComiler`, but didn't add it to the list of tests as it will add more time each time.  Is it possible to only trigger a test when the types file changes?

scripts/build-dts.js
dcodeIO: Would still prefer a better solution than copy & pasting the definitions below, which might change any time without anyone noticing.

willemneal: Hmm yeah in fact when this is used typescript doesn't like that `@unmaganged` isn't defined.  Is there a reason why these definitions need to be the same as those in `std`?

Do you have any ideas on how to solve this issue?  Currently I've given up on just depending on this directly and just depending on the parts of compiler that I need directly.  Especially because in my use cases I don't need binaryen.

Issue #761, opened 2019-08-14
[Bug] Returning an unmanaged class results in a retain


jtenner:
```ts
@unmanaged
class ExampleClass {}

function returnExample(): ExampleClass {
  return new ExampleClass();
}

let t = returnExample();
```

```wat
 (func $start (; 13 ;) (type $FUNCSIG$v)
  (local $0 i32)
  call $~lib/rt/tlsf/__alloc
  local.tee $0
  i32.const 160
  i32.gt_u
  if
   local.get $0
   i32.const 16
   i32.sub
   call $~lib/rt/pure/increment
  end
  local.get $0
  global.set $rc/unmanaged-return/t
 )
```

It's not accounting for the `@unmanaged Example` class.

jtenner:
```
16
Error
    at r.compileCallDirect (C:\Users\jtenner\Desktop\projects\assemblyscript\dist\assemblyscript.js:1:293524)
    at r.compileCallExpression (C:\Users\jtenner\Desktop\projects\assemblyscript\dist\assemblyscript.js:1:288764)
    at r.compileExpression (C:\Users\jtenner\Desktop\projects\assemblyscript\dist\assemblyscript.js:1:245513)
    at r.compileVariableStatement (C:\Users\jtenner\Desktop\projects\assemblyscript\dist\assemblyscript.js:1:240872)
```

Looks like it's coming from the variableStatement.

```ts
        initExpr = this.compileExpression(declaration.initializer, Type.auto,
          Constraints.WILL_RETAIN
        ); // reports
        initAutoreleaseSkipped = this.skippedAutoreleases.has(initExpr);
```
How do you determine if an `Expression` is managed?

dcodeIO:
The compiler should only insert RT calls if `relevantType.isManaged` (e.g. `instance.signature.returnType.isManaged` if a call). Maybe this check is missing somewhere.

jtenner:
I'm trying to trace it.  It's very hard.

jtenner:
I hate to be a bother, but I did find another compiler bug for sure.

```ts
static report<T>(actual: T): void {
    // get the stack trace
    Actual.stackTrace = getStackTrace();

    // if T is a reference type...
    if (isReference<T>()) {

      if (isNullable<T>()) {
        if (actual === null) {
          Actual.type = ValueType.Null;
          return;
        }
      }

      let ptr = changetype<usize>(actual);

      // if the reference is managed, retain it for later reporting
[...]

PR #762, opened 2019-08-14
Do not retain unmanaged allocations in generated ctors

dcodeIO:
potentially fixes https://github.com/AssemblyScript/assemblyscript/issues/761

### DIFF
--- a/src/compiler.ts
+++ b/src/compiler.ts
@@ -7871,16 +7871,14 @@ export class Compiler extends DiagnosticEmitter {
     //   this.b = Y
     //   return this
     // }
+    var allocExpr = this.makeAllocation(classInstance);
+    if (classInstance.type.isManaged) allocExpr = this.makeRetain(allocExpr);
     stmts.push(
       module.if(
         module.unary(nativeSizeType == NativeType.I64 ? UnaryOp.EqzI64 : UnaryOp.EqzI32,
           module.local_get(0, nativeSizeType)
         ),
-        module.local_set(0,
-          this.makeRetain(
-            this.makeAllocation(classInstance)
-          )
-        )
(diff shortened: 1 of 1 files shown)


jtenner:
I can try this later. Does the example fail in #761

jtenner:
I built the compiler, tested it against my example and I still get a rogue release problem. Something is wrong and I can't figure it out. Will investigate tomorrow.

dcodeIO:
The example works for me

```wat
 (func $empty/ExampleClass#constructor (; 12 ;) (type $FUNCSIG$ii) (param $0 i32) (result i32)
  local.get $0
  i32.eqz
  if
   i32.const 0
   i32.const 0
   call $~lib/rt/tlsf/__alloc
   (; retain in question was here ;)
   local.set $0
  end
  local.get $0
 )
 (func $empty/returnExample (; 13 ;) (type $FUNCSIG$i) (result i32)
  i32.const 0
  call $empty/ExampleClass#constructor
 )
 (func $start:empty (; 14 ;) (type $FUNCSIG$v)
  call $empty/returnExample
  global.set $empty/t
 )
```

According to rtrace `t` is leaking there, which is correct since it hasn't been `__free`d manually and is ignored by GC.

jtenner:
In this case, Yeah, it looks like the example is good but I've identified another unmanaged reference problem. My testing suite passes each test but crashes without explanation. Just need some time to get settled today and I promise to try and help.

jtenner:
Great news. I tracked down the problem. It's actually within my testing software now.

https://github.com/jtenner/as-pect/blob/master/packages/assembly/assembly/internal/report/Actual.ts#L111

This is because the reporting mechanism does not account for `@unmanaged` class references. Thanks for your help in this.

PR #769, opened 2019-08-17
Add toString methods to Number and Boolean

willemneal:
This removes IDE errors when trying use the toString method on those primitive types.


dcodeIO:
Thanks! :)

PR #789, opened 2019-08-23
Improve memory allocation for Array#push

### LINKED ISSUE #788: Array#push causes memory problems
`Array#push` calls `ensureSize`, which reallocates on every push (assuming there is no preallocated memory) https://github.com/AssemblyScript/assemblyscript/blob/12aed07918e45feced676a713ee4d5588a49f4f3/std/assembly/array.ts#L17. This causes O(n^2) memory to be allocated when doing push O(n) times.

bowenwang1996:
Double the capacity of array when reallocating to avoid quadratically allocating memory. Fixes #788

### DIFF
--- a/std/assembly/array.ts
+++ b/std/assembly/array.ts
@@ -199,8 +199,14 @@ export class Array<T> extends ArrayBufferView {
 
   push(value: T): i32 {
     var length = this.length_;
+    var align = alignof<T>();
+    var dataLength = changetype<ArrayBufferView>(this).dataLength >>> align;
     var newLength = length + 1;
-    ensureSize(changetype<usize>(this), newLength, alignof<T>());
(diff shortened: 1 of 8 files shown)


dcodeIO:
It'd be more efficient to do this on the memory allocator level. The `__realloc` implementation of TLSF for instance does this already, taking advantage of the fact that it knows how much overhead a block has anyway, so it only has to copy when absolutely necessary. Ideally, the `__realloc` impl of the stub runtime would do the same, taking advantage of alignment (block sizes are 16 + n * 16). For instance, of a block of size `8` is created, it's guaranteed to have `8` bytes overhead that it can reuse (by only updating `.rtSize`) before copying. The stub allocator can also try to resize the last segment indefinitely since it knows that everything after is free anyway.

Issue #850, opened 2019-09-19
Mutating the same TypedArray on either side of the JS or WASM fence


michaeltheory:
Hi,

Thanks for all the work on AssemblyScript -- the tooling is fantastic.

I'm aware there are other issues that ask a similar question, but unfortunately I couldn't find a specific answer, so I'm hoping for some clarification.

My search started in [this thread ](https://github.com/AssemblyScript/assemblyscript/issues/263) but what I'm finding hard to follow is that OP seems to be trying to take two steps at once, both sharing memory and packing different data sets into it.

I did notice [this comment](https://github.com/AssemblyScript/assemblyscript/issues/263#issuecomment-422137685) which appears to be exactly what I want to do. An array that can be mutated and read from on either side WASM/JS.

I followed some links [to this repo](https://github.com/kyr0/assemblyscript-js-wasm-interop-example) which sort of worked after I used what appeared to be the new API in `assemblyscript/lib/loader` with `__allocArray` and `__get(X)Array` instead. The issue here though is that `__get` appears to be creating a new typed array, or atleast view of it. I also have to call `__get` in JS every time the data has been mutated in AS, which isn't possible in my use case.

The most promising avenue I found was [this pull request](https://github.com/AssemblyScript/assemblyscript/pull/268) but I wanted to ask this before trying to piece that together, because it seems like the API might have changed in the last year since those questions were asked.

My use case is I have a 3D engine and I want to do the expensive work in WASM. For example, A `Vector3` class has a `Float32Array` with 3 floats. I would like to mutate it in JS, but then allow AS to have a reference to and be able to use those 3 floats in computation.

Could someone guide me to the best solution?

dcodeIO:
* `getArrayView`
* `getFloat32Array` etc.

returns a typed array view on the Wasm memory that can be mutated on both sides **as long as the respective managed object is alive and has not been garbage collected**, while

* `getArray`

does copy from Wasm memory to a new JS array. The reason for having both versions is that a normal `Array<T>` on the Wasm side can grow its backing buffer automatically, for example when `.push`ing to it, and any previously obtained view would become invalidated in this case (hence copying makes sense if that can happen).

dcodeIO:
For completeness, instead of storing a reference, one can also trick the GC:

```ts
export function createFloat32Array(length: i32): Float32Array {
    let array = new Float32Array(length);
    __retain(changetype<usize>(array));
    return array;
}
```

[...]

PR #857, opened 2019-09-20
Fix typed array views obtained using the loader

dcodeIO:
As discovered in https://github.com/AssemblyScript/assemblyscript/issues/850 the views created by the loader are intended to view the memory without copying, so `TypedArray#subarray` or the respective constructor must be used instead of `TypedArray#slice`, which copies.


dcodeIO:
cc @MaxGraey As a consequence, I believe the initial benchmarks done here are wrong since these involve copying overhead.

PR #861, opened 2019-09-22
Switch to GitHub Actions

dcodeIO:
This PR moves our tests from Travis-CI to GitHub Actions. Not because Travis-CI didn't serve us well, rather because our code is on GitHub anyway so we can as well make use of the closer integration.


dcodeIO:
The badge is not working currently, but I assume it will start working once merged.

dcodeIO:
Ok, let's see how this turns out.

Issue #887, opened 2019-10-05
Sharing Linear Memory/Runtime Between Modules


sampaioletti:
While evaluating AS as the basis for our new plugin system for indis.io we had been playing around with passing AS types (i.e. string/array/classes/etc) between multiple instances, similar to the question in #606. 

To be clear they goal for us differs from some of the other PRs like #278 as we aren't targeting multi-threaded at the moment, we are just trying to make it easier for plugin authors to not have to deal with the idiosyncrasies of wasm.

i have created a fork/branch with one of the implementations we have been looking at. https://github.com/sampaioletti/assemblyscript/tree/shared-rt/examples/shared-rt

In this sample we are generating a blank module with the run time, and exporting it and the rt methods to two other 'plugins' so they can share the allocater/linear mem. Leaving the plugins very lean since the rely on the imported runtime.

It works as expected with the exception of what happens to wasm data sections (memory initializes) as they will overwrite each other as they are imported.

I'm sure some other issues will show up if this is taken further, but the I'm posting it here as its no where near PR ready and was hoping for some feedback before I dig deeper.

- Is this a feasible/useful pattern for the main AS project
- What would be the best way to handle the memory initializers (a compiler flag that turns them into runtime allocations/ separation using the --memoryBase flag..etc)

Past this we have looked at some other options but this was our favorite for a first try

Thanks, great project!

dcodeIO:
That's an interesting approach. As you noted, there'll most likely be more to it, but I like the idea of making the RT a shared module. Goes into the direction of dynamic linking, though, which there isn't a good story for in Wasm yet without multiple memories.

Wondering whether something like a `--noStatic` flag to disable the generation of static data would help to get things started. Some of stdlib internals might expect static data when inlining for example. Is also likely to have a runtime cost because everything must be dynamically allocated, refcounted and such, but apart from that would be more convenient than alternatives.

Laying out memory regions for all modules in advance would also work, but I assume that'd be a somewhat painful process. Essentially, the first module would do things normally, with the second using the final value of `__heap_base` of the first as `--memoryBase` and so on, very similar to static linking. The shared RT module would then use the final `__heap_base` of the last module as its own `__heap_base` so it recognizes all the static data.

[...]

PR #890, opened 2019-10-09
Relocatable module experiment

dcodeIO:
Experimental branch investigating https://github.com/AssemblyScript/assemblyscript/issues/887 that doesn't work because we can't have `(data ...)` and `(elem ...)` offsets in the form `someConstantGlobal + someConstantValue`. See also [this comment](https://github.com/AssemblyScript/assemblyscript/issues/887#issuecomment-539930249). Pinned here for reference, even though it's not possible currently.


sampaioletti:
Just to validate this as a future option (when supported by the spec) I'll pull in a reference the the design doc

> In the future, operators like i32.add could be added to allow more expressive base + offset load-time calculations.

https://github.com/WebAssembly/design/blob/master/Modules.md#initializer-expression

sampaioletti:
After rereading https://github.com/WebAssembly/design/blob/master/DynamicLinking.md (for the 100th time) I feel like this may be possible with the current spec

I played around with it a little, and your pr is currently generating the following for two globals

```wat
(data (i32.add (global.get $__memory_base)(i32.const 8)) "\16\00\00\00\01\00\00\00\01\00\00\00\16\00\00\00a\00p\00p\001\00 \00g\00l\00o\00b\00a\00l\00")
(data (i32.add (global.get $__memory_base)(i32.const 48)) "@\00\00\00\01\00\00\00\01\00\00\00@\00\00\00t\00h\00i\00s\00 \00i\00s\00 \00a\00n\00o\00t\00h\00e\00r\00 \00g\00l\00o\00b\00a\00l\00 \00f\00r\00o\00m\00 \00a\00p\00p\001\00")

```
I dug through a couple of implementations and it appears the the data segments are additive by default, so you only need this

```wat
(data (global.get $__memory_base) "\16\00\00\00\01\00\00\00\01\00\00\00\16\00\00\00a\00p\00p\001\00 \00g\00l\00o\00b\00a\00l\00")
(data (i32.const 48) "@\00\00\00\01\00\00\00\01\00\00\00@\00\00\00t\00h\00i\00s\00 \00i\00s\00 \00a\00n\00o\00t\00h\00e\00r\00 \00g\00l\00o\00b\00a\00l\00 \00f\00r\00o\00m\00 \00a\00p\00p\001\00")
```
I've done a quick translation of the above using wat2wasm, and it actually seemed to  function correctly?

I set my first import of memory_base to 8 to mimic what you did in your generated code and it built/ran fine. I'm sure I'm missing something, but it looks promising to me.

Thoughts?

dcodeIO:
> I dug through a couple of implementations and it appears the the data segments are additive/relative by default

Afaik these are absolute, unfortunately. If your module works, I either assume because `__memory_base = 0` or it just so happens to run fine with corrupt memory.

sampaioletti:
Yeah you were correct they are absolute.. I do have good initial results merging the data statements(by hand) into one large statement with the imported offset.. But I haven't figured out if there is a way to have binaryen generate that.

dcodeIO:
That's a great idea!

dcodeIO:
Thinking one step further: It won't be possible to use the same workaround for the table since we can't "merge" table elements like memory segments. Unfortunate.

sampaioletti:
Very cool, I'll play around with a bit and see where it goes. Thanks!

sampaioletti:
So far looks quite good..but i have the issue of not really needing tables for our project (yet?) but from a memory side seems to work a treat. I'll integrate that with my original shared-runtime example and see what the next issue becomes

dcodeIO:
The table becomes important as soon as a function is passed around by reference, like when providing a callback to `Array#forEach` or similar. It's possible to write code without that ofc, but still too common to make relocatable modules a general compiler feature.

dcodeIO:
Regarding https://github.com/AssemblyScript/assemblyscript/pull/891: The compiler didn't yet relocate inlined offsets, which it does now, but leads to various issues in these cases. Accessing `staticArray[0]` for example will compile a static `@inline` string that cannot actually be inlined anymore since it's a computed offset now. That's where it becomes more complicated :)

sampaioletti:
I couldn't sleep thinking about this so I put some more time into it I do apologize this isn't compatible with your current commit, as it would have been to much work to rebase my repo (at the moment).

I am able to build, pass validation and perform the following

```wat
const someStaticStuff: i32[] = [0];

someStaticStuff;

//verifies that the ptr is > memory_base and that the strings match
@external("env","log")
declare function log<T>(obj:T,length:i32):void
//verifies that the ptr is greater than memory_base
@external("env","checkPtr")
declare function checkPtr<T>(obj:T):void
const test1="relocatable"
export function main():void{
    someStaticStuff[0]=1
    checkPtr(someStaticStuff)
    log(test1,test1.length)
    const test2="relocatable"    
    log(test2,test2.length)
    let test3=String.UTF16.encode(test2)
    log(test3,test2.length)
}
```
https://github.com/sampaioletti/assemblyscript/commit/10aca346eebd8140405b72c2905265c3842db6a3

Please have a look if you dont mind I'd like your advice on if you would prefer the way you did your last commit, if so I'll spend some time to update before I make a PR

dcodeIO:
Also still thinking how to do this. I guess a better approach to my previous is to make the relocation opaque to the module's logic so we can keep using constants and so on, with the downside that any pointer leaving the module must be relocated by the loader / RT. Last change makes it so that even if we have a `const foo = "somestring";`, the global `foo` will hold the non-relocated pointer, which is fine as long as all subsequent `load`s and `store`s add the necessary offset. The missing piece there is that a shared runtime must account for this in that a normal call to `__retain` receives a non-relocated pointer which the RT must relocate upon the whole shared memory.

sampaioletti:
I continued down the path I started to see how far I could get and was able to get it working with my shared-rt example. Take a look and see what you think..i also updated my relocatable fork if you would like to only see the changes to the compiler for relocatable. (keep in mind my relocatable is out of sync with yours)

https://github.com/sampaioletti/assemblyscript/tree/shared-rt-sp-relocatable/examples/shared-rt

I also created a DESIGN.md trying to start explaining how shared-rt works

I need to develop some tests and start looking into tables...but I feel like I'm still making progress.

sampaioletti:
https://github.com/AssemblyScript/assemblyscript/pull/890#issuecomment-540134407

I've been playing with tables and I can't get the compiler to generate more than one elem section? It always just appends to the first.. what would cause it to generate more than one line so I can test some things?

dcodeIO:
> I've been playing with tables and I can't get the compiler to generate more than one elem section? It always just appends to the first..

That's how it's currently generated, yeah. Might have something to do with how Binaryen handles this. The API there is `BinaryenSetFunctionTable` that takes an array of functions with no way yet to specify an offset. A combined elem is actually good news though, because we can make it work by implementing a start offset on the Binaryen side.

```wat
(elem (i32.const 0) $null $start:empty~anonymous|0 $start:empty~anonymous|1)
     ; ^ use __table_base
```

dcodeIO:
Last commit implements table relocation using a custom Binaryen build including https://github.com/WebAssembly/binaryen/pull/2380.

dcodeIO:
Remaining challenge is to obtain `memory_size` / `table_size` prior to instantiation. Without that, we don't know the parameters of the imported memory and table. We can't even reliably instantiate a module in some sort of sandbox to determine these since doing so already requires an imported memory and table that must both be large enough. I don't see a good solution to this yet.

dcodeIO:
One way to expose this information might be to hijack the proposed [dylink](https://github.com/WebAssembly/tool-conventions/blob/master/DynamicLinking.md#the-dylink-section) section (while it's not available) by adding an API to Binaryen that lets us emit arbitrary custom sections, and use it by means of [WebAssembly.Module.customSections](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WebAssembly/Module/customSections) today.

dcodeIO:
Last commit hijacks the dylink section as mentioned using latest Binaryen. Removes the `memory_size` and `table_size` exports in favor of evaluating the section, which we can do before instantiation now. Should theoretically be sufficient now to build a working dynamic linker.

sampaioletti:
Great work, I did some mods to the compiler test runner to be able to run multiple modules in one test (via a modules section in the test .json config) so I'll pull these in and see if I can build a few tests showing some possible use cases.

Sam

________________________________

dcodeIO:
So, regarding a shared runtime, it appears that with each module having its own set of runtime types along the respective compiler generated logic to visit these, that garbage collection has to remain local to the module at least partly, requiring changing a few APIs to make this possible. Affected APIs are `__release` and `__collect`, which call `__visit_members`, and `__visit` itself. The difference to the static runtime would be that there's another argument when these are called on the shared runtime, a reference to `__visit_members`, and that calling `__collect` globally would do a collection within each module, in turn triggering a collection with the respective visitor. At least that's what I figured so far.

dcodeIO:
Last commit has a shared runtime concept with an [overview here](https://github.com/AssemblyScript/assemblyscript/blob/c6cdd20b503ef4702a6de0dac0ee9be52d82c89c/lib/asrt/README.md). It still remains to be seen if this works, though :)

sampaioletti:
@dcodeIO thats looking real good, I still need to explore a few things and I would like to write some tests that support our use cases. As I mentioned before I reworked the compiler test runner to allow child modules to be instantiated. Is that a direction you would like to go, or are we flirting with more of an 'integration' test and should make a separate runner i.e. `test:integration -- shared-rt` or... 

I at least want to do my internal testing in a way that I could eventually (without much rework) contribute them if they benefit the project

dcodeIO:
This draft aged surprisingly well, and it might still be feasible to return to relocation eventually. Keeping it open just in case. ([see](https://github.com/AssemblyScript/assemblyscript/issues/1277))

HerrCai0907:
clean outdated draft PR

PR #891, opened 2019-10-09
tests\compiler\relocatable test offsets for globals

sampaioletti:
Instantiating the data segments to memory is working correctly for what I can tell, but need to add offset logic to the index values

for example from the relocatable.untouched.wat fixture that I updated

```wat
  ;;line 43
 (func $relocatable/main (; 5 ;) (type $FUNCSIG$v)
  (local $0 i32)
  i32.const 80 ;; offset from mem_base
  call $~lib/rt/stub/__retain
```

should be

```wat
  ;;line 43
 (func $relocatable/main (; 5 ;) (type $FUNCSIG$v)
  (local $0 i32)
  global.get $__memory_base
  i32.const 80
  i32.add
  call $~lib/rt/stub/__retain
```

I'll look into it...but may be above my paygrade (:

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -16,6 +16,7 @@ under the licensing terms detailed in LICENSE:
 * Willem Wyndham <willem@cs.umd.edu>
 * Bowen Wang <bowen@nearprotocol.com>
 * Emil Laine <laine.emil@gmail.com>
+* Sam Paioletti <sampaioletti@gmail.com>
 
 Portions of this software are derived from third-party works licensed under
 the following terms:
--- a/tests/compiler/relocatable.js
+++ b/tests/compiler/relocatable.js
@@ -3,7 +3,16 @@ exports.preInstantiate = function(imports, exports) {
   // of env.memory_base yet, hence we need to import a suitable memory as well:
   imports["env"] = {
     "memory": new WebAssembly.Memory({ initial: 2 }),
-    "memory_base": 65536,
-    "table_base": 100
+    "memory_base": 2,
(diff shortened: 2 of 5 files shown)


sampaioletti:
Sorry I guess it could be considered bad form to PR a known failing test...I thought it might be easier than just explaining it.

sampaioletti:
@dcodeIO just looked through your commit, i was going about it a little differntly i split out the i32 functions

```js
//src/module.ts:528
//new func when a ptr is needed rather than just a val
  i32Ptr(value: i32): ExpressionRef{
    var out = this.lit;
    _BinaryenLiteralInt32(out, value);
    return this.relocMem(_BinaryenConst(this.ref, out));
  }
  i64Ptr(valueLow: i32, valueHigh: i32 = 0): ExpressionRef{
    var out = this.lit;
    _BinaryenLiteralInt64(out, valueLow, valueHigh);
    return this.relocMem(_BinaryenConst(this.ref, out));
  }
//old funcs
  i32(value: i32): ExpressionRef {
    var out = this.lit;
    _BinaryenLiteralInt32(out, value);
    return _BinaryenConst(this.ref, out);
  }

  i64(valueLow: i32, valueHigh: i32 = 0): ExpressionRef {
    var out = this.lit;
    _BinaryenLiteralInt64(out, valueLow, valueHigh);
    return _BinaryenConst(this.ref, out);
  }
```
and was going around and replacing it where i needed to.....but i'll start testing your new code with mine...thanks

PR #901, opened 2019-10-11
Added stdout and stderr

willemneal:


dcodeIO:
Not yet sure how to deal with this because the compiler uses stdout to output the resulting binary or text format if no file name is given, and logging to stdout will interfere with that. Perhaps the `log` method should always print to stderr, and we can add `stdout` and `stderr` properties to the transform in case someone really wants to hijack these?

willemneal:
Yeah I'd just document that they should prefer to use stderr.  My guess is that if someone wants to dive this deep they'd understand the risk ;-)

dcodeIO:
Closing in favor of https://github.com/AssemblyScript/assemblyscript/pull/892

PR #908, opened 2019-10-15
Added missing type declaration for 'u64'

Iainmon:
The type 'u64' was missing a type declaration and making it hard for VSCode to know what 'u64' meant.

### DIFF
--- a/dist/assemblyscript.d.ts
+++ b/dist/assemblyscript.d.ts
@@ -6,6 +6,7 @@ declare type isize = number;
 declare type u8 = number;
 declare type u16 = number;
 declare type u32 = number;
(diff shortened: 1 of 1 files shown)


dcodeIO:
Since the compiler itself is portable and doesn't utilize `u64` for that reason, I wonder why this change is necessary. Could you explain which issue this is intended to solve?

Also note that `dist/assemblyscript.d.ts` is a generated file, and any changes will be overwritten on build again.

Iainmon:
My apologies, I didn't know that `dist/assemblyscript.d.ts` would be overwritten. I wanted to add `u64` for editor support.

PR #914, opened 2019-10-16
Dylink example

sampaioletti:
Here is an example with dynamic linking. I had to revise some of the relocatable branch, specifically after playing with it I was having better results when i went to absolute pointers.

The example is working, including passing strings and classes between relocatable modules, which really makes me excited about the viability of this.

Couple thoughts/notes
- only added support for static strings (missing arrays)
- toyed with adding a `@dynlib` decorator instead of using `@external`, didn't end up including it, but i think it may be how it should be done in the future, so we can implement the rest of the WA dynlink proposal and allow some additional compile time benefits
- currently 'leaking' but that may be my lack of understanding of rtrace
- I think a lot of the logic in the linker could be put into a AS module, to reduce the amount of work a linker has to do


MaxGraey:
@dcodeIO Could you look at this?

sampaioletti:
Gents, just fyi I won't be offended if it gets deleted, it was just an experiment but I am still using it in a production environment. More or less just wanted to share ideas, it's far from a finished design/implementation.

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

dcodeIO:
Closing this PR as part of [2020 vacuum](https://github.com/AssemblyScript/assemblyscript/issues/1277) as it states it's rather an experiment. But I'd be interested to learn more about the problems with the relocatable branch that you encountered, if you still remember.

PR #941, opened 2019-11-07
Publish preparations

dcodeIO:
This PR contains the initial preparation work to start publishing to npm. Idea is:

* Create a `release` branch that is always up to date with the tip of our releases
* Once a day
  * Merge `master` into `release`
  * Build and commit dist files
  * Obtain the current version from `package.json`
  * Check if the respective tag (with a prepended "v") does exist
    * If it exists, make a nightly of the form `v$VERSION-nightly.YYYYMMDD`
    * If it does not exist, make a release of the form `v$VERSION`
  * Push dist files back to `release` and create the tag
  * Publish to npm (not yet enabled)
  * Publish to GH package registry (not yet enabled)

This gives us:

* To bump the version, all we have to do is edit `package.json`
* If the dist files at the tip of our releases are the same as those we just built, the job fails (commit is empty) so we don't publish unnecessary duplicates.

With caveats being:

* Publishing happens exactly once a day, so it might take a while till these end up on npm etc.
* Installing `AssemblyScript/assemblyscript` via npm will no longer work once dist files are removed from master (currently still there)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,5 +1,9 @@
 name: CI
-on: [push, pull_request]
+on:
+  push:
+    branches:
+    - master
+  pull_request:
 jobs:
   check:
     name: "Check preconditions"
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,76 @@
+name: Publish
+on:
+  schedule:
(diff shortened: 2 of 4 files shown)


dcodeIO:
Alright, let's see how this turns out. If it works as expected, there should be a `v0.8.0` tag in about an hour.

PR #953, opened 2019-11-13
Add concrete SIMD types

dcodeIO:
This is an experiment to add the more concrete SIMD types

* i8x16
* u8x16
* i16x8
* u16x8
* i32x4
* u32x4
* i64x2
* u64x2
* f32x4
* f64x2

in addition to (or potentially replacing) the v128 type.

I started with the goal to support `someVector[0]` compiling to `i32x4.extract_lane` / `i32x4.replace_lane` etc. and to provide `someVector * someVector` operator overloads, but then realized that especially the latter can't be expressed in TypeScript since `someVector` must be of type `number` so it doesn't trigger TS errors, but after the binary operation loses the index signature, again leading to a TS error.

So I stopped there for now, but pinning it here in case someone has an idea. Maybe it's not even necessary, don't know.


dcodeIO:
Closing this PR as part of [2020 vacuum](https://github.com/AssemblyScript/assemblyscript/issues/1277) as it hit major road blocks and should be re-evaluated once we have concrete information on how SIMD will be used in the wild. I expect that something like this will eventually be brought up, including suggestions on how it should be designed most likely in a different way than proposed here.

PR #973, opened 2019-11-23
Clean up portable memory definitions

dcodeIO:
Related: https://github.com/AssemblyScript/assemblyscript/pull/971

This PR cleans up the portable memory definitions by, well, removing them completely. These have been of very little use anyway because there isn't a good memory story in JS. The compiler now uses bindings to Binaryen's `_malloc` etc. directly, which is much clearer. There's more to consider once we want to compile the compiler to Wasm, but leaving that for later.


src/glue/js/binaryen.d.ts
```diff
@@ -10,3 +10,8 @@ declare namespace binaryen {
   }
   function wrapModule(ptr: number): Module;
 }
+
+/** Stores an 8-bit value to Binaryen memory. */
+declare function store<T extends u8>(ptr: usize, value: T): void;
+/** Loads an 8-bit value from Binaryen memory. */
+declare function load<T extends u8>(ptr: usize): T;
```
dcodeIO: These are somewhat awkward still. Ideally, Binaryen would expose a `_store8`, `_store32` etc. along `_malloc` and friends so we don't need any JS glue when importing and interacting with Binaryen's memory.

dcodeIO: There is [this Binaryen PR](https://github.com/WebAssembly/binaryen/pull/2476) now suggesting a set of Binaryen exports named after the respective Wasm instructions so we can get rid of the store and load stuff as well. Instead of mixing in suboptimal functions, we'd then be able to simply call Binaryen's `_i32_store8` just like we call `_malloc`.

dcodeIO:
With https://github.com/WebAssembly/binaryen/pull/2476 merged and Binaryen updated, the last commit introduces the concept of a portable d.ts.

What this does is if compiling with tsc, it will pick up the js file next to the d.ts normally, while asc ignores the js file and uses the d.ts as if it was a normal AssemblyScript file, in turn making a WebAssembly import of all the `declare`s. As such, when compiling to WebAssembly, we can provide binaryen.js (no matter if compiled as JS or Wasm) as an import named `binaryen` with no need for special glue code, with the JS engine acting as a linker for Wasm exports to Wasm imports directly.

PR #984, opened 2019-12-01
CI: run compiler tests on Windows

jhwgh1968:
Tests the compiler on Windows as a separate CI job. This is designed to catch "bash-isms" that could otherwise sneak in, rather than provide full Windows coverage. Thus it only tests the stable node version.

On Linux, nvm is used to determine the current stable version. Since that does not work on Windows, the job uses [nvm-windows](https://github.com/coreybutler/nvm-windows) as a substitute. That project is written in Golang, so a pre-compiled binary is downloaded and invoked directly.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,6 +55,31 @@ jobs:
       run: npm run build
     - name: Test distribution
       run: npm test
+  test-windows:
+    name: "Test compiler on Windows with node: node"
+    runs-on: windows-latest
+    needs: check
+    steps:
+    - uses: actions/checkout@v1.0.0
+      with:
+        node-mirror: https://nodejs.org/download/v8-canary/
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -55,6 +55,31 @@ jobs:
...
     - name: Test distribution
       run: npm test
+  test-windows:
+    name: "Test compiler on Windows with node: node"
+    runs-on: windows-latest
+    needs: check
+    steps:
+    - uses: actions/checkout@v1.0.0
+      with:
+        node-mirror: https://nodejs.org/download/v8-canary/
```
dcodeIO: Looks like these two lines don't do anything

dcodeIO:
LGTM, great to have this in CI. Just a minor comment above.

dcodeIO:
Thanks!

Issue #983, opened 2019-11-30
Consider renaming __start to _start


dcodeIO:
The `--explicitStart` option currently yields an export named `__start`, but it would be nice to align the name with WASI's `_start` function. There's still [some discussion](https://github.com/WebAssembly/WASI/issues/19) going on on the WASI side regarding that name, though, so we might have to reconsider something else later.

dorin131:
Is this a good first issue?
Could I give it a try?

dcodeIO:
I suspect you'd be disappointed of this as a first issue. It's mostly renaming at a couple places, but it's also a breaking change so it might make sense to collect more feedback before going for it.

dorin131:
Do we want then to wait and see what decision WASI comes to and take it from there?

In the meantime, could you point me to a good first issue or task please?

PR #994, opened 2019-12-07
Rename __start to _start

dcodeIO:
This renames the explicit start function we generate when `--explicitStart` is used from `__start` to `_start` to align with WASI. Essentially, one would then always use `--explictStart` when building a WASI module and it just works™. Should probably increment the version to 0.9 as well since this is a breaking API change.

fixes https://github.com/AssemblyScript/assemblyscript/issues/983


jtenner:
This is a breaking change for as-pect users.

dcodeIO:
This made it into the latest nightly without a new version tag yet, so that should be good to test-drive for when we slice the 0.9.

PR #998, opened 2019-12-08
Fix source contents in source maps

dcodeIO:
As reported in https://github.com/AssemblyScript/assemblyscript/issues/996 there is something wrong with source maps currently. The issue reported there seems to be caused by using the wrong key inside of the JSON-like `.map` file for source contents, namely `sourceContents` instead of the correct `sourcesContent`.

This PR renames the key to the proper one, but I actually have no idea how this did work in the past, which it definitely did when I last tested. Makes me wonder if browsers used to also accept `sourceContents` up to some point, and removed that.

Another issue I had locally was that, when just `--sourceMap` was used without an absolute path, we are setting the `sourceMappingURL` section to `thefile.wasm.map`, assuming that the browser will load this relative to `thefile.wasm`, even if in a subdirectory. This mechanism seems to be broken in my local tests, so when I do

```js
WebAssembly.instantiateStreaming(fetch("./somedir/thefile.wasm"), ... )...
```

my version of Chrome always attempts to fetch `/thefile.wasm.map` at the root instead of `/somedir/thefile.wasm.map`. Again, I am puzzled how this did work in the past and doesn't anymore, but I don't have a fix for this one. The compiler emits `sourceMappingURL="./thefile.wasm.map"` now, which at least indicates the intention, even though that doesn't change anything functionally.

The PR also makes it so that we don't generate another `assemblyscript://` root anymore but simply use the Wasm file's name as the `sourceRoot`. See the attached screenshot on how it's structured now.

smap

smapff

### DIFF
--- a/cli/asc.js
+++ b/cli/asc.js
@@ -67,9 +67,6 @@ exports.version = exports.isBundle ? BUNDLE_VERSION : require("../package.json")
 /** Available CLI options. */
 exports.options = require("./asc.json");
 
-/** Common root used in source maps. */
-exports.sourceMapRoot = "assemblyscript:///";
-
 /** Prefix used for library files. */
 exports.libraryPrefix = assemblyscript.LIBRARY_PREFIX;
 
@@ -767,16 +764,17 @@ exports.main = function main(argv, options, callback) {
 
     // Write binary
     if (args.binaryFile != null) {
+      let basename = path.basename(args.binaryFile);
       let sourceMapURL = args.sourceMap != null
         ? args.sourceMap.length
           ? args.sourceMap
-          : path.basename(args.binaryFile) + ".map"
+          : "./" + basename + ".map"
         : null;
 
       let wasm;
       stats.emitCount++;
       stats.emitTime += measure(() => {
-        wasm = module.toBinary(sourceMapURL)
+        wasm = module.toBinary(sourceMapURL);
       });
 
       if (args.binaryFile.length) {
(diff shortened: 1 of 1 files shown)


dcodeIO:
Seems that relative source maps as mentioned above still work as before in Firefox, but don't in Chrome. Potentially a Chrome bug.

echoloyuk:
Hi, I still cannot get the sourcemap file in chrome dev tools. the screenshot below shows chrome doesn't load the map file. Btw, my chrome dev tools shows quiet different from yours. The chrome of mine loading wasm is an empty file with an hash code like filename. Is there are sth wrong of my working? Here is my step:

1. `mkdir` an folder names: 'as2'
2. execute: `cd as2` to get in the folder
3. execute: `npm init` to generate a package.json
4. execute: `npm install --save-dev assemblyscript` as the document shows.
5. execute: `npx asinit .` to generate an assemblyscript directory structure.
6. execute: `npm run asbuild` as the doc shows to build wasm files.\
7. add index.html in root path. code shows below.
8. I installed http-server to run a static server.
9. run http-server and type: 127.0.0.1:8080 in my chrome.
10. I can get the result in console, which is means wasm is worked, but I cant set debug point in dev tools, cause the map file is empty.

index.html:

```
<!DOCTYPE html>
<html lang="en">
<head>
  <meta charset="UTF-8">
  <meta name="viewport" content="width=device-width, initial-scale=1.0">
  <meta http-equiv="X-UA-Compatible" content="ie=edge">
  <title>Document</title>
</head>
<body>
  <script src="browser.js"></script>
</body>
</html>
```

browser.js

```
(async () => {
  const importObject = {
    env: {
      abort(_msg, _file, line, column) {
        console.log(_msg, _file, line, column);
      }
    }
  };
  const obj = await WebAssembly.instantiateStreaming(
    fetch("build/untouched.wasm"),
    importObject
  );
  const exports = obj.instance.exports;
  console.log(exports.add(8, 10));
})();
```

index.ts

```
// The entry file of your WebAssembly module.

  return a + b;
}
```

my chrome dev tools screenshot:

image

Is there sth wrong of my step? chrome version: 79.0.3945.117

echoloyuk:
And I try to paste the wasm and wasm.map file to the root, and change the load path to the root, it still the same.

dcodeIO:
Are you using the latest nightly or master?

echoloyuk:
Yes, I `npm i assemblyscript@0.8.1-nightly.20200114` and still the same

Issue #1000, opened 2019-12-10
[Feature] Add Table Export Option


willemneal:
Currently there is a function for exporting the table:
https://github.com/AssemblyScript/assemblyscript/blob/20e882c88586e79b4038e47d013aa7ddd2c692b5/src/module.ts#L1069)

but no way to access this through the cli API.

jtenner:
Context:

Aspect could benefit from inspecting the function table at testing time to get the name of the function and display it. 

Yes. This causes binaryen to avoid optimizing and Inlining other functions.

No. This isn't game breaking for aspect users because it's used for debugging purposes (and aspect by default doesn't use any optimizations at all anyway)

jtenner:
Suggestion:

- add `--exportTable` flag

PR #1001, opened 2019-12-10
Add an option to export the function table

dcodeIO:
Adds an `--exportTable` option to export the function table to the host. Background: We are currently not exporting it by default (unlike memory) do allow Binaryen to "directize" indirect calls, which exporting the table prohibits since the table can be modified externally.

fixes https://github.com/AssemblyScript/assemblyscript/issues/1000


cli/asc.json
```diff
@@ -138,7 +138,12 @@
     "default": 0
   },
   "importTable": {
-    "description": "Imports the function table instance provided by the embedder.",
+    "description": "Imports the function table provided by the embedder.",
+    "type": "b",
+    "default": false
+  },
+  "exportTable": {
+    "description": "Exports the function table.",
```
willemneal: Perhaps mention that it will be at the top level namespace and the other imports are in the `env` namespace.

dcodeIO: Exports don't have namespaces, like imports, but I can mention that it becomes exported as `table`. Is that what you meant?

willemneal: I thought that you could do the following: `@external("env", "foo")`
So that the instance will have the namespace `env`.  I didn't mean the typescript namespace, rather just `instance.env.table` and `imports.env.table`.

dcodeIO: Yeah, one can do that on imports, which have a module name and a name, but exports just have a name without nesting. As such, there's no `exports.env.table` by design, only our mangling like `exports["env.table"]` as the loader understands it.

PR #1002, opened 2019-12-11
[Feature] Implement TypedArray#set with tests

jtenner:


jtenner:
Tests are probably naive. Could be made better, but it appears to be functional. Any help/guidance would be appreciated.

std/assembly/typedarray.ts
```diff
@@ -1718,3 +1762,41 @@ function WRAP<TArray extends ArrayBufferView, T>(buffer: ArrayBuffer, byteOffset
...
+  let sourceDataStart = source.dataStart;
+
+  // if the types align and match, use memory.copy() instead of manual loop
+  if (isInteger<T>() == isInteger<U>() && alignof<T>() == alignof<U>()) {
+    memory.copy(targetDataStart, sourceDataStart, source.byteLength);
+  } else {
+    let count = source.length;
+    for (let i = 0; i < count; i++) {
+      let targetStoreOffset = <usize>i << alignof<T>();
+      let sourceLoadOffset = <usize>i << alignof<U>();
```
MaxGraey: `targetStoreOffset` and `sourceLoadOffset` better move inplace (inlined) inside `store`

jtenner: Won't this just get optimized by binaryen?

MaxGraey: I guess not necessarily.

tests/compiler/std/typedarray.ts
```diff
@@ -693,3 +693,47 @@ testArrayWrap<Int64Array, i64>();
...
+    if (vala != valb) {
+      trace(nameof<T>(), 3, i, <f64>vala, <f64>valb);
+      assert(false);
+    }
+  }
+}
+
+let setSource1: i32[] = [1, 2, 3];
+let setSource2: f32[] = [4, 5, 6];
+let setSource3: f64[] = [Infinity, NaN, -Infinity];
```
MaxGraey: I guess need add `i64[]` and typed array sources as well

std/assembly/typedarray.ts
jtenner: The `if` statement is inlined because the conditions are statically known at runtime.

tests/compiler/std/typedarray.ts
jtenner: Okay. I'll add a few more.

std/assembly/typedarray.ts
MaxGraey: I mean use `load(ptr + (i << alignof<T>()))` instead  `load(ptr + cachedStoreOffset)`. That's that case when caching is not helps

jtenner: Oh that makes a lot of sense. No problem!

std/assembly/typedarray.ts
```diff
@@ -1718,3 +1762,59 @@ function WRAP<TArray extends ArrayBufferView, T>(buffer: ArrayBuffer, byteOffset
...
+              // @ts-ignore: cast to T is valid for numeric types here
+              ? select<T>(0, select<T>(255, <T>value, value > 255), value < 0)
+              : 0
+          );
+        } else {
+          let value = load<U>(sourceDataStart + (<usize>i << alignof<U>()));
+          store<T>(
+            targetDataStart + (<usize>i << alignof<T>()),
+            // @ts-ignore: cast to T is valid for numeric types here
+            select<T>(0, select<T>(255, <T>value, value > 255), value < 0)
```
MaxGraey: Suggestion:
```ts
if (sizeof<T>() == 4) {
  ...
  ~(<i32>value >> 31) & (((255 - <i32>value) >> 31) | value)
} else {
   ...
   ~(<i64>value >> 63) & (((255 - <i64>value) >> 63) | value)
}
```

std/assembly/typedarray.ts
```diff
@@ -1718,3 +1762,59 @@ function WRAP<TArray extends ArrayBufferView, T>(buffer: ArrayBuffer, byteOffset
...
+      // if TArray is Uint8ClampedArray, then values must be clamped
+      if (target instanceof Uint8ClampedArray) {
+        if (isFloat<U>()) {
+          let value = load<U>(sourceDataStart + (<usize>i << alignof<U>()));
+          store<T>(
+            targetDataStart + (<usize>i << alignof<T>()),
+            isFinite<U>(value)
+              // @ts-ignore: cast to T is valid for numeric types here
+              ? select<T>(0, select<T>(255, <T>value, value > 255), value < 0)
+              : 0
```
MaxGraey:
```ts
<T>max<U>(0, min<U>(255, value))
```

std/assembly/typedarray.ts
jtenner: This is all set!

std/assembly/typedarray.ts
jtenner: This optimization was a big deal! Thanks for the help.

tests/compiler/std/typedarray.ts
```diff
@@ -693,3 +693,78 @@ testArrayWrap<Int64Array, i64>();
...
+
+function testTypedArraySet<T extends ArrayBufferView>(): void {
+  let setSource4: Int64Array = new Int64Array(3);
```
MaxGraey:
```suggestion
  let setSource4 = new Int64Array(3);
```

tests/compiler/std/typedarray.ts
```diff
@@ -693,3 +693,78 @@ testArrayWrap<Int64Array, i64>();
...
+
+  let b = new Int64Array(4);
+  b[0] = -10;
+  b[1] = 100;
+  b[2] = 10;
+  b[3] = 300;
+
+  let c = new Int32Array(2);
+  c[0] = 300;
+  c[1] = -1;
```
MaxGraey: Also add plz `let d = new Uint32Array` and `let e = new Int16Array`

jtenner: Please check now? Thanks!

MaxGraey: looks good

jtenner:
Anything else I can do to complete this pull request?

MaxGraey:
Looks good to me

std/assembly/typedarray.ts
```diff
@@ -1718,3 +1762,70 @@ function WRAP<TArray extends ArrayBufferView, T>(buffer: ArrayBuffer, byteOffset
...
+  if (isReference<U>()) {
+    ERROR(E_NOTIMPLEMENTED);
+  }
+
+  // Uncaught RangeError: offset is out of bounds
+  if (offset < 0) throw new RangeError(E_INDEXOUTOFRANGE);
+  if (source.length + offset > target.length) throw new RangeError(E_INDEXOUTOFRANGE);
+
+  // if the types align and match, use memory.copy() instead of manual loop
+  if (isInteger<T>() == isInteger<U>() && alignof<T>() == alignof<U>()) {
```
dcodeIO: What about floats?

MaxGraey: Hah, I also asked this. So if both integers equality will true, but if both floats this equality also true (`false == false` => true).

dcodeIO: Ah, very nice!

std/assembly/typedarray.ts
```diff
@@ -1718,3 +1762,70 @@ function WRAP<TArray extends ArrayBufferView, T>(buffer: ArrayBuffer, byteOffset
...
+      // if TArray is Uint8ClampedArray, then values must be clamped
+      if (target instanceof Uint8ClampedArray) {
+        if (isFloat<U>()) {
+          let value = load<U>(sourceDataStart + (<usize>i << alignof<U>()));
+          store<T>(
+            targetDataStart + (<usize>i << alignof<T>()),
+            isFinite<U>(value) ? <T>max<U>(0, min<U>(255, value)) : 0
+          );
+        } else {
+          let value = load<U>(sourceDataStart + (<usize>i << alignof<U>()));
```
dcodeIO: Wondering if we'll get more sign-extension/masking here than necessary, and if so whether this can be avoided by typing the intermediate local `i/u32`. Like, if `value` is typed `u16` and we cast that to `i8`, the explicit conversion will wrap, even though the `store8` would already wrap.

MaxGraey: You mean `store<u8>(<u8>~( .... ))` is unnecessary and just need `load<u8>(~(... ))`?

dcodeIO: Mostly relevant if `T` and `U` have different sizes, which Wasm has distinct instructions for when loading and storing. The `store<T>` builtin is more of a `store<T>(<U>...)`. It evaluates the argument type in order to pick the right instruction, which implicitly wraps the value. So, if we explicitly convert `<T>someU` that might already be implicit in the store, and any explicit wrapping being unnecessary. Didn't look at the fixtures yet.

jtenner: Happy to make changes for that. Just want to make sure I understand it before I pull changes from master and make a qualitative change to the code.

Can you help me understand this?

dcodeIO: I believe what happens in the `T=i8, U=i16` case is that on `value = load<i16>(...), the compiler remembers that `value` is `i16` and wrapped (because the load returns a properly wrapped value). Then, `min<i16>(some, thing)` requires properly wrapped values to determine the minimum, and fortunately it's already wrapped so nothing to do there. After that, the value is cast to `<T>` (`i8` here), which wraps the i16 to an i8 discarding garbage bits (explicit cast does this), but that's unnecessary since what follows is a `store<i8>(...)` translating into an `i32.store8` that already discards garbage bits.

One first thing to test can be to simply remove the `<T>` at the start of the store argument, and see what happens.

Wow. @dcodeIO that was an incredible find. A typecast causing an optimization problem like that seems so obscure, I can't believe it.

Anyway, this should be fixed and brought current with master!

tests/compiler/std/typedarray.ts
dcodeIO: Came here looking for an i16->i8 test. Might make sense to have a signed and unsigned small int here, a signed and unsigned normal i32 and a signed and unsigned i64.

std/assembly/typedarray.ts
```diff
@@ -1718,3 +1762,70 @@ function WRAP<TArray extends ArrayBufferView, T>(buffer: ArrayBuffer, byteOffset
...
+          }
+        }
+        // if U is a float, then casting float to int must include a finite check
+      } else if (isFloat<U>() && !isFloat<T>()) {
+        let value = load<U>(sourceDataStart + (<usize>i << alignof<U>()));
+        // @ts-ignore: cast to T is valid for numeric types here
+        store<T>(targetDataStart + (<usize>i << alignof<T>()), isFinite<U>(value) ? <T>value : 0);
+      } else {
+        // @ts-ignore: cast to T is valid for numeric types here
+        store<T>(targetDataStart + (<usize>i << alignof<T>()), <T>load<U>(sourceDataStart + (<usize>i << alignof<U>())));
```
dcodeIO: Looks like this is the int->int case, in which case the `<T>` is unnecessary for the same reason. I believe the tests currently do not catch this due to missing small and long int cases.

jtenner:
```
trace("hit!");
store<T>(targetDataStart + (<usize>i << alignof<T>()), <T>load<U>(sourceDataStart + (<usize>i << alignof<U>())));
```

This is the line in question yes?

```
- instantiate untouched OK (33.971 ms)
  716 allocs, 716 frees, 2077 increments, 2077 decrements

  [fun] __start = function 674() { [native code] }
  [obj] memory = [object WebAssembly.Memory]

  [start]
  trace: Hit!
  trace: Hit!
  trace: Hit!
...
```

This line is hit quite regularly. I will simply remove the offending `<T>` and see what happens.

dcodeIO: Is it being hit with small int types for both T and U?

```ts
if (sizeof<U>() < 4) trace("hit U!");
if (sizeof<Z>() < 4) trace("hit Z!");
```

jtenner: Okay. I have evidence that this optimization isn't too necessary, but it does seem to help.

image

This is only in the `untouched` version, and the optimized version remains unchanged.

I had to change it so that the cast happens if the conversion is from integer to float, however, by adding another `else if` branch to take advantage of this optimization.

dcodeIO: The `i64` to native `i32` (or smaller) and vice-versa cases are interesting for the optimizer because it can potentially remain on i32 or i64 exclusively, without a `.wrap_i64` instruction for example.

Other interesting cases are `i8` -> `i32`, `u8` -> `i16`, `u16` -> `i8` because these involve sign extension/wrapping. That's what I'm trying to get to, as these aren't currently present in the tests.

tests/compiler/std/typedarray.ts
jtenner: Please forgive me for finally getting around to this. I added a few more tests and actually found an edge case involving `Uint8ClampedArray`.

I would be happy to add some more tests if you'd like! I just need to get around to them tomorrow.

Can I get some feedback on the ones I put in?

std/assembly/typedarray.ts
jtenner: I removed the offending `<T>`, and this should at least be resolved from an implementation standpoint, if not from the testing standpoint. See above!

Can we mark this conversation as resolved?

std/assembly/typedarray.ts
jtenner: Okay. I'll talk with @MaxGraey to get the test cases all covered.

tests/compiler/std/typedarray.ts
dcodeIO: Looks good now with the additional Uint8Array and Int16Array cases.

PR #1009, opened 2019-12-16
Update Binaryen

dcodeIO:
This PR updates Binaryen to latest, most notably accounting for

* Creation and expansion of multi-value types
* Removal of function types (API now takes multi-value types)
* Required type parameter when creating `local.tee`s (currently polyfilled to the old behavior)

in turn also removing our own Binaryen-like signature handling. Signature changes are responsible for the huge fixture diff.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -25,9 +25,9 @@
       }
     },
     "@types/node": {
-      "version": "12.12.7",
-      "resolved": "https://registry.npmjs.org/@types/node/-/node-12.12.7.tgz",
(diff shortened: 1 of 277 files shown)


jtenner:
Multivalue returns are going to open up the *world* for writing very efficient parsers.

willemneal:
And can make iterators better too!  Currently `iterator.next()` returns an object like `{done: false, value: 1}`, which would mean allocating a new object each iteration.  I got around this in #862 by using the iterator itself as the return value.  However, now we can just return the value and the done boolean!

Also tuples!!!

dcodeIO:
Don't get too excited, as we don't have multi-value types yet. This is merely initial preparation on the Binaryen side. Also added a little test that doesn't validate at this point, but can be used to check in the future.

PR #1011, opened 2019-12-18
Change type definitions for `unreachable` to make typescript recognize it as an end point

vladimir-tikhonov:
This will help in the following scenarios:

```typescript
function bar(): bool {
  unreachable();
  return false; // typescript will report "Unreachable code detected"
}

function getImpl(implId: i32): Impl {
  switch ( implId ) {
    case 1: return new Impl1();
    case 2: return new Impl2();
    default: unreachable();
  }
  // typescript won't complain about missing return statement here.
}

```


dcodeIO:
Seems reasonable, thanks!

PR #1032, opened 2019-12-30
Fix premature layout in resolveClass recursion

dcodeIO:
As reported in https://github.com/AssemblyScript/assemblyscript/issues/1030 there is an issue with premature layout when resolving classes. This is due to the recursion missing the case where a base class's members (not their base) again trigger resolving a derived class with a base class that isn't fully layed out yet (around 4 corners?), in turn prematurely laying out the derived class with missing base fields.

This is an attempt to solve this by yielding within the recursion (but never on the first API call) instead of doing an improper layout, then triggering finalization as soon as the base class is fully layed out, essentially abusing the execution stack to roll back the recursion (by adding another recursion). I must admit though that this solution feels kinda wild west and there's a chance that it's still not perfect.

fixes https://github.com/AssemblyScript/assemblyscript/issues/1030

### DIFF
--- a/src/builtins.ts
+++ b/src/builtins.ts
@@ -850,7 +850,7 @@ export function compileCall(
         }
         offset = (<Field>field).memoryOffset;
       } else {
-        offset = classType.currentMemoryOffset;
+        offset = classType.nextMemoryOffset;
       }
       if (compiler.options.isWasm64) {
         // implicitly wrap if contextual type is a 32-bit integer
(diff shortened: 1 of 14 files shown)


vladimir-tikhonov:
Thanks for fixing that, @dcodeIO! By inspecting the `.wat` file from the original project I can see that two troublesome child constructors are allocating more memory, and offsets used for accessing their properties were updated as well. Memory allocation increased from 8 bytes up to 92 bytes (+84 bytes) while the parent class allocates only 81 byte if called directly, but I believe it's due to alignment?

dcodeIO:
Yeah, the size of the allocation is the exact number of bytes required to store an instance of the class in memory, before alignment if a hypothetical new field would be inserted. So, if the last field of the class is an `i8`, `u8` or `bool` that would be right.

Issue #1081, opened 2020-01-27
Exported mutable global breaks nodejs 10 as runtime environment


webmaster128:
[AssemblyScript 0.9.0 adds an exported mutable global (`__argumentsLength`)](https://github.com/AssemblyScript/assemblyscript/releases/tag/v0.9.0). When loaded in a NodeJS 10 environment, this triggers a [Wasm decoding failed: mutable globals cannot be exported](https://circleci.com/gh/confio/cosmwasm/1446).

<img width="1082" alt="Bildschirmfoto 2020-01-27 um 23 32 19" src="https://user-images.githubusercontent.com/2603011/73219836-4e524580-415d-11ea-892f-96c93d2d987c.png">

In Node JS 12, the error disappears. I think this is a serious limitation to be aware of. Personally, I don't mind too much since I only need nodejs for unittesting and can use version 12. Feel free to close if you consider this unproblematic.

dcodeIO:
Oh, yeah, that's a problem. I guess we need a fallback mechanism there, but since mutable globals has been standardized and is enabled by default now, that'd then be behind `--disable mutable-globals`. Can take a look.

mhonert:
This also seems to affect current versions of the Safari Browser. I could only check on iOS, but according to the compatibility table (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WebAssembly/Global), the desktop version does not support WebAssembly.Global either.
So, it would be great, if there was a flag to disable this feature.

PR #1084, opened 2020-01-28
Add a fallback for argumentsLength without mutable-globals

dcodeIO:
This adds a fallback for the mutable `__argumentsLength` global if `--disable mutable-globals` is given, in that it exports a function named `set:__argumentsLength` instead. Updates the loader accordingly.

fixes https://github.com/AssemblyScript/assemblyscript/issues/1081


dcodeIO:
Still wondering if we should just use the function (for now), even if it makes the module slightly larger.

dcodeIO:
This essentially reverts the mutable global change for now, always emitting a helper function until there's broader support for the feature. The loader is able to work with both to remain compatible to 0.9.0, and activating it again is just a matter of uncommenting a few lines in compiler.ts.

PR #1086, opened 2020-01-29
Categorize CLI options

dcodeIO:
The CLI options are becoming somewhat hard to grasp, so this PR reorders them and adds categories:

```
SYNTAX
  asc [entryFile ...] [options]

EXAMPLES
  asc hello.ts
  asc hello.ts -b hello.wasm -t hello.wat
  asc hello1.ts hello2.ts -b -O > hello.wasm

OPTIONS

 General

  --version, -v         Prints just the compiler's version and exits.
  --help, -h            Prints this message and exits.
  --noColors            Disables terminal colors.

 Optimizations

  --optimize, -O        Optimizes the module. Typical shorthands are:

                         Default optimizations   -O / -O3s
                         Make a release build    -O --noAssert
                         Make a debug build      --debug
                         Optimize for speed      -O3
                         Optimize for size       -O3z --converge

  --optimizeLevel       How much to focus on optimizing code. [0-3]
  --shrinkLevel         How much to focus on shrinking code size. [0-2, s=1, z=2]
  --converge            Re-optimizes until no further improvements can be made.
  --noAssert            Replaces assertions with just their value without trapping.
  --runtime             Specifies the runtime variant to include in the program.

                         full  Default runtime based on TLSF and reference counting.
                         half  Same as 'full', but not exported to the host.
                         stub  Minimal stub implementation without free/GC support.
                         none  Same as 'stub', but not exported to the host.


 Output

  --outFile, -o         Specifies the output file. File extension indicates format.
  --binaryFile, -b      Specifies the binary output file (.wasm).
  --textFile, -t        Specifies the text output file (.wat).
  --asmjsFile, -a       Specifies the asm.js output file (.js).
  --idlFile, -i         Specifies the WebIDL output file (.webidl).
  --tsdFile, -d         Specifies the TypeScript definition output file (.d.ts).

 Debugging

  --sourceMap           Enables source map generation. Optionally takes the URL
                        used to reference the source map from the binary file.
  --debug               Enables debug information in emitted binaries.

 Features

  --importMemory        Imports the memory provided as 'env.memory'.
  --sharedMemory        Declare memory as shared by settings the max shared memory.
  --importTable         Imports the function table provided as 'env.table'.
  --exportTable         Exports the function table as 'table'.
  --explicitStart       Exports an explicit '_start' function to call.
  --enable              Enables WebAssembly features being disabled by default.

                         sign-extension      Sign-extension operations
                         bulk-memory         Bulk memory operations.
                         simd                SIMD types and operations.
                         threads             Threading and atomic operations.
                         reference-types     Reference types and operations.

  --disable             Disables WebAssembly features being enabled by default.

                         mutable-globals     Mutable global imports and exports.

  --use, -u             Aliases a global object under another name, e.g., to switch
                        the default 'Math' implementation used: --use Math=JSMath
                        Can also be used to introduce an integer constant.
  --memoryBase          Sets the start offset of compiler-generated static memory.

 API

  --transform           Specifies the path to a custom transform to 'require'.

 Binaryen

  --validate, -c        Validates the module using Binaryen. Exits if invalid.
  --trapMode            Sets the trap mode to use.

                         allow  Allow trapping operations. This is the default.
                         clamp  Replace trapping operations with clamping semantics.
                         js     Replace trapping operations with JS semantics.

  --runPasses           Specifies additional Binaryen passes to run after other
                        optimizations, if any. See: Binaryen/src/passes/pass.cpp

 Other

  --baseDir             Specifies the base directory of input and output files.
  --noUnsafe            Disallows the use of unsafe features in user code.
                        Does not affect library files and external modules.
  --noEmit              Performs compilation as usual but does not emit code.
  --measure             Prints measuring information on I/O and compile times.
  --pedantic            Make yourself sad for no good reason.
  --lib                 Adds one or multiple paths to custom library components and
                        uses exports of all top-level files at this path as globals.
  --path                Adds one or multiple paths to package resolution, similar
                        to node_modules. Prefers an 'ascMain' entry in a package's
                        package.json and falls back to an inner 'assembly/' folder.
  --traceResolution     Enables tracing of package resolution.
  --listFiles           Lists files to be compiled and exits.
  --printrtti           Prints the module's runtime type information to stderr.
  -- ...                Specifies node.js options (CLI only). See: node --help
```

### DIFF
--- a/cli/asc.json
+++ b/cli/asc.json
@@ -1,92 +1,62 @@
 {
   "version": {
+    "category": "General",
     "description": "Prints just the compiler's version and exits.",
     "type": "b",
     "alias": "v"
   },
   "help": {
+    "category": "General",
     "description": "Prints this message and exits.",
     "type": "b",
     "alias": "h"
   },
+  "noColors": {
+    "category": "General",
+    "description": "Disables terminal colors.",
+    "type": "b",
+    "default": false
+  },
+
   "optimize": {
+    "category": "Optimizations",
     "description": [
-      "Optimizes the module. Also has the usual shorthands:",
+      "Optimizes the module. Typical shorthands are:",
       "",
-      " -O     Uses defaults. Equivalent to -O3s",
-      " -O0    Equivalent to --optimizeLevel 0",
-      " -O1    Equivalent to --optimizeLevel 1",
-      " -O2    Equivalent to --optimizeLevel 2",
-      " -O3    Equivalent to --optimizeLevel 3",
-      " -Oz    Equivalent to -O but with --shrinkLevel 2",
-      " -O3s   Equivalent to -O3 with --shrinkLevel 1 etc.",
+      " Default optimizations   -O / -O3s",
+      " Make a release build    -O --noAssert",
+      " Make a debug build      --debug",
+      " Optimize for speed      -O3",
+      " Optimize for size       -O3z --converge",
       ""
     ],
     "type": "b",
     "alias": "O"
   },
   "optimizeLevel": {
+    "category": "Optimizations",
     "description": "How much to focus on optimizing code. [0-3]",
     "type": "i"
   },
   "shrinkLevel": {
+    "category": "Optimizations",
     "description": "How much to focus on shrinking code size. [0-2, s=1, z=2]",
     "type": "i"
   },
   "converge": {
+    "category": "Optimizations",
     "description": "Re-optimizes until no further improvements can be made.",
     "type": "b",
     "default": false
   },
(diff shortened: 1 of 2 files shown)


dcodeIO:
Would also like to rename `--listFiles` to `--traceFiles`, which can then be combined with `--noEmit`, and `-printrtti` to `--traceRtti`. That'd be a breaking change though.

dcodeIO:
Also updated [the docs](https://docs.assemblyscript.org/details/compiler) accordingly.

PR #1120, opened 2020-02-20
Fix const enum compilation failure

DuncanUszkay1:
Issue: https://github.com/AssemblyScript/assemblyscript/issues/574

If you tried to use a const enum with multiple members, you would receive a compilation failure.

This was because the compiler relied upon each enum value to be registered as a global value in order to determine what the next enum value was. As a result, when we inlined the enum values and did not register that global value, compilation would fail.

To fix this, I've changed this method around a bit to rely upon the last expression value used (held in `initExpr`) rather than trying to load from the global store.

I'm not sure why it was implemented the way it was originally- it's possible that there is some tradeoff involved here I'm unaware of which causes reading from the global store to be faster/more efficient than simply reusing the last expression statement used in the loop. This certainly seems to work from a functional perspective 🤷‍♂

### DIFF
--- a/src/compiler.ts
+++ b/src/compiler.ts
@@ -1084,10 +1084,16 @@ export class Compiler extends DiagnosticEmitter {
               (<EnumValue>member).identifierNode.range.atEnd
             );
           }
-          initExpr = module.binary(BinaryOp.AddI32,
-            module.global_get(previousValue.internalName, NativeType.I32),
-            module.i32(1)
-          );
+          if (isInline) {
+            let value = i64_add(previousValue.constantIntegerValue, i64_new(1));
+            assert(!i64_high(value));
+            initExpr = module.i32(i64_low(value));
+          } else {
+            initExpr = module.binary(BinaryOp.AddI32,
+              module.global_get(previousValue.internalName, NativeType.I32),
+              module.i32(1)
+            );
+          }
           initExpr = module.precomputeExpression(initExpr);
           if (getExpressionId(initExpr) != ExpressionId.Const) {
             if (element.is(CommonFlags.CONST)) {
--- a/NOTICE
+++ b/NOTICE
@@ -19,6 +19,7 @@ under the licensing terms detailed in LICENSE:
 * jhwgh1968 <jhwgh1968@protonmail.com>
 * Jeffrey Charles <jeffreycharles@gmail.com>
 * Vladimir Tikhonov <reg@tikhonov.by>
+* Duncan Uszkay <duncan.uszkay@shopify.com>
 
 Portions of this software are derived from third-party works licensed under
 the following terms:
--- a/tests/compiler/enum.ts
+++ b/tests/compiler/enum.ts
@@ -12,6 +12,13 @@ export const enum ImplicitConst {
   THREE
 }
 
+const enum ImplicitConstNoExport {
+  ZERO,
+  ONE,
+  TWO,
+  THREE
+}
+
 export enum Explicit {
   ZERO = 0,
   ONE = 0 + 1,


dcodeIO:
I believe the underlying idea was that if one has an enum like

```ts
enum A {
  A
}
```

that the value of `A` would be `0` since it's the first value (no initializer, no previousValue) and if one has an enum like

```ts
enum A {
  B = 2;
  C;
}
```

that the value of C would become `B + 1`.

Regarding the PR specifically, it is not possible to reuse the same expression as a member of two different expressions, so if `B` is initialized with `initExpr`, that same `initExpr` cannot be used again to initialize `C`, but one must `global.get` the value of `A.B` (use a new expression).

DuncanUszkay1:
You say it isn't possible, but the code seems to work 🤔 here's the example that wasn't previously working:
```typescript
// foo.ts
export const enum Foo {
  Boo1, Boo2, Boo3, Boo4
}

// main.ts
import { Foo } from './foo';

export function test(): i32 {
  return Foo.Boo1 + Foo.Boo2 + Foo.Boo3 + Foo.Boo4);
}
```

With this change it compiles and outputs: `6`. Seems like tests pass as well

What's the scenario where this fails?

dcodeIO:
Not reusing the same expression multiple times is a Binaryen convention. From their README:

> Notes when working with Binaryen IR:
> 
> As mentioned above, Binaryen IR has a tree structure. As a result, each expression should have exactly one parent - you should not "reuse" a node by having it appear more than once in the tree. The motivation for this limitation is that when we optimize we modify nodes, so if they appear more than once in the tree, a change in one place can appear in another incorrectly.

DuncanUszkay1:
Ah I see. Back to the drawing board, I'll look into what you suggested above. Thanks for the review

src/compiler.ts
```diff
@@ -1084,8 +1084,11 @@ export class Compiler extends DiagnosticEmitter {
               (<EnumValue>member).identifierNode.range.atEnd
             );
           }
+          var previousValueExpression = isInline ?
+            module.i32(changetype<i32>(previousValue.constantIntegerValue)) :
```
DuncanUszkay1: This works under the assumption that inline enums are immutable- this is an assertion that I think makes sense but I lack context

dcodeIO: That's the right location and idea, yeah. Note that `previousValue.constantIntegerValue` is a portable I64 here, intended to be accessed via the `i64_*` helpers (it's a Long.js instance in JS, an i64 in Wasm):

```ts
var previousValueExpr: ExpressionRef;
if (isInline) {
  let value = previousValue.constantIntegerValue;
  assert(!i64_high(value));
  previousValueExpr = module.i32(i64_low(value));
} else {
  previousValueExpr = module.global_get(previousValue.internalName, NativeType.I32);
}
```

Should work fine then :)

DuncanUszkay1:
This should be ready for review again

src/compiler.ts
```diff
@@ -1084,8 +1084,16 @@ export class Compiler extends DiagnosticEmitter {
               (<EnumValue>member).identifierNode.range.atEnd
             );
           }
+          var previousValueExpr: ExpressionRef;
+          if (isInline) {
+            let value = previousValue.constantIntegerValue;
+            assert(!i64_high(value));
+            previousValueExpr = module.i32(i64_low(value));
```
dcodeIO: Now that I look at it again, we can even precompute the add here in the inline case using `i64_add(..., i64_new(1))` and set `initExpr` directly.

dcodeIO:
Thank you! :)

Issue #1117, opened 2020-02-20
Importing and Exporting a class of the same name returns an undescriptive Error


torch2424:
cc @dcodeIO @MaxGraey 

So I've been running into a bunch of random edge cases like this in my AssemblyScript error. So I figured I should start recording / reporting them.

Here is my [Test Case](https://github.com/torch2424/as-playground/tree/master/name-conflict-assertion-error).

But if you try to build a module that imports a class, but also has a definition of a class with the same name, you get the following error:

![Screenshot from 2020-02-20 02-27-21](https://user-images.githubusercontent.com/1448289/74925141-8b8aab80-5388-11ea-853d-7c219fa6aa59.png)

Which is not a good developer experience. 

Again, I don't think this is super high priority, but I just think we should be aware :smile:

torch2424:
Oh I just realized you closed this. Thanks @dcodeIO ! :smile: :tada:

PR #1124, opened 2020-02-26
Fix assertion on duplicate imported identifiers

dcodeIO:
This is an attempt to fix https://github.com/AssemblyScript/assemblyscript/issues/1117 by removing an assertion and providing a better (local) identifier reference for the emitted diagnostic.


torch2424:
Thanks for fixing my small issue @dcodeIO :smile: :+1:

PR #1134, opened 2020-02-29
Add asbuild utility, update asinit to use it

jhwgh1968:
At a high level, this separates out invocation of `asc` from `npm` (and several quirks related to targets passing arguments) by creating a new command: `asbuild`.

It supports explicitly passing flags to `asc`, but mainly focuses on *build profiles*, each of which corresponds to a certain set of flags. These profiles can easily be added or updated by updating a JSON file.

Usage examples:
```
asbuild  # runs a debug build
asbuild --release  # runs a build with maximum optimization
asbuild --relmin  # runs a build that is optimized for size
asbuild --debug --release  # runs both a debug and a release build
```

This new tool uses existing command-line parsing, does not add any dependencies, and allows explicitly specifying output file names the same way that `asc` does.

This completes the work started in #975, and incorporates all feedback from that discussion.

### DIFF
--- a/bin/asbuild
+++ b/bin/asbuild
@@ -1,0 +1,10 @@
+#!/usr/bin/env node
+
+/* tslint:disable */
+
+try { require("source-map-support").install(); } catch (e) {}
+
+const asbuild = module.exports = require("../cli/asbuild/asbuild.js");
+if (/\basbuild$/.test(process.argv[1])) {
+  process.exitCode = asbuild.main(process.argv.slice(2));
+}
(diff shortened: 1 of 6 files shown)


dcodeIO:
Has taken me a while to get back to this issue, sorry. While I believe we need something better here, and this is going into the right direction, I am uncertain whether we have pinpointed the optimal solution yet. The new `asconfig` efforts highlight some of the potential conflicts in that their `extends` mechanism would cover AS-provided presets, and their `targets` mechanism would allow multiple targets with slightly varying config options. Like, in a world with `asconfig`, I'd imagine `asbuild` to become a more convenient shorthand for `asc --config asbuild.json --target something`, i.e. `asbuild something`, so I have been wondering if both efforts can be combined?

jhwgh1968:
Ah, there is another effort on this front since I opened this!

I will post a comment over on #1230.

PR #1144, opened 2020-03-03
Modernize source file headers

dcodeIO:
Source files were still using some hacky comments to be compatible with tsdoc, which turned out we don't use/need. So this PR updates the comments to a more friendly format while also including license headers.

* The full header is present in `src/index.ts`
* Other source files use a `@license` tag with the SPDX identifier
* Distribution files have a single closure-library style short header


dcodeIO:
About adding my name to the headers: That's something that lately came up in a correspondence between a corporate user of long.js and me, since "The XY Contributors" was too vague for their legal department as that's not a legal entity I suppose. Now I know that a lot of open source projects do it that other way, but I guess it's always better to be safe than sorry.

PR #1145, opened 2020-03-03
Add postInitialize transform step

jtenner:
This pull request aims to add a transform step to expose the `program` object before the compiler has completed the compile step.

Completed:

- [x] Add Transform Definition to .d.ts file
- [x] Call initialize conditionally and only once
- [x] Add Transform to asc.js

Please advise :)


dcodeIO:
Perfect, thanks! :)

PR #1161, opened 2020-03-12
StaticArray type fix

jtenner:
`StaticArray`s are `@sealed`, so declaring them in the `index.d.ts` file as `abstract` causes vscode to report errors when using the `StaticArray` `constructor`.

This pull request removes the `abstract` keyword to prevent users from experiencing confusion about how `StaticArray` works.

### DIFF
--- a/std/assembly/index.d.ts
+++ b/std/assembly/index.d.ts
@@ -1400,8 +1400,8 @@ declare class Array<T> {
   toString(): string;
 }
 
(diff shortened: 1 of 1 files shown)


dcodeIO:
Thanks!

dcodeIO:
For the time being:

```ts
import { StaticArray } from "staticarray";

new StaticArray<i32>(10);
```

PR #1164, opened 2020-03-13
tsconfig can use node resolution instead of relative

willemneal:
This way when asinit is used globally it won't point to the global definition and is less fragile.

### DIFF
--- a/bin/asinit
+++ b/bin/asinit
@@ -159,10 +159,9 @@ function ensureAssemblyDirectory() {
 
 function ensureTsconfigJson() {
   console.log("- Making sure that 'assembly/tsconfig.json' is set up...");
-  const base = tsconfigBase.replace(/\\/g, "/");
   if (!fs.existsSync(tsconfigFile)) {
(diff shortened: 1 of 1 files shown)


dcodeIO:
Cool, didn't know this is possible. Does this work if the compiler has been installed globally exclusively?

willemneal:
Just tried it and it doesn't; personally I've found it useful to have the assemblyscript local for the version and CI, etc

dcodeIO:
Perhaps the TS guys are interested in a PR to make this possible. Would be nice to make it just work in all circumstances :)

dcodeIO:
In the meantime, would it be possible to make this conditional? Let's say `asinit` finds that assemblyscript is local, it would use node resolution, and otherwise use the exact path as before?

dcodeIO:
Any thoughts on this? :)

dcodeIO:
Closing this PR in favor of https://github.com/AssemblyScript/assemblyscript/pull/1322.

Issue #1176, opened 2020-03-16
Empty use abort declaration triggers asc assertion since 0.9.3


jeffcharles:
When I call `asc` with `--use abort=` as documented in https://docs.assemblyscript.org/details/debugging#overriding-abort, I see:

```
➜  mar16 npx asc src/mar16.ts -b build/mar16.wasm --sourceMap --validate --optimize --use abort= --runtime none --lib=node_modules 
ERROR: AssertionError: assertion failed
    at assert (/Users/jeffreycharles/scripts/mar16/node_modules/assemblyscript/dist/webpack:/assemblyscript/std/portable/index.js:184:9)
    at g.initialize (/Users/jeffreycharles/scripts/mar16/node_modules/assemblyscript/dist/webpack:/assemblyscript/src/program.ts:1007:22)
    at Object.t.initializeProgram (/Users/jeffreycharles/scripts/mar16/node_modules/assemblyscript/dist/webpack:/assemblyscript/src/index.ts:221:11)
    at /Users/jeffreycharles/scripts/mar16/node_modules/assemblyscript/cli/asc.js:595:22
    at measure (/Users/jeffreycharles/scripts/mar16/node_modules/assemblyscript/cli/asc.js:1103:3)
    at Object.main (/Users/jeffreycharles/scripts/mar16/node_modules/assemblyscript/cli/asc.js:594:26)
    at /Users/jeffreycharles/scripts/mar16/node_modules/assemblyscript/bin/asc:21:47
```

It looks like src/program.ts on line 1007 has `let name = assert(globalAliases.get(alias));` and `globalAliases` has a key called `abort` which has a value of `''` which is falsey in JavaScript which causes the `assert` to fire.

DuncanUszkay1:
Going to look into this

PR #1177, opened 2020-03-16
Change assert to allow blank strings

DuncanUszkay1:
Issue: https://github.com/AssemblyScript/assemblyscript/issues/1176

When using the documented `--use abort=`abort would be set to an empty string, which resulted in the call to global aliases returning ''. Then since blank strings are falsey, this assertion would fail.

I've changed the assertion to assert that the name is a string, which catches the case that the alias is missing but does not fail in the case of an empty string

examples:
If the name isn't found in the aliases table:
```typescript
//let alias = unchecked(_keys[i]);
let alias = "not the right key"
let name = globalAliases.get(alias);
assert(isString(name));
```
We get: `ERROR: AssertionError: assertion failed` as expected.

In the case where you specify an abort that does not match any export:
`ERROR: no such global element: main/nope` as before

In the case where you use a blank abort, compilation is successful.

### DIFF
--- a/src/program.ts
+++ b/src/program.ts
@@ -1044,7 +1044,7 @@ export class Program extends DiagnosticEmitter {
       // TODO: for (let [alias, name] of globalAliases) {
       for (let _keys = Map_keys(globalAliases), i = 0, k = _keys.length; i < k; ++i) {
         let alias = unchecked(_keys[i]);
-        let name = changetype<string>(globalAliases.get(alias))
+        let name = changetype<string>(globalAliases.get(alias));
         assert(name != null);
         if (!name.length) continue; // explicitly disabled
         let firstChar = name.charCodeAt(0);
--- a/src/program.ts
+++ b/src/program.ts
@@ -1044,7 +1044,9 @@ export class Program extends DiagnosticEmitter {
       // TODO: for (let [alias, name] of globalAliases) {
       for (let _keys = Map_keys(globalAliases), i = 0, k = _keys.length; i < k; ++i) {
         let alias = unchecked(_keys[i]);
-        let name = assert(globalAliases.get(alias));
+        let name = globalAliases.get(alias);
+        assert(isString(name));
+        name = name!
         if (!name.length) continue; // explicitly disabled
         let firstChar = name.charCodeAt(0);
         if (firstChar >= CharCode._0 && firstChar <= CharCode._9) {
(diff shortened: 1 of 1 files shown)


src/program.ts
```diff
@@ -1044,7 +1044,9 @@ export class Program extends DiagnosticEmitter {
       // TODO: for (let [alias, name] of globalAliases) {
       for (let _keys = Map_keys(globalAliases), i = 0, k = _keys.length; i < k; ++i) {
         let alias = unchecked(_keys[i]);
-        let name = assert(globalAliases.get(alias));
+        let name = globalAliases.get(alias);
```
dcodeIO: Can just use `changetype<Map<string,string>>(globalAliases.get(alias))` since the value can't be `null`

DuncanUszkay1: but it can be undefined right?
```typescript
alias = "bad key"
let name = changetype<string>(globalAliases.get(alias))
if (!name.length) continue; // explicitly disabled
```
throws
```
ERROR: TypeError: Cannot read property 'length' of undefined
```
I suppose that's not really much different than the assertion though and shouldn't really happen anyway

dcodeIO: Hmm, yeah, that's strange. Might be `cli/asc.js` wrongly inserting an `undefined` into the map.

dcodeIO: Btw, `alias = "bad key"` doesn't happen since we `_keys = Map_keys(globalAliases)`

DuncanUszkay1: Ah okay- well in that case maybe we should just use `!` and ignore the undefined case

src/program.ts
```diff
@@ -1044,7 +1044,7 @@ export class Program extends DiagnosticEmitter {
       // TODO: for (let [alias, name] of globalAliases) {
       for (let _keys = Map_keys(globalAliases), i = 0, k = _keys.length; i < k; ++i) {
         let alias = unchecked(_keys[i]);
-        let name = assert(globalAliases.get(alias));
+        let name = changetype<string>(globalAliases.get(alias))!
```
dcodeIO: Hmm, this will now result in an `is never null` warning when compiling to Wasm. `npm run asbuild` should show this. Perhaps

```
let name = changetype<string>(globalAliases.get(alias));
assert(name != null);
```

dcodeIO: Appears that after https://github.com/AssemblyScript/assemblyscript/pull/1177/commits/e8ef003255025732fc4117b13184354b70372029 the `!` leading to the warning is still there. Looking good to me otherwise :)

DuncanUszkay1: Ah of course the `!`, missed that

DuncanUszkay1:
Not sure I understand why CI is failing here- doesn't appear to be related to the PR

dcodeIO:
Seems to me that there's something wrong with GitHub's git, hmm. Will try to rerun the CI job at a later time.

DuncanUszkay1:
Is this blocked on CI failures?

DuncanUszkay1:
CI is passing now, should be good to merge?

PR #1179, opened 2020-03-20
Remove option checking for v128 registration

### LINKED ISSUE #1162: Properties that return v128 cause compiler error even though they are conditionally used
```ts
class ExampleClass {
  a: f64;
  b: f64;

  get ab(): v128 {
    return v128.load(changetype<usize>(this), offsetof<ExampleClass>("a"));
  }
}

if (!ASC_FEATURE_SIMD) {
  let a = new ExampleClass();
  a.a = 1;
}
```

Minimal reproduction causes a compiler error that reports like this:

```
  ERROR TS2304: Cannot find name 'v128'.

     get ab(): v128 {
               ~~~~
   in simd-error.ts(6,12)
```

jtenner:
Fixes #1162 hopefully

This particular change doesn't break the test suite, and I believe something similar should also occur for the `anyref` feature as well.

What is the next step?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -429,8 +429,7 @@
     "arrify": {
       "version": "1.0.1",
       "resolved": "https://registry.npmjs.org/arrify/-/arrify-1.0.1.tgz",
-      "integrity": "sha1-iYUI2iIm84DfkEcoRWhJwVAaSw0=",
-      "dev": true
+      "integrity": "sha1-iYUI2iIm84DfkEcoRWhJwVAaSw0="
     },
     "asn1.js": {
       "version": "4.10.1",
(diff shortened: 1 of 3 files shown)


jtenner:
wow something broke. That's so weird.

jtenner:
It looks like the builtins all perform the following checks for the simd functions:

```ts
function builtin_v128_extract_lane(ctx: BuiltinContext): ExpressionRef {
  var compiler = ctx.compiler;
  var module = compiler.module;
  if (
    checkFeatureEnabled(ctx, Feature.SIMD) |
    checkTypeRequired(ctx, true) |
    checkArgsRequired(ctx, 2)
  ) return module.unreachable();
```

So all the builtins should be okay.

I'm going to start taking guesses here though...

jtenner:
Tests run fine locally. Are these checks appropriate?

dcodeIO:
After thinking a bit more about this, it appears best to just allow these types and leave validation to Binaryen, but improve the error output we get from it. Otherwise we'll end up with lots of checks all over the place, which is not only unnecessary clutter but also inefficient to do for every single declaration etc.

Perhaps we can make it so that Binaryen also returns the expression leading to a validation error, so we can look it up from debug info and print a better summary of validation errors.

PR #1185, opened 2020-03-25
Check precondition in CI on PRs only

dcodeIO:
Turned out that the author check in our CI action can use different author email addresses for the PR and the merge, so this PR excludes the author check from merges as well.


dcodeIO:
Alright, this didn't work. Now doesn't run anything on merge at all :)

dcodeIO:
Fixed by a force-push.

PR #1187, opened 2020-03-25
StaticArray loader support

jtenner:
This pull request aims to provide loader support for StaticArray.


jtenner:
I think my only question so far is how do I determine if the provided id is a StaticArray type?

```ts
const STATICARRAY = 1 << 23; // this can't be correct
```

jtenner:
I will get around to duplicating a lot of the tests in the loader module for StaticArray. Is the current approach okay?

lib/loader/index.js
```diff
@@ -27,6 +27,9 @@ const KEY_FLOAT = 1 << 20;
 const KEY_NULLABLE = 1 << 21;
 const KEY_MANAGED = 1 << 22;
 
+// TODO: is this the right location for STATICARRAY flag?
+const STATICARRAY = 1 << 23;
```
dcodeIO: There is [no such flag yet](https://github.com/AssemblyScript/assemblyscript/blob/master/std/assembly/shared/typeinfo.ts#L25). The compiler will have to add it when building RTTI by checking whether a class extends StaticArray.

jtenner: Okay! May I add it to the compiler in this pull request? I just need a bit of guidance for this part.

dcodeIO: Yeah, without this PR won't work. Relevant code location, besides adding the flag to typeinfo, is: https://github.com/AssemblyScript/assemblyscript/blob/master/src/builtins.ts#L7888. Should be possible to reuse most of what's done for arrays already. Suggesting to move all other flags one up, and adding the static array flag below ARRAY. Must then also update the flag values in the loader accordingly.

jtenner: Okay. Got it.

jtenner:
@dcodeIO it looks like we need to update the fixtures, and it looks like we now hit an assertion error.

```
  AssertionError: assertion failed
      at assert (C:\Users\Joshua Tenner\projects\assemblyscript\std\portable\index.js:184:9)
      at Class.getArrayValueType (C:\Users\Joshua Tenner\projects\assemblyscript\src\program.ts:3763:17)
      at Object.compileRTTI (C:\Users\Joshua Tenner\projects\assemblyscript\src\builtins.ts:7932:32)
      at Compiler.compile (C:\Users\Joshua Tenner\projects\assemblyscript\src\compiler.ts:500:54)
      at Object.compile (C:\Users\Joshua Tenner\projects\assemblyscript\src\index.ts:232:32)
      at C:\Users\Joshua Tenner\projects\assemblyscript\cli\asc.js:606:31
      at measure (C:\Users\Joshua Tenner\projects\assemblyscript\cli\asc.js:1129:3)
      at Object.main (C:\Users\Joshua Tenner\projects\assemblyscript\cli\asc.js:605:26)
      at C:\Users\Joshua Tenner\projects\assemblyscript\tests\compiler.js:237:9
      at Object.main (C:\Users\Joshua Tenner\projects\assemblyscript\cli\asc.js:609:12)
```

jtenner:
Okay! Compiler tests pass. Now to add some tests in the loader for StaticArrays.

jtenner:
Looks like we should also support __getArray too. Is it really a good idea to add `__get*StaticArray` methods? It increases the surface area, and maybe it's not necessary. We could probably just fix up the `__getTypedArray()` method and remove the new methods.

lib/loader/index.d.ts
```diff
@@ -84,6 +84,47 @@ export interface ASUtil {
...
+    /** Copies an StaticArray<u32>'s values from the module's memory. */
+    __getUint32StaticArray(ptr: number): Uint32Array;
+    /** Copies an StaticArray<i64>'s values from the module's memory. */
+    __getInt64StaticArray?(ptr: number): BigInt64Array;
+    /** Copies an StaticArray<u64>'s values from the module's memory. */
+    __getUint64StaticArray?(ptr: number): BigUint64Array;
+    /** Copies a StaticArray<f32>'s values from the module's memory. */
+    __getFloat32StaticArray(ptr: number): Float32Array;
+    /** Copies a StaticArray<f64>'s values from the module's memory. */
+    __getFloat64StaticArray(ptr: number): Float64Array;
```
dcodeIO: Makes me wonder if we should instead expose an enum of types and have a single function to do this, like `__getStaticArray(theType, ptr)`?

jtenner: My suggestion was to modify getTypedArray() to accept static array pointers. Would this suffice?

jtenner: Also we need to handle static arrays of references...

lib/loader/index.js
```diff
@@ -259,76 +259,40 @@ function postInstantiate(extendedExports, instance) {
   function getTypedArrayView(Type, alignLog2, ptr) {
     const buffer = memory.buffer;
     const U32 = new Uint32Array(buffer);
-    const bufPtr = U32[ptr + ARRAYBUFFERVIEW_DATASTART_OFFSET >>> 2];
+    const id = U32[ptr + ID_OFFSET >>> 2];
+    const info = getInfo(id);
```
dcodeIO: @MaxGraey Are you ok with this? Iirc this helper originated from trying to avoid every inch of overhead, including not having to read RTTI with `getInfo`.

jtenner: Oh. Yeah. Should we just revert back to what I had before? I really don't mind having a larger surface area of functions like I had before.

jtenner:
@MaxGraey and @dcodeIO what is the strategy we should use here?

- I really liked the initial approach I had with the `get(Type)StaticArray()`. They had a lot of reduced overhead
- The second approach that I have currently modifies the `get(Type)Array()` methods to obtain data views which ends up requiring inspecting RTTI
- A third approach might be to create a single function with an enum

```ts
// I can work the types to function correctly return the correct type
function getStaticArrayView(type: StaticArrayType, pointer: number): TypedArray {
  const id = U32[arr + ID_OFFSET >>> 2];
  // assert( (type & getInfo(id)) === type )
  // perform some kind of switch statement here
  // return the typed array view
}
```

I prefer the first method which only added two new functions and increased our surface area API by a lot.

dcodeIO:
What about something like

```ts
[Int8Array, Uint8Array, ...].forEach(ctor => {
  // normal
  extendedExports["__get" + ctor.name] = getTypedArray.bind(null, ctor, 31 - Math.clz(ctor.BYTES_PER_ELEMENT));
  // static
  ...
});
```

Might get us all the functions with less redundant code. Wdyt?

jtenner:
Wow that looks great.

lib/loader/index.js
```diff
@@ -267,32 +267,34 @@ function postInstantiate(extendedExports, instance) {
...
 
-  extendedExports.__getFloat32Array = getTypedArray.bind(null, Float32Array, 2);
-  extendedExports.__getFloat32ArrayView = getTypedArrayView.bind(null, Float32Array, 2);
-  extendedExports.__getFloat64Array = getTypedArray.bind(null, Float64Array, 3);
-  extendedExports.__getFloat64ArrayView = getTypedArrayView.bind(null, Float64Array, 3);
+  ctors.forEach(ctor => {
+    let name = "__get" + ctor.name.replace("Array", "");
+    let align = 31 - Math.clz(ctor.BYTES_PER_ELEMENT);
+    extendedExports[name + "ArrayView"] = getTypedArrayView.bind(null, ctor, align);
+    extendedExports[name + "Array"] = getTypedArrayView.bind(null, ctor, align);
```
dcodeIO: Is it correct that these are the same?

jtenner: This was fixed :)

jtenner:
Just making small optimizations now. @MaxGraey @dcodeIO how should we test the loader for these functions?

jtenner:
Something strange happened on ci. All my tests work fine locally.

dcodeIO:
Error is

```
  trace: origLowerCode != expectLowerCode 42951, 18446744069414562000, 18446744069414562000
  abort: null in std/string-casemapping.ts(148:3)
- instantiate untouched ERROR: RuntimeError: unreachable
```

cc @MaxGraey

jtenner:
I didn't change anything. That error shouldn't be happening. Could be a fluke.

jtenner:
trace: 18446744069414561735 <- actual
  trace: 18446744069414561736 <- expected

This change happened because node updated :).

jtenner:
This pull request is now done to the best of my ability and knowledge. It duplicated all the tests, and then changed the function calls to use the StaticArray functions where applicable.

Any feedback would be appreciated, and I would love to get this into a nightly release very soon. :)

lib/loader/index.js
```diff
@@ -233,30 +263,34 @@ function postInstantiate(extendedExports, instance) {
...
-  extendedExports.__getFloat64Array = getTypedArray.bind(null, Float64Array, 3);
-  extendedExports.__getFloat64ArrayView = getTypedArrayView.bind(null, Float64Array, 3);
+
+  ctors.forEach(ctor => {
+    let name = "__get" + ctor.name.slice(0, -5);
+    let align = 31 - Math.clz32(ctor.BYTES_PER_ELEMENT);
+    extendedExports[name + "ArrayView"] = getTypedArrayView.bind(null, ctor, align);
+    extendedExports[name + "Array"] = getTypedArray.bind(null, ctor, align);
+    extendedExports[name + "StaticArrayView"] = getStaticArrayView.bind(null, ctor, align);
+    extendedExports[name + "StaticArray"] = getStaticArray.bind(null, ctor, align);
```
dcodeIO: Wdyt of naming the static variants `__getStaticInt8ArrayView` etc? Feels a bit more ergonomic to me and also makes this setup code a little simpler (no need to slice name).

jtenner: It's a `StaticArray`, and putting those two words together seems important to me to indicate the purpose of the function. But this is *your* project.

dcodeIO: Also wondering: Do the `BigInt64Array` ctors now give us `__getBigInt64Array` etc. functions? I think these used to be `__getInt64Array` etc.

jtenner: I'm happy to change it to whatever feels better.

dcodeIO: Got me thinking: Should we actually handle the static arrays like typed arrays here in the first place? Since there are neither `Int8StaticArray`s nor `StaticInt8Array`s as of my initial suggestion, this might be confusing anyway. Wdyt of designing the API analogous to generic `Array`s instead, `__getStaticArray`?

jtenner: My thoughts:

- Getting specific methods for returning the typed arrays makes things very fast, (even faster than ArrayBufferView functions)
- Naming them `__getStatic*(View?)` is perfectly fine and I can fix that.

I propose renaming the functions as you describe and leaving the api AS-IS.

jtenner: The reason I feel this way is because we can already use the `__getArrayView()` function to obtain the same result without having to worry about the type of array.

dcodeIO: Still feels like feature creep to me. Perhaps there is a way to refactor the API in general, to less functions, that are still fast?

Issue #1183, opened 2020-03-24
Unmanaged class Initialized with object literal returns unexpected value


jtenner:
Minimal repro: https://webassembly.studio/?f=3b3mzeyveii

Referenced: https://github.com/jtenner/as-pect/issues/278

PR #1189, opened 2020-03-26
Handle instantiation of unmanaged class via object literal

dcodeIO:
As reported in https://github.com/AssemblyScript/assemblyscript/issues/1183 instantiating an unmanaged class from an object literal, like in

```ts
@unmanaged
class Foo {
  a: i32;
}

var foo: Foo = { a: 1 };
```

results in unexpected behavior. Wasn't able to reproduce the issue Wasm Studio produces on current master, but it was indeed trying to retain/release unmanaged memory, which might have been the cause and should be fixed now, along a proper diagnostic when attempting this in `--noUnsafe` contexts. @jtenner Does this fix the original problem?

### DIFF
--- a/src/compiler.ts
+++ b/src/compiler.ts
@@ -6628,13 +6628,20 @@ export class Compiler extends DiagnosticEmitter {
   }
 
   /** Checks that an unsafe expression is allowed. */
-  private checkUnsafe(reportNode: Node): void {
+  private checkUnsafe(reportNode: Node, relatedReportNode: Node | null = null): void {
     // Library files may always use unsafe features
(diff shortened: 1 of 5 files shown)


jtenner:
It looks like it! @DuncanUszkay1 should review this as well.

jtenner:
Also, this is going to make instantiating iovecs a bit more ergonomic.

dcodeIO:
Would be great if someone with that issue could validate that the PR indeed fixes the problem :)

jtenner:
I yeah tried out Duncan's example and it does work.

jtenner:
Any update?

PR #1194, opened 2020-03-27
Always register non-MVP types and check on actual use

dcodeIO:
This PR should fix https://github.com/AssemblyScript/assemblyscript/issues/1162 in that `v128` and `anyref` will always be present, even if the respective feature is not enabled, so these can be used conditionally. Chances are that I am missing diagnostics somewhere, but general strategy is to check on use of a storage type

* Declaring a local of that type
* Declaring a global of that type
* Reading or writing a field of that type
* Compiling a function with a respective this, parameter or return type
* Instantiating a class with a field of that type

and when a new value of the type is created by one of the respective builtins, which should also cover all sorts of inference if no type is annotated on a storage type. That should give us the minimum amount of diagnostics, instead of erroring on every single compiled expression resolving to that type, along minimal checking overhead.

Unless discovered in review, we'll notice that a diagnostic is missing if Binaryen fails to validate a module due to unexpected types.

### DIFF
--- a/src/compiler.ts
+++ b/src/compiler.ts
@@ -60,7 +60,8 @@ import {
   CommonNames,
   INDEX_SUFFIX,
   Feature,
-  Target
+  Target,
+  featureToString
 } from "./common";
 
 import {
@@ -113,6 +114,7 @@ import {
   DecoratorKind,
   AssertionKind,
   SourceKind,
+  FunctionTypeNode,
 
   Statement,
   BlockStatement,
@@ -961,6 +963,7 @@ export class Compiler extends DiagnosticEmitter {
           return false;
         }
         global.setType(resolvedType);
+        this.checkTypeSupported(global.type, typeNode);
 
       // Otherwise infer type from initializer
       } else if (initializerNode) {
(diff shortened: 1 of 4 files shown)


jtenner:
Should I validate that this works locally?

dcodeIO:
Would be great :)

jtenner:
After waiting about ten minutes for the dist files to generate, it appears this works locally for me.

```ts
class Test {
  a: f32 = 0.0;
  b: f32 = 0.0;
  c: f32 = 0.0;
  d: f32 = 0.0;
  get abcd(): v128 {
    return v128.load(changetype<usize>(this), offsetof<Test>("a"));
  }
  set abcd(value: v128) {
    v128.store(changetype<usize>(this), value, offsetof<Test>("a"));
  }
}

let a = new Test();
if (ASC_FEATURE_SIMD) {
  let vec = v128.mul<f32>(a.abcd, v128.splat<f32>(2));
  a.abcd = vec;
} else {
  a.a *= 2;
  a.b *= 2;
  a.c *= 2;
  a.d *= 2;
}
```

It compiles when simd is enabled and when simd is not enabled.

Thanks for the fix.

PR #1198, opened 2020-03-28
Asinit yes

### LINKED ISSUE #1063: Add --yes argument to asinit
It would be cool if `./node_modules/.bin/asinit . --yes` would answer all questions with _yes_ to allow for quick or automatic project setups of throw away projects.

Interface could work like `npm init --yes`:

```
$ npm help init
NPM-INIT(1)                                                        NPM-INIT(1)



NAME
       npm-init - create a package.json file

   Synopsis
         npm init [--force|-f|--yes|-y|--scope]
         npm init <@scope> (same as `npx <@scope>/create`)
         npm init [<@scope>/]<name> (same as `npx [<@scope>/]create-<name>`)

   Examples
       Create    a    new    React-based    project   using   create-react-app
       https://npm.im/create-react-app:

         $ npm init react-app ./my-react-app

       Create    a    new    esm-compatible    package    using     create-esm
       https://npm.im/create-esm:

         $ mkdir my-esm-lib && cd my-esm-lib
         $ npm init esm --yes

       Generate a plain old package.json using legacy init:

         $ mkdir my-npm-pkg && cd my-npm-pkg
         $ git init
         $ npm init

       Generate it without having it ask any questions:

         $ npm init -y

   Description
       npm  init  <initializer>  can  be  used to set up a new or existing npm
       package.

       initializer in this case is an npm package named  create-<initializer>,
       which  will  be  installed by npx https://npm.im/npx, and then have its
       main bin executed -- presumably creating or updating  package.json  and
       running any other initialization-related operations.

       The  init  command  is  transformed to a corresponding npx operation as
       follows:

       o npm init foo -> npx create-foo

       o npm init @usr/foo -> npx @usr/create-foo

       o npm init @usr -> npx @usr/create


       Any additional options will be passed directly to the command,  so  npm
       init foo --hello will map to npx create-foo --hello.

       If  the  initializer  is  omitted (by just calling npm init), init will
       fall back to legacy init behavior. It will ask you  a  bunch  of  ques-
       tions,  and  then write a package.json for you. It will attempt to make
       reasonable guesses based on existing fields, dependencies, and  options
       selected.  It is strictly additive, so it will keep any fields and val-
       ues that were already set. You can also use -y/--yes to skip the  ques-
       tionnaire  altogether.  If  you  pass  --scope, it will create a scoped
       package.

   See Also
       o https://github.com/isaacs/init-package-json

       o npm help package.json

       o npm help version

       o npm help scope




                                 December 2019                     NPM-INIT(1)
```

jtenner:
Fixes #1063 

Open to suggestions. This seems to do the trick though.


jtenner:
I used link locally and validated this works now.

bin/asinit
```diff
@@ -6,8 +6,26 @@ const fs = require("fs");
...
+    "category": "General",
+    "description": "Skips asking the user if they want to setup an AssemblyScript project.",
+    "type": "b",
+    "alias": "y"
+  }
+};
+
+const cliOptions = options.parse(process.argv, asinitOptions);
+
+if (cliOptions.help) printHelp();
```
jtenner: Oh. Need to fix this :)

bin/asinit
```diff
@@ -132,8 +145,20 @@ rl.question(colors.white("Do you want to proceed?") + " [Y/n] ", answer => {
...
+if (cliOptions.options.yes) {
+  createProject("y");
+} else {
+  const rl = require("readline").createInterface({
+    input: process.stdin,
+    output: process.stdout
+  });
+  rl.question(colors.white("Do you want to proceed?") + " [Y/n] ", result => {
+    createProject(result);
+    rl.close();
```
dcodeIO: Can close before creating the project here

bin/asinit
```diff
@@ -19,12 +37,7 @@ function printHelp() {
   process.exit(0);
 }
 
-const rl = require("readline").createInterface({
-  input: process.stdin,
-  output: process.stdout
-});
-
-const projectDir = path.resolve(process.argv[2]);
+const projectDir = path.resolve(cliOptions.arguments[2]);
```
dcodeIO: Does this work? I believe the options util strips the first two components (node and script path).

jtenner: I thought so too, but I checked the output of the parse function manually.

jtenner:
```ts
const cliOptions = options.parse(process.argv, asinitOptions);
console.log(cliOptions.arguments);
process.exit(0);
```

Outputs:

```
[
  'C:\\Program Files\\nodejs\\node.exe',
  'C:\\Users\\Joshua Tenner\\projects\\test-project\\node_modules\\assemblyscript\\bin\\asinit',
  'thePath'
]
```

bin/asinit
```diff
@@ -6,8 +6,26 @@ const fs = require("fs");
...
+  },
+  "yes": {
+    "category": "General",
+    "description": "Skips asking the user if they want to setup an AssemblyScript project.",
+    "type": "b",
+    "alias": "y"
+  }
+};
+
+const cliOptions = options.parse(process.argv, asinitOptions);
```
dcodeIO: Parse is typically called with `process.argv.slice(2)` so it doesn't attempt to do anything with node or script path

bin/asinit
dcodeIO: I see now, see the other comment :)

bin/asinit
jtenner: That's why. Got it.

bin/asinit
jtenner: Okay.

dcodeIO:
What happens now if no first argument is provided, like just `asinit`?

jtenner:
```
PS C:\Users\Joshua Tenner\projects\test-project> npx asinit
internal/validators.js:120
    throw new ERR_INVALID_ARG_TYPE(name, 'string', value);
    ^

TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received undefined
    at validateString (internal/validators.js:120:11)
    at Object.resolve (path.js:139:9)
    at Object.<anonymous> (C:\Users\Joshua Tenner\projects\assemblyscript\bin\asinit:40:25)
    at Module._compile (internal/modules/cjs/loader.js:1147:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1167:10)
    at Module.load (internal/modules/cjs/loader.js:996:32)
    at Function.Module._load (internal/modules/cjs/loader.js:896:14)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:71:12)
    at internal/main/run_main_module.js:17:47 {
  code: 'ERR_INVALID_ARG_TYPE'
}
```

jtenner:
We should make that better. :)

jtenner:
Also, could it be possible to default to `"."` as the path instead of requiring passing a path?

bin/asinit
```diff
@@ -6,8 +6,26 @@ const fs = require("fs");
...
+const asinitOptions = {
+  "help": {
+    "category": "General",
+    "description": "Prints a help message.",
+    "type": "b",
+    "alias": "h"
+  },
+  "yes": {
+    "category": "General",
+    "description": "Skips asking the user if they want to setup an AssemblyScript project.",
```
webmaster128: I'd define this a little more future proof: "Answers all quetions with their default options for non-interactive usage."

jtenner:
All comments have been addressed. This is ready for review again.

dcodeIO:
Suggesting to make it so that the help message is shown if zero non-option arguments have been provided, similar to what asc and other tools are doing.

webmaster128:
> Suggesting to make it so that the help message is shown if zero non-option arguments have been provided, similar to what asc and other tools are doing.

Such that

```
asinit --yes     # defaults to .
asinit .         # runs interactively
asinit           # fails with help message
```

? That would be a somewhat strange compisition of flags and positional argument. So I guess we'd have to give up the default value for the folder.

No strong opinion here, but my intuition is that `asinit` should run interactively in the current folder.

jtenner:
I'm with @webmaster128 on this opinion, but I'm probably going to change it so that it shows the help screen with 0 arguments.

dcodeIO:
What I'm proposing is

```
asinit --yes     # fails with help message
asinit           # fails with help message
asinit .         # runs interactively
asinit --yes .   # runs without asking
```

The check is basically `!arguments.length`, ignoring any options.

webmaster128:
Okay, yeah, this is the _giving up  the default value for the folder_ route. I'm not a big fan of this, but it is consistent and makes sense in code and to the user. Go for it.

jtenner:
Sorry for the delay. I had other things to work on.

I verified the proposal works by testing each input as suggested by @dcodeIO .

```
asinit --yes     # fails with help message
asinit           # fails with help message
asinit .         # runs interactively
asinit --yes .   # runs without asking
```

dcodeIO:
Thanks! But please reset the changes to package-lock before I merge :)

jtenner:
This should be all set now. Sorry for the trouble. Merge? 🙂

Issue #1165, opened 2020-03-14
Unresolved const type causing assertion failure


DuncanUszkay1:
When you create a constant with an non-existent type, and refer to it in another constant, you get an assertion failed error. The compilation would fail either way of course, but since the assertion failure provides no additional information without inspecting the source it's a bit of a hassle to debug. In our case it was just a typo in the type name.

Here's a minimal example:
```typescript
const INNER: fake = 1;
const OUTER: i32 = INNER;
```
Output upon compilation:
```typescript
ERROR: AssertionError: assertion failed
    at assert (/Users/duncanuszkay/src/github.com/Shopify/as-fork/assemblyscript/std/portable/index.js:184:9)
    at Compiler.compileIdentifierExpression (/Users/duncanuszkay/src/github.com/Shopify/as-fork/assemblyscript/src/compiler.ts:7631:9)
    at Compiler.compileExpression (/Users/duncanuszkay/src/github.com/Shopify/as-fork/assemblyscript/src/compiler.ts:3153:21)
    at Compiler.compileGlobal (/Users/duncanuszkay/src/github.com/Shopify/as-fork/assemblyscript/src/compiler.ts:940:25)
    at Compiler.compileTopLevelStatement (/Users/duncanuszkay/src/github.com/Shopify/as-fork/assemblyscript/src/compiler.ts:1748:20)
    at Compiler.compileFile (/Users/duncanuszkay/src/github.com/Shopify/as-fork/assemblyscript/src/compiler.ts:804:12)
    at Compiler.compile (/Users/duncanuszkay/src/github.com/Shopify/as-fork/assemblyscript/src/compiler.ts:406:14)
    at Object.compile (/Users/duncanuszkay/src/github.com/Shopify/as-fork/assemblyscript/src/index.ts:190:32)
    at /Users/duncanuszkay/src/github.com/Shopify/as-fork/assemblyscript/cli/asc.js:577:31
    at measure (/Users/duncanuszkay/src/github.com/Shopify/as-fork/assemblyscript/cli/asc.js:1068:3)
```

Did a little bit of digging in the source and found out what's going on:
We're failing an assertion in the compilation stage because it thinks that the type of `INNER` is `void`
```typescript
// in compiler.ts
// when compiling the INNER identifier used by OUTER
if (!this.compileGlobal(<Global>target)) { // reports; not yet compiled if a static   field
  return module.unreachable();
}
let type = (<Global>target).type;
[...]

PR #1202, opened 2020-03-31
Fail early when a top level statement fails to resolve

DuncanUszkay1:
Issue: https://github.com/AssemblyScript/assemblyscript/issues/1165

There's more context in the issue itself. Basically we could fail to compile a global top level statement, and we would just keep going compiling the next statement. This caused an assertion failure.

This PR makes the compileTopLevelStatement return a boolean to indicate that it has failed to compile, which is then used to halt compilation early and show an error message. The advantage of this approach is that we don't run into assertion failures later on, and instead show an error that is descriptive to the user.

### Example
```typescript
const INNER: fake = 1;
const OUTER: i32 = INNER;
```

On master we get:
```
    ...stacktrace
```

With this branch we get:
```
ERROR TS2304: Cannot find name 'fake'.

 const INNER: fake = 1;
              ~~~~
 in main.ts(5,14)

ERROR: 1 compile error(s)
    at Object.main (/Users/duncanuszkay/src/github.com/Shopify/as-fork/assemblyscript/cli/asc.js:614:21)
    at /Users/duncanuszkay/src/github.com/Shopify/as-fork/assemblyscript/bin/asc:21:47
```

### DIFF
--- a/src/compiler.ts
+++ b/src/compiler.ts
@@ -908,7 +908,7 @@ export class Compiler extends DiagnosticEmitter {
     var flow = startFunction.flow;
     this.currentFlow = flow;
     for (let statements = file.source.statements, i = 0, k = statements.length; i < k; ++i) {
-      this.compileTopLevelStatement(statements[i], startFunctionBody);
+      if(!this.compileTopLevelStatement(statements[i], startFunctionBody)) return;
     }
     this.finishAutoreleases(flow, startFunctionBody);
     // no need to insert unreachable since last statement should have done that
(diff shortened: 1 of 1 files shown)


dcodeIO:
Suggesting to fix the assertion at the place it originates, since otherwise we can't provide proper error highlighting (of all errors to fix in a file) in IDEs. The red markers would just stop at some point.

DuncanUszkay1:
Ah Right, I wasn't thinking about the IDE case. I'll try a different approach then

PR #1203, opened 2020-03-31
Implement Array#flat()

jtenner:
Turns out this was a bit simpler than I thought.

Review please? :)


std/assembly/array.ts
```diff
@@ -493,6 +493,68 @@ export class Array<T> {
...
+    var dataStart = __alloc(byteLength, idof<ArrayBuffer>());
+
+    // create the return value and initialize it
+    var result = __alloc(offsetof<T>(), idof<T>());
+    store<i32>(result, size, offsetof<T>("length_"));
+
+    // byteLength, dataStart, and buffer are all readonly
+    store<i32>(changetype<usize>(result), byteLength, offsetof<T>("byteLength"));
+    store<usize>(changetype<usize>(result), dataStart, offsetof<T>("dataStart"));
+    store<usize>(changetype<usize>(result), __retain(dataStart), offsetof<T>("buffer"));
```
MaxGraey, line 522: `changetype<usize>(result)` is unnecessary due to `__alloc` has `usize` return type

std/assembly/array.ts
```diff
@@ -493,6 +493,68 @@ export class Array<T> {
...
+      memory.copy(
+        dataStart + (<usize>resultIndex << usize(alignof<valueof<T>>())),
+        load<usize>(child, offsetof<T>("dataStart")),
+        <usize>childDataLength,
+      );
+
+      // advance the result length
+      let childLength = childDataLength >> <i32>alignof<valueof<T>>();
+      resultIndex += childLength - 1;
+    }
```
MaxGraey, line 546: it look like you could simplify this:
```ts
var resultIndex = 0;
for (let i = 0; i < length; i++) {
   ...
   // resultIndex++; don't need this

   resultIndex += childLength;
}
  ...
}
```

std/assembly/array.ts
```diff
@@ -493,6 +493,68 @@ export class Array<T> {
...
+
+    // if the `valueof<T>` type is managed, we must call __retain() on each reference
+    if (isManaged<valueof<T>>()) {
+      for (let i = 0; i < size; i++) {
+        __retain(load<usize>(dataStart + (<usize>i << usize(alignof<valueof<T>>()))));
```
MaxGraey, line 551:
```suggestion
        __retain(load<usize>(dataStart + (<usize>i << alignof<valueof<T>>())));
```

tests/compiler/std/array.ts
```diff
@@ -1038,3 +1038,25 @@ function assertSortedDefault<T>(arr: Array<T>): void {
...
+
+  var flatStringTarget: Array<Array<string | null>> = [["one"], ["two", null, "three"], ["four", "five", "six"], ["seven"]];
+  var stringResult = flatStringTarget.flat();
+  var expected = ["one", "two", null, "three", "four", "five", "six", "seven"];
+  for (let i = 0; i < expected.length; i++) {
+    assert(stringResult[i] == expected[i]);
+  }
+  assert(stringResult.length == 8);
+
+  let testArray: i32[][] = [[], [], [], [], [], [], [], []];
```
MaxGraey, line 1060:
```suggestion
  let testArray: i32[][] = [[], []];
```

tests/compiler/std/array.ts
```diff
@@ -1038,3 +1038,25 @@ function assertSortedDefault<T>(arr: Array<T>): void {
...
+  }
+  assert(result.length == 10);
+
+  var flatStringTarget: Array<Array<string | null>> = [["one"], ["two", null, "three"], ["four", "five", "six"], ["seven"]];
+  var stringResult = flatStringTarget.flat();
+  var expected = ["one", "two", null, "three", "four", "five", "six", "seven"];
+  for (let i = 0; i < expected.length; i++) {
+    assert(stringResult[i] == expected[i]);
+  }
+  assert(stringResult.length == 8);
```
MaxGraey, line 1058: plz move this before loop

tests/compiler/std/array.ts
```diff
@@ -1038,3 +1038,25 @@ function assertSortedDefault<T>(arr: Array<T>): void {
...
 // Unleak globals
 __release(changetype<usize>(arr));
+
+
+// Array#flat //////////////////////////////////////////////////////////////////////////////////
+{
+  var flatTarget: i32[][] = [[0], [1, 2, 3], [4, 5, 6], [7, 8, 9]];
+  var result = flatTarget.flat();
+  for (let i = 0; i < 10; i++) {
+    assert(result[i] == i, (i).toString());
```
MaxGraey, line 1048:
```suggestion
    assert(result[i] == i);
```

std/assembly/array.ts
```diff
@@ -493,6 +493,68 @@ export class Array<T> {
...
+
+    // byteLength, dataStart, and buffer are all readonly
+    store<i32>(changetype<usize>(result), byteLength, offsetof<T>("byteLength"));
+    store<usize>(changetype<usize>(result), dataStart, offsetof<T>("dataStart"));
+    store<usize>(changetype<usize>(result), __retain(dataStart), offsetof<T>("buffer"));
+
+    // set the elements
+    var resultIndex = -1;
+    for (let i = 0; i < length; i++) { // for each child
+      let child = load<usize>(selfDataStart + (i << alignof<T>()));
```
MaxGraey, line 527:
```suggestion
      let child = load<usize>(selfDataStart + (<usize>i << alignof<T>()));
```

std/assembly/array.ts
```diff
@@ -493,6 +493,68 @@ export class Array<T> {
...
+      // ignore null arrays
+      if (child == 0) continue;
+
+      // move to the next array index
+      resultIndex++;
+
+      // copy the underlying buffer data to the result buffer
+      let childDataLength = load<i32>(child, offsetof<T>("byteLength"));
+      memory.copy(
+        dataStart + (<usize>resultIndex << usize(alignof<valueof<T>>())),
```
MaxGraey, line 538:
```suggestion
        dataStart + (<usize>resultIndex << alignof<valueof<T>>()),
```

std/assembly/array.ts
jtenner: Oh right. I just ported over the old code from a previous pull request. Looks like a lot of improvements can be made :)

std/assembly/array.ts
```diff
@@ -493,6 +493,68 @@ export class Array<T> {
     return <string>unreachable();
   }
 
+  flat(): T {
+    if (!isArray<T>()) {
+      ERROR("Cannot call flat() on Array<T> where T is not an Array.");
```
MaxGraey, line 498: I'm not sure we should restrict this. What about just return shallow clone of array? Btw it will be follow spec

std/assembly/array.ts
jtenner: This was to suppress a typescript error in the IDE.

MaxGraey: use `@ts-ignore`

std/assembly/array.ts
jtenner: This change was to prevent an IDE error.

std/assembly/array.ts
jtenner: This is impossible because the return type must be known at compile time.

`valueof<T>[]` will throw an error, and we cannot be spec compliant in this way.

MaxGraey:
```ts
if (!isArray<T>) {
   return this.slice();
}
```
?

std/assembly/array.ts
jtenner: image

Even using ts-ignore, the problems don't go away.

@dcodeIO what kind of solution should we implement here?

std/assembly/array.ts
jtenner: This should be all set :)

std/assembly/array.ts
jtenner: actually, that part was required by the algorithm. I made a compromise and refactored the algorithm. Please check it again.

tests/compiler/std/array.ts
jtenner: This should be all set.

std/assembly/array.ts
jtenner: Should be all set.

std/assembly/array.ts
jtenner: Please see discussion above.

std/assembly/array.ts
jtenner: In the case of `u8[]`, calling flat will return `u8`, and not `u8[]`.

Returning `this.slice() as T` will return `u8` instead of `u8[]`.

tests/compiler/std/array.ts
jtenner: This should be all set.

tests/compiler/std/array.ts
jtenner: Yes this is all set.

std/assembly/array.ts
jtenner: @dcodeIO any thoughts?

std/assembly/array.ts
dcodeIO: Looks like a TS bug, hmm. As if it'd interpret that as `i < <alignof` expecting an `<alignof>` type assertion. Strange.

std/assembly/array.ts
MaxGraey: Ideally it should define as:
```ts
flat<U>(this: U[]): U[] {
   ...
}
```
But I'm not sure it support for now

std/assembly/array.ts
dcodeIO: Seems to happen on `<<` only. Minimal repro:

```ts
function alignof<T>(): number {
  return 1;
}

function foo<T>(): number {
  return 1 << alignof<T[]>();
}
```

Might have something to do with how they rescan on potentially ambiguous `<` sequences that might be a type assertion or not.

MaxGraey: So it ts lang server issue?

dcodeIO: https://github.com/microsoft/TypeScript/issues/37760

jtenner: does this result in less optimal code, and should I find an alternative?

tests/compiler/std/array.ts
jtenner: This should be all set now.

std/assembly/array.ts
jtenner: we need to support conditional return types for this to work. If `T` isn't an array, then it should return `Array<T>` instead of `T`.

jtenner: Suggesting to leave as-is until more work is done on the compiler directly.

jtenner:
Still looking for feedback. Any thoughts?

std/assembly/array.ts
dcodeIO: Suggesting to leave it with broken syntax highlighting for now, if there are no other issues. Otherwise, like if there are also lint errors, perhaps add a TODO linking to the TS bug and use a workaround.

PR #1213, opened 2020-04-10
Fix CI once again

dcodeIO:
Have made what I hope are a few improvements to the underlying action, and this PR uses them.


jtenner:
Does this pull request fix an Issue I'm experiencing? Every night, it runs a publish event on my `jtenner/master` branch, and I get an email.

dcodeIO:
Added something in the last commit so it only tries to publish from 'AssemblyScript/assemblyscript'. The job might still run, though.

jtenner:
Thank you for addressing my comments!

PR #1214, opened 2020-04-10
Fix option property access

jtenner:
When I was porting this module to TypeScript for a personal project, (and referencing it via the apache license,) I found this error.


jtenner:
Oh. I see. There's an `options` object there. That's what typescript was complaining about.

dcodeIO:
The fix seemed fine, that intended?

jtenner:
Upon further inspection, it does look like a bug. I got confused by typescript error reporting.

PR #1218, opened 2020-04-13
Decouple the parser from the program

jtenner:
This pull request aims to make the parser more accessible for others by decoupling the program from the parser by forcing the `Parser` class to accept two arrays instead.

This would probably be a breaking change for users of the `Parser` class, but my guess is that people weren't using the Parser class directly to begin with.


src/parser.ts
```diff
@@ -111,9 +110,12 @@ export class Parser extends DiagnosticEmitter {
   dependees: Map<string, Source> = new Map();
 
   /** Constructs a new parser. */
-  constructor(program: Program) {
-    super(program.diagnostics);
-    this.program = program;
+  constructor(
+    public diagnostics: DiagnosticMessage[] = [],
```
dcodeIO, line 114: This re-redefines the base class's field, so should remove the `public` modifier.

src/parser.ts
```diff
@@ -111,9 +110,12 @@ export class Parser extends DiagnosticEmitter {
...
+    public sources: Source[] = []
+  ) {
+    super(diagnostics);
+
```
dcodeIO, line 118: nit: empty line

src/parser.ts
```diff
@@ -111,9 +110,12 @@ export class Parser extends DiagnosticEmitter {
   dependees: Map<string, Source> = new Map();
 
   /** Constructs a new parser. */
-  constructor(program: Program) {
-    super(program.diagnostics);
-    this.program = program;
+  constructor(
+    public diagnostics: DiagnosticMessage[] = [],
+    public sources: Source[] = []
```
dcodeIO, line 115: Perhaps make this one a simple parameter as well, assigning in the constructor, so all the fields can be documented together above?

src/parser.ts
jtenner: Just a clarification question:

I should remove the `public` keyword and set the `this.diagnostics` property. right?

dcodeIO: Just removing the `public` keywords should be enough, since the argument is forward to the `super` constructor which sets `this.diagnostics`.

jtenner: All set!

src/parser.ts
jtenner: Whoops. All set.

src/parser.ts
jtenner: Yep. This is fine. All set!

src/parser.ts
```diff
@@ -109,11 +104,16 @@ export class Parser extends DiagnosticEmitter {
...
   /** Constructs a new parser. */
-  constructor(program: Program) {
-    super(program.diagnostics);
-    this.program = program;
+  constructor(
+    diagnostics: DiagnosticMessage[] | null = null,
+    sources: Source[] | null = null
+  ) {
+    super(diagnostics);
+    this.sources = sources == null ? new Array<Source>() : sources;
```
dcodeIO, line 116:
```suggestion
    this.sources = sources ? sources : new Array<Source>();
```

Mostly curious since you might have chosen to write it this way due to a null checking issue. Is there one?

jtenner: I was following the code style in the parent class. I would have coded it `sources || new Array<Source>()`.

jtenner:
All set!

PR #1220, opened 2020-04-13
Switch to initializer property

jtenner:
This pull request is more of a question than a suggestion.

`EnumValueDeclaration` extends `VariableLikeDeclaration` and the `initializer` property remains unused. This change simply removes the `value` property and switches to use the `initializer` property, reducing memory usage for each `EnumValueDeclaration`.

Thoughts?

### DIFF
--- a/src/ast.ts
+++ b/src/ast.ts
@@ -685,7 +685,7 @@ export abstract class Node {
 
   static createEnumValueDeclaration(
     name: IdentifierExpression,
-    value: Expression | null,
+    initializer: Expression | null,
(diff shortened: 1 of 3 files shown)


src/ast.ts
```diff
@@ -694,7 +694,7 @@ export abstract class Node {
     node.range = range;
     node.flags = flags;
     node.name = name;
-    node.value = value;
+    node.initializer = initializer;
```
dcodeIO, line 697: Nice!

PR #1222, opened 2020-04-13
Export astbuilder

### LINKED ISSUE #1212: Expose ASTBuilder
Is it possible to expose the ASTBuilder class so that external projects can use it?

Could we add another project to the lib directory?

jtenner:
As discussed on slack. This software would be useful in some of our projects and for orm codegen.

For instance, `as-pect` has had lots of trouble with debugging sources when transforms are included, so this would enable developers to output transformed source code to inspect certain kinds of problems with transforms.

cc @willemneal because he definitely wanted this too.

Fixes #1212 as a first step.

Something more should be done.

### DIFF
--- a/src/extra/ast.ts
+++ b/src/extra/ast.ts
@@ -905,7 +905,7 @@ export class ASTBuilder {
       sb.push("class");
     }
     var typeParameters = node.typeParameters;
-    if (typeParameters && typeParameters.length) {
+    if (typeParameters != null && typeParameters.length > 0) {
       sb.push("<");
       this.visitTypeParameter(typeParameters[0]);
       for (let i = 1, k = typeParameters.length; i < k; ++i) {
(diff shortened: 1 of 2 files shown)


willemneal:
You definitely don't have to do this, but it would also be a good opportunity to abstract the visitor part of the ASTBuilder so that it extends a base visitor.  This way external visitors don't have to recreate it and it'll ensure that it's up to date with the compiler since it'll be used in the testing of parser.

Here is the most abstract visitor I made: https://github.com/willemneal/visitor-as/blob/master/src/visitor.ts

I used that so that you can call `visitor.visit(collection of nodes)`

If you don't want to I could, I can take a crack at this tomorrow.

jtenner:
It was advised to me that we should wait for vtables, per the slack conversation. He said this was a good "first step".

dcodeIO:
Thinking about this, since we are not using the ASTBuilder in Wasm, it might as well be fine to abstract it now, as long as it compiles to Wasm and has no obvious issues anymore after vtables have landed.

jtenner:
I think I've run out of time for this today. Perhaps @willemneal could take the helm on this request then?

dcodeIO:
Merging so we have it exposed now. Feel free to think about abstracting it in a useful way :)

PR #1229, opened 2020-04-17
Fix / Enhancement for Omitted Object Literals

### LINKED ISSUE #1226: Add Check for Omitted fields in compileObjectLiteral
Currently, `compileObjectLiteral` in `compiler.ts` it is missing to check which fields are not present, and either error if that's non-null and has no initializer, pick the initializer or make it 0.

This was found, and discussed in a Slack discussion. @dcodeIO has context on this as well :smile: :+1:

torch2424:
closes #1226

This fixes the bug where omitted Object literals would just give random memory from the runtime. By introducing an error for class fields that cannot be omitted.

This Enhances Object literals, by allowing the [basic AS Types](https://docs.assemblyscript.org/basics/types). To be omitted, with default values of zero :smile: 

TODO: Either we add support for nullable types to default to null and/or add support for making a new instance of a class if the constructor has no parameters in this same PR. Or we can do that work in a seperate PR :smile: 

cc @dcodeIO @jtenner 

P.S. Yay! My first real AS Compiler PR :smile: :tada: 

# Screenshots

Screenshot from 2020-04-16 17-45-32
Screenshot from 2020-04-16 17-45-06
Screenshot from 2020-04-16 17-44-29

### DIFF
--- a/src/compiler.ts
+++ b/src/compiler.ts
@@ -8701,7 +8701,7 @@ export class Compiler extends DiagnosticEmitter {
       }
     }
 
-    // Iterate through the members definted in our expression
+    // Iterate through the members defined in our expression
     for (let i = 0, k = numNames; i < k; ++i) {
       let member = members ? members.get(names[i].text) : null;
       if (!member || member.kind != ElementKind.FIELD) {
@@ -8720,11 +8720,11 @@ export class Compiler extends DiagnosticEmitter {
         expr = this.makeRetain(expr);
       }
       exprs.push(this.module.store( // TODO: handle setters as well
-       fieldType.byteSize,
-       this.module.local_get(tempLocal.index, this.options.nativeSizeType),
-       expr,
-       fieldType.toNativeType(),
-       fieldInstance.memoryOffset
+        fieldType.byteSize,
+        this.module.local_get(tempLocal.index, this.options.nativeSizeType),
+        expr,
+        fieldType.toNativeType(),
+        fieldInstance.memoryOffset
       ));
 
       // This member is no longer omitted, so delete from our omitted fields
@@ -8735,8 +8735,9 @@ export class Compiler extends DiagnosticEmitter {
 
     // Iterate through the remaining omittedClassFieldMembers.
     if (members) {
-      for(let _values = Set_values(omittedClassFieldMembers), i = 0, v = _values.length; i < v; ++i) {
-        let omittedMemberKey = _values[i];
+
+      for (let _values = Set_values(omittedClassFieldMembers), j = 0, l = _values.length; j < l; ++j) {
+        let omittedMemberKey = _values[j];
         let member = assert(members.get(omittedMemberKey));
         
         let fieldInstance = <Field>member;
@@ -8744,15 +8745,14 @@ export class Compiler extends DiagnosticEmitter {
 
         if (fieldType.is(TypeFlags.REFERENCE) && fieldType.classReference !== null) {
           // TODO: Check if it is a class, with a default value (constructor with no params).
-          // TODO: Check if it can be null, and set to null
-
-          // Otherwise, error
-          this.error(
-            DiagnosticCode.Property_0_is_missing_in_type_1_but_required_in_type_2,
-            expression.range, "<object>", classReference.toString()
-          );
-          hasErrors = true;
-          continue;
+          if (!fieldType.is(TypeFlags.NULLABLE)) {
+            this.error(
+              DiagnosticCode.Property_0_is_missing_in_type_1_but_required_in_type_2,
+              expression.range, "<object>", classReference.toString()
+            );
+            hasErrors = true;
+            continue;
+          }
         }
 
         switch(fieldType.kind) {
@@ -8771,15 +8771,14 @@ export class Compiler extends DiagnosticEmitter {
           case TypeKind.F32: 
           case TypeKind.F64: {
             exprs.push(this.module.store( // TODO: handle setters as well
-             fieldType.byteSize,
-             this.module.local_get(tempLocal.index, this.options.nativeSizeType),
-             this.makeZero(fieldType),
-             fieldType.toNativeType(),
-             fieldInstance.memoryOffset
+              fieldType.byteSize,
+              this.module.local_get(tempLocal.index, this.options.nativeSizeType),
+              this.makeZero(fieldType),
+              fieldType.toNativeType(),
+              fieldInstance.memoryOffset
             ));
             continue;
           }
-          default: {}
         }
 
         // Otherwise, error
--- a/src/compiler.ts
+++ b/src/compiler.ts
@@ -8688,6 +8688,12 @@ export class Compiler extends DiagnosticEmitter {
       ? flow.getAutoreleaseLocal(classReference.type)
       : flow.getTempLocal(classReference.type);
     assert(numNames == values.length);
+    if (names.length == 0) {
+      console.log('expression names', names);
+      console.log('class members', members);
+    }
+
+    // Currently iterates through the passed names, should instead iterate through the members
     for (let i = 0, k = numNames; i < k; ++i) {
       let member = members ? members.get(names[i].text) : null;
       if (!member || member.kind != ElementKind.FIELD) {
--- a/src/compiler.ts
+++ b/src/compiler.ts
@@ -8693,7 +8693,9 @@ export class Compiler extends DiagnosticEmitter {
       console.log('class members', members);
     }
 
-    // Currently iterates through the passed names, should instead iterate through the members
+    // torch2424: Currently iterates through the passed names, should instead iterate through the members
+    
+
     for (let i = 0, k = numNames; i < k; ++i) {
       let member = members ? members.get(names[i].text) : null;
       if (!member || member.kind != ElementKind.FIELD) {
--- a/src/compiler.ts
+++ b/src/compiler.ts
@@ -8690,13 +8690,13 @@ export class Compiler extends DiagnosticEmitter {
     assert(numNames == values.length);
 
     // Assume all class fields will be omitted, and add them to our omitted list
-    let omittedClassFieldMembers = new Map<string, string>();
+    let omittedClassFieldMembers = new Set<string>();
     if (members) {
       for (let _keys = Map_keys(members), i = 0, k = _keys.length; i < k; ++i) {
         let memberKey = _keys[i];
         let member = assert(members.get(memberKey));
         if (member !== null && member.kind == ElementKind.FIELD) {
-          omittedClassFieldMembers.set(member.name, '');
+          omittedClassFieldMembers.add(member.name);
         }
       }
     }
@@ -8714,10 +8714,15 @@ export class Compiler extends DiagnosticEmitter {
       }
       let fieldInstance = <Field>member;
       let fieldType = fieldInstance.type;
+
+      let expr = this.compileExpression(values[i], fieldType, Constraints.CONV_IMPLICIT | Constraints.WILL_RETAIN);
+      if (isManaged && fieldType.isManaged && !this.skippedAutoreleases.has(expr)) {
+        expr = this.makeRetain(expr);
+      }
       exprs.push(this.module.store( // TODO: handle setters as well
        fieldType.byteSize,
        this.module.local_get(tempLocal.index, this.options.nativeSizeType),
-       this.compileExpression(values[i], fieldInstance.type, Constraints.CONV_IMPLICIT),
+       expr,
        fieldType.toNativeType(),
        fieldInstance.memoryOffset
       ));
@@ -8730,8 +8735,8 @@ export class Compiler extends DiagnosticEmitter {
 
     // Iterate through the remaining omittedClassFieldMembers.
     if (members) {
-      for(let _keys = Map_keys(omittedClassFieldMembers), i = 0, k = _keys.length; i < k; ++i) {
-        let omittedMemberKey = _keys[i];
+      for(let _values = Set_values(omittedClassFieldMembers), i = 0, v = _values.length; i < v; ++i) {
+        let omittedMemberKey = _values[i];
         let member = assert(members.get(omittedMemberKey));
         
         let fieldInstance = <Field>member;
@@ -8743,15 +8748,15 @@ export class Compiler extends DiagnosticEmitter {
 
           // Otherwise, error
           this.error(
-            DiagnosticCode.Object_literal_is_missing_class_member_fields_that_must_be_defined,
-            expression.range, classReference.toString()
+            DiagnosticCode.Property_0_is_missing_in_type_1_but_required_in_type_2,
+            expression.range, "<object>", classReference.toString()
           );
           hasErrors = true;
           continue;
         }
 
         switch(fieldType.kind) {
-          // i32 Types
+          // Number Types (and Number alias types)
           case TypeKind.I8:
           case TypeKind.I16:
           case TypeKind.I32:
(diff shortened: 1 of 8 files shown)


torch2424:
Oops, seems like I am failing the self-compilation tests :upside_down_face:  

@dcodeIO did you get things self compiling? Should I go back and fix these? Shouldn't be too hard, all I think I really need to do is get rid of the phat arraw function I added :smile: But how do I test the self compiler from my machine? :thinking:

src/compiler.ts
```diff
@@ -8682,12 +8682,36 @@ export class Compiler extends DiagnosticEmitter {
...
     var flow = this.currentFlow;
     var tempLocal = isManaged
       ? flow.getAutoreleaseLocal(classReference.type)
       : flow.getTempLocal(classReference.type);
     assert(numNames == values.length);
+
+    // Assume all class fields will be omitted, and add them to our omitted list
+    var omittedClassFieldMembers = new Map();
+    if (members) {
+      for (let memberKey of members.keys()) {
```
dcodeIO, line 8695:
```suggestion
      for (let _keys = Map_keys(members), i = 0, k = _keys.length; i < k; ++i) {
        let memberKey = _keys[i];
```
Necessary because we don't have iterators (and thus no `for ... of`) yet.

src/compiler.ts
```diff
@@ -8682,12 +8682,36 @@ export class Compiler extends DiagnosticEmitter {
...
+        let member = members.get(memberKey);
+        if (member && member.kind == ElementKind.FIELD) {
+          omittedClassFieldMembers.set(member.name, '');
```
dcodeIO, line 8698: Wondering if this can be a `Set`, if initial values are empty?

src/compiler.ts
```diff
@@ -8682,12 +8682,36 @@ export class Compiler extends DiagnosticEmitter {
...
       ? flow.getAutoreleaseLocal(classReference.type)
       : flow.getTempLocal(classReference.type);
     assert(numNames == values.length);
+
+    // Assume all class fields will be omitted, and add them to our omitted list
+    var omittedClassFieldMembers = new Map();
+    if (members) {
+      for (let memberKey of members.keys()) {
+        let member = members.get(memberKey);
+        if (member && member.kind == ElementKind.FIELD) {
```
dcodeIO, line 8697:
```suggestion
        if (member !== null && member.kind == ElementKind.FIELD) {
```
That's necessary due to some obscurities when compiling logical expressions. This way, all parts are `bool`.

src/compiler.ts
```diff
@@ -8682,12 +8682,36 @@ export class Compiler extends DiagnosticEmitter {
...
+    // Define a function to add a new expression on our Object literal
+    let storeObjectLiteralField = (fieldInstance: Field, fieldType: Type, valueExpr: ExpressionRef) => {
+      exprs.push(this.module.store( // TODO: handle setters as well
+        fieldType.byteSize,
+        this.module.local_get(tempLocal.index, this.options.nativeSizeType),
+        valueExpr,
+        fieldType.toNativeType(),
+        fieldInstance.memoryOffset
+      ));
+    };
```
dcodeIO, line 8712: Suggesting to inline this to match other code. Also, having this in-line avoids an indirect call.

src/compiler.ts
```diff
@@ -8700,27 +8724,114 @@ export class Compiler extends DiagnosticEmitter {
...
+
+      // This member is no longer omitted, so delete from our omitted fields
+      omittedClassFieldMembers.delete(member.name);
     }
     this.currentType = classReference.type.nonNullableType;
     if (hasErrors) return module.unreachable();
 
+    // Iterate through the remaining omittedClassFieldMembers.
+    if (members) {
+      for(let omittedClassFieldMemberKey of omittedClassFieldMembers.keys()) {
```
dcodeIO, line 8741:
```suggestion
      for (let _keys = Map_keys(omittedClassFieldMemberKey), i = 0, k = _keys.length; i < k; ++i) {
        let omittedClassFieldMemberKey = _keys[i];
```

src/compiler.ts
```diff
@@ -8700,27 +8724,114 @@ export class Compiler extends DiagnosticEmitter {
...
+        let member = members.get(omittedClassFieldMemberKey);
```
dcodeIO, line 8742:
```suggestion
        let member = assert(members.get(omittedClassFieldMemberKey));
```
This is mostly for TypeScript, where the result may be `undefined`. Since the key is in `omittedClassFields` here, we know that this will never be the case.

src/compiler.ts
```diff
@@ -8700,27 +8724,114 @@ export class Compiler extends DiagnosticEmitter {
...
+        
+        if(member) {
```
dcodeIO, line 8744:
```suggestion
        if (member) {
```

src/compiler.ts
```diff
@@ -8700,27 +8724,114 @@ export class Compiler extends DiagnosticEmitter {
...
+
+          let fieldInstance = <Field>member;
+          let fieldType = fieldInstance.type;
+
+          if (fieldType.classReference) {
```
dcodeIO, line 8749:
```suggestion
          if (fieldType.is(TypeFlags.REFERENCE) && fieldType.classReference !== null) {
```
There may be types that have a `classReference` (at some point), but are not actually classes, like the wrapper class `I32` of `i32`.

src/compiler.ts
```diff
@@ -8700,27 +8724,114 @@ export class Compiler extends DiagnosticEmitter {
...
+            // TODO: Check if it is a class, with a default value (constructor with no params).
+            // TODO: Check if it can be null, and set to null
+
+            // Otherwise, error
+            this.error(
+              DiagnosticCode.Object_literal_is_missing_class_member_fields_that_must_be_defined,
```
dcodeIO, line 8755: This looks like a custom error message. What I typically do is look through https://github.com/microsoft/TypeScript/blob/master/src/compiler/diagnosticMessages.json and check if there is an existing one we can reuse, if we don't have it already. Only if there is none we can reuse, I add a new one.

dcodeIO:
Regarding self-compilation: Yes, this is indeed part of CI now. The relevant scripts are `npm run asbuild` to build it, and `npm run astest` to test it. Unless on top of the vtable branch, the test will not do much and ultimately fail, which is expected, while on top of the vtable branch, it is good enough to output a simple module in text format.

MaxGraey:
Isn't this relate to #1155 ?

src/compiler.ts
torch2424: @dcodeIO I looked through the errors (CTRL+F for "literal") and I couldn't find anything that matched. :smile: 

I feel like this error will be needed, even after we add support for nullablity and things, since I expect some classes will have funny stuff with their constructors and things? :thinking: I purposely tried to make it vague enough to be like, Hey, your missing a field that you need to add manually haha!

torch2424:
@dcodeIO Thanks for the quick review! :smile: :+1:  Gonna start addressing comments in a little bit!

@MaxGraey Ummmm, so I am not too sure? I think it's similar, but a different issue? I'll let @dcodeIO answer that.

src/compiler.ts
torch2424: Yeah I originally was thinking a set, but then I got distracted and used a map when I came back :upside_down_face: Will switch to a set! :smile:

src/compiler.ts
dcodeIO: Perhaps, if there is a field on the object that is not on the class type

> `"Property '{0}' does not exist on type '{1}'.": 2339`

and if there is a missing field

> `"Property '{0}' is missing in type '{1}' but required in type '{2}'.": 2741`

with `{1}` being `<object>` and `{2}` being the class type?

dcodeIO:
It's somewhat related conceptually, yeah. Let's say we call the (default) ctor on object creation, then the ctor may initialize some of these fields, so specifying these is not necessary even though the field is not nullable. However, we are not calling the ctor here even though we should.

Suggesting to do the following to be safe for now: As soon as the class has a constructor, disallow creating it from an object literal. This should avoid any issues, and we can look into actually calling ctors in a later PR.

src/compiler.ts
torch2424: Oh wait! I remember, I chose not to use a set, because Set.values() returns am Iterator in JS, but an Array in AS :upside_down_face:

dcodeIO: Yeah, `Map`s and `Set`s have the same problem there. For `Set`s, the helper is `Set_values`. Both can't be iterated in a uniform way in TS/AS yet (no `for ... of`).

torch2424: Oh wait ignore me, Map.keys() also returns an iterator haha!

jtenner:
It's unfortunate that we aren't like rust, and get to have traits where object literals might make sense.

torch2424:
@dcodeIO Made all requested changes, and asbuild works on my machine :smile: :+1: This is good to go!

Screenshot from 2020-04-17 15-53-45

src/compiler.ts
```diff
@@ -8682,12 +8682,26 @@ export class Compiler extends DiagnosticEmitter {
...
-    var exprs = new Array<ExpressionRef>(numNames + 2);
+    var exprs = new Array<ExpressionRef>();
     var flow = this.currentFlow;
     var tempLocal = isManaged
       ? flow.getAutoreleaseLocal(classReference.type)
       : flow.getTempLocal(classReference.type);
     assert(numNames == values.length);
+
+    // Assume all class fields will be omitted, and add them to our omitted list
+    let omittedClassFieldMembers = new Map<string, string>();
```
dcodeIO, line 8693:
```suggestion
    let omittedClassFieldMembers = new Set<string>();
```
Let's make this a `Set` and use `_values = Set_values` instead of `_keys = Map_keys` below.

src/compiler.ts
```diff
@@ -8700,27 +8714,122 @@ export class Compiler extends DiagnosticEmitter {
...
-        fieldType.toNativeType(),
-        fieldInstance.memoryOffset
-      );
+      exprs.push(this.module.store( // TODO: handle setters as well
+       fieldType.byteSize,
+       this.module.local_get(tempLocal.index, this.options.nativeSizeType),
+       this.compileExpression(values[i], fieldInstance.type, Constraints.CONV_IMPLICIT),
+       fieldType.toNativeType(),
+       fieldInstance.memoryOffset
+      ));
```
dcodeIO, line 8723: This looks problematic, but already was so before so nothing introduced by this PR. If `fieldType` is managed here, and the class is managed as well, we need to retain a reference that the instance now holds, ideally accounting for pre-retained values (from function calls or ctors) which we can trigger by `WILL_RETAIN` and checking `skippedAutoreleases`.

```suggestion
      let expr = this.compileExpression(values[i], fieldType, Constraints.CONV_IMPLICIT | Constraints.WILL_RETAIN);
      if (isManaged && fieldType.isManaged && !this.skippedAutoreleases.has(expr)) {
        expr = this.makeRetain(expr);
      }
      exprs.push(this.module.store( // TODO: handle setters as well
        fieldType.byteSize,
        this.module.local_get(tempLocal.index, this.options.nativeSizeType),
        expr,
        fieldType.toNativeType(),
        fieldInstance.memoryOffset
      ));
```

src/compiler.ts
```diff
@@ -8700,27 +8714,122 @@ export class Compiler extends DiagnosticEmitter {
...
+        let fieldInstance = <Field>member;
+        let fieldType = fieldInstance.type;
+
+        if (fieldType.is(TypeFlags.REFERENCE) && fieldType.classReference !== null) {
+          // TODO: Check if it is a class, with a default value (constructor with no params).
+          // TODO: Check if it can be null, and set to null
+
+          // Otherwise, error
+          this.error(
+            DiagnosticCode.Object_literal_is_missing_class_member_fields_that_must_be_defined,
```
dcodeIO, line 8746: I made a suggesting for reusable error messages in a comment below. What do you think of it?

src/compiler.ts
```diff
@@ -8700,27 +8714,122 @@ export class Compiler extends DiagnosticEmitter {
...
+          case TypeKind.U8:
+          case TypeKind.U16:
+          case TypeKind.U32:
+          case TypeKind.USIZE: 
+          case TypeKind.ISIZE:
+          case TypeKind.BOOL: {
+            exprs.push(this.module.store( // TODO: handle setters as well
+             fieldType.byteSize,
+             this.module.local_get(tempLocal.index, this.options.nativeSizeType),
+             this.module.i32(0),
```
dcodeIO, line 8767: Most of this can be simplified by using 

```suggestion
             this.makeZero(fieldType),
```

so we don't need all these cases.

jtenner:
Nice work @torch2424

src/compiler.ts
torch2424: Dope! Yeah that sounds good, I will use those error messages! :smile:

torch2424:
Thanks @jtenner ! :smile: :tada: 

And made the requested changes @dcodeIO ! :smile: 

Screenshot from 2020-04-20 11-31-32
Screenshot from 2020-04-20 11-31-15
Screenshot from 2020-04-20 11-30-07

src/compiler.ts
```diff
@@ -8700,27 +8714,94 @@ export class Compiler extends DiagnosticEmitter {
...
+      let expr = this.compileExpression(values[i], fieldType, Constraints.CONV_IMPLICIT | Constraints.WILL_RETAIN);
+      if (isManaged && fieldType.isManaged && !this.skippedAutoreleases.has(expr)) {
+        expr = this.makeRetain(expr);
+      }
+      exprs.push(this.module.store( // TODO: handle setters as well
+       fieldType.byteSize,
+       this.module.local_get(tempLocal.index, this.options.nativeSizeType),
+       expr,
+       fieldType.toNativeType(),
+       fieldInstance.memoryOffset
```
dcodeIO, line 8727: nit: indentation

src/compiler.ts
```diff
@@ -8700,27 +8714,94 @@ export class Compiler extends DiagnosticEmitter {
...
+      // This member is no longer omitted, so delete from our omitted fields
+      omittedClassFieldMembers.delete(member.name);
     }
     this.currentType = classReference.type.nonNullableType;
     if (hasErrors) return module.unreachable();
 
+    // Iterate through the remaining omittedClassFieldMembers.
+    if (members) {
+      for(let _values = Set_values(omittedClassFieldMembers), i = 0, v = _values.length; i < v; ++i) {
+        let omittedMemberKey = _values[i];
```
dcodeIO, line 8739: nit: indentation (and other code tends to use j/l)
```suggestion
      for (let _values = Set_values(omittedClassFieldMembers), j = 0, l = _values.length; j < l; ++j) {
        let omittedMemberKey = _values[j];
```

src/compiler.ts
```diff
@@ -8700,27 +8714,94 @@ export class Compiler extends DiagnosticEmitter {
...
+          // TODO: Check if it is a class, with a default value (constructor with no params).
+          // TODO: Check if it can be null, and set to null
+
+          // Otherwise, error
+          this.error(
+            DiagnosticCode.Property_0_is_missing_in_type_1_but_required_in_type_2,
+            expression.range, "<object>", classReference.toString()
+          );
+          hasErrors = true;
+          continue;
```
dcodeIO, line 8755: Reminds me that you asked for this for your use case, and this can be supported by only erroring if the field's type is not nullable, and otherwise letting it store a `0` below:

```suggestion
          if (!fieldType.is(TypeFlags.NULLABLE)) {
            this.error(
              DiagnosticCode.Property_0_is_missing_in_type_1_but_required_in_type_2,
              expression.range, "<object>", classReference.toString()
            );
            hasErrors = true;
            continue;
          }
```

src/compiler.ts
```diff
@@ -8700,27 +8714,94 @@ export class Compiler extends DiagnosticEmitter {
...
+            exprs.push(this.module.store( // TODO: handle setters as well
+             fieldType.byteSize,
+             this.module.local_get(tempLocal.index, this.options.nativeSizeType),
+             this.makeZero(fieldType),
+             fieldType.toNativeType(),
+             fieldInstance.memoryOffset
+            ));
+            continue;
+          }
+          default: {}
```
dcodeIO, line 8782: nit: This line can be removed.

src/compiler.ts
```diff
@@ -8700,27 +8714,94 @@ export class Compiler extends DiagnosticEmitter {
...
+          case TypeKind.I64:
+          case TypeKind.U64: 
+          case TypeKind.F32: 
+          case TypeKind.F64: {
+            exprs.push(this.module.store( // TODO: handle setters as well
+             fieldType.byteSize,
+             this.module.local_get(tempLocal.index, this.options.nativeSizeType),
+             this.makeZero(fieldType),
+             fieldType.toNativeType(),
+             fieldInstance.memoryOffset
```
dcodeIO, line 8778: nit: indentation

src/compiler.ts
```diff
@@ -8682,12 +8682,26 @@ export class Compiler extends DiagnosticEmitter {
...
+      for (let _keys = Map_keys(members), i = 0, k = _keys.length; i < k; ++i) {
+        let memberKey = _keys[i];
+        let member = assert(members.get(memberKey));
+        if (member !== null && member.kind == ElementKind.FIELD) {
+          omittedClassFieldMembers.add(member.name);
+        }
+      }
+    }
+
+    // Iterate through the members definted in our expression
```
dcodeIO, line 8704: nit: typo in "defined"

src/compiler.ts
torch2424: Oooooo nice! Thank you for suggesting that quick fix for support! :smile: :+1:

src/compiler.ts
torch2424: My Vim was not very friendly to me :joy:

torch2424:
@dcodeIO Made requested changes, tests still pass, asbuild still working :smile: :+1:

dcodeIO:
Thank you! :)

torch2424:
Yayyyy! :smile: @dcodeIO Thank you very much for the review. Literally tweeting this out now :joy:

Issue #1003, opened 2019-12-12
Consider a file extension other than .ts


lastmjs:
Hi,

I'm the author of [Zwitterion](https://github.com/lastmjs/zwitterion), and I'm currently attempting to add support for AssemblyScript. Zwitterion's goal is to allow transpiling (to JS) or compiling (to Wasm) any language for front-end browser development. Languages are detected based on their file extension. This is very simple and allows Zwitterion to differentiate between JavaScript (.js), TypeScript (.ts), Rust (.rs), Wasm (.wasm), etc.

AssemblyScript not having its own file extension makes this rather difficult. Now, besides that use case, I think there are many other reasons why having a separate file extension for AssemblyScript makes sense. Static analysis tools and developer understanding come to mind. The ES modules spec allows for arbitrary file extensions, so this should not be a problem there. Though there is controversy over these issues, I personally think it's clear that any extension should be allowed in a module path. Deno.js has dealt with these issues, I believe creating a VS Code plugin that allows for .ts extensions (it just stops the type error). That may have changed recently, but they've been thinking this through as well.

Sorry for rambling. I hope a separate file extension for AssemblyScript will be considered. Thanks!

jtenner:
This would be useful for a language server in vscode too!

dcodeIO:
This is something we might consider in the future, but at the current state, that is without our own language server, reusing the .ts extension appears to be the best we can reasonably do to provide a good dev experience.

There are a couple of indicators, though, that tools can use to determine what's AS code, like

* If `package.json` has an `ascMain`, that's pointing at an AS entry file inside of a directory with other AS code
* Similarly, if `package.json` has an `asbuild` script, its subscripts point at an AS entry file
* AS code usually lives in `assembly/` if not configured otherwise

lastmjs:
> This is something we might consider in the future, but at the current state, that is without our own language server, reusing the .ts extension appears to be the best we can reasonably do to provide a good dev experience.

I understand. Though, isn't it simple to instruct VS Code to treat `.as` files as TypeScript files? I believe it is. Seems like that would be a much simpler way to get the benefits of TypeScript's static analysis in VS Code or similar editors

dcodeIO:
Last time I checked tsc hard coded the supported file extensions and the only way to change it was to maintain a fork. That was around the time of the asc prototype, though, don't know if it changed.

lastmjs:
[...]

PR #1231, opened 2020-04-17
Support alternative file extensions

dcodeIO:
This PR enables experimenting with alternative file extensions, by making the default extension overridable in asc, using the (temporary) `--extension` option. While this works, there are various follow-up issues obviously that need to be addressed eventually, like allowing multiple file extensions to be valid if one imports a foreign module's sources directly that are using `.ts`, but I guess this can wait until we know more (stdlib is extension agnostic).

Related: https://github.com/AssemblyScript/assemblyscript/issues/1003, https://github.com/AssemblyScript/meta/issues/32


jtenner:
I don't suppose we could do something like `--allow-extensions ts,asc` instead?

dcodeIO:
That's what I meant with additional work being necessary, but needs some refactoring around asc. For now this just enables experimenting with it.

PR #1236, opened 2020-04-21
Fix asc build

dcodeIO:
While investigating build times on CI I noticed that asc builds are 4.5mb and include Binaryen. This must have gone wrong when `binaryen.ready` was introduced a while ago, leading to CI spending most of the time in webpack, so this PR fixes that (faster builds, ~830kb asc).


jtenner:
Seriously. My webpack builds would take at least five minutes on my local pc. Thank you for this.

dcodeIO:
CI before: [13m 46s](https://github.com/AssemblyScript/assemblyscript/runs/606411581)
CI after:  [5m 29s](https://github.com/AssemblyScript/assemblyscript/runs/606433707)

PR #1237, opened 2020-04-22
Update Binaryen and utilize new ExpressionRunner API

dcodeIO:
This uses Binaryen's new [ExpressionRunner](https://github.com/WebAssembly/binaryen/pull/2702) that we have been waiting for to continue work on https://github.com/AssemblyScript/assemblyscript/pull/1111. Overall this works well and `untouched` output is much more true to the source now since we no longer have to precompute conditions, but can instead drop them once we evaluated their outcome.

But: There's one remaining issue as can be seen on the failing `std/typedarray` test, where it fails to evaluate expressions like

```wat
(block
 (drop
  (local.get $0)
 )
 (i32.const 0)
)
```

due to bailing early when encountering a `local.get` with an unknown value, here breaking statically precomputed `instanceof`. A bit unfortunate, but going to look into this.


dcodeIO:
Binaryen PR: https://github.com/WebAssembly/binaryen/pull/2787

dcodeIO:
With the new issue mentioned at the WG meeting fixed meanwhile, this is waiting for the PR above again. However, there have been objections on what the interpreter should do or not, so there's some uncertainty there.

dcodeIO:
Implemented a workaround now that mimics the PR linked above by suppressing certain drops to be emitted. Affected code is not really untouched, like before. Even though not optimal, it somewhat allows us to use the new runner API today.

PR #1240, opened 2020-04-25
Closures

DuncanUszkay1:
A Closure implementation which works as follows:
- Closures get an extra context argument appended to them
- While compiling a closure, we point to offsets from that context pointer in linear memory to access closed over variables
- After compiling a closure, we generate a class for that closure of this form:
```typescript
class Closure|X { // X is unique per closure 
  fn: usize // holds function ptr
  x: i32 // outside lexical scope which is referenced from the closure
  ...
}
```
- The object which is returned from defining a closure is an instantiation of the above object that was generated
- When this object is called in scope:
  - We copy in the values of the local scope into the `Closure|X` class instance
  - We load the closure context argument, followed by the rest of the normal arguments
  - We call the anonymous method
- When this object is converted into a function type (a type only signified by a type signature). This typically occurs because we're returning it from a function or passing it as an argument to a function
  - We copy in the values of the local scope into the `Closure|X` class instance one more time
  - We right-shift the bits of the address by 4, removing the 0s that result from 16-bit address alignment
  - We set the MSB
- When an anonymous function is called
  - We check to see if the MSB is set, that implies that it's a closure
  - Apply the same procedure as when it's called in scope, except without copying in local values (since we don't have access to them here)

There are still some improvements to be made, but I think we're in a pretty stable state and things seem to all be working.

cc @willemneal

## CURRENT LIMITATIONS
In the interest of time, I'm leaving in some limitations intentionally. They should be represented in the tests closure-limitations and closure-limitations-runtime. Here's a list for reference:
- writing to a closed value
- creating a closure within an anonymous function
- returning a closure from an exported function
- Calling a closure in the same scope it was defined in within a function, i.e.:
```typescript
class Foo {
  field: i32;
}
var compClosuresSameScope = (): void => {
  var foo: Foo = {
    field: 8
  }
  var withFoo1 = (): void => {
    foo.field += 1;
  }
  var withFoo2 = (): void => {
    withFoo1();
  }
}
```

### DIFF
--- a/src/compiler.ts
+++ b/src/compiler.ts
@@ -195,6 +195,8 @@ import {
   isPowerOf2
 } from "./util";
 
+const CLOSURE_TAG = 1 << 31;
+
 /** Compiler options. */
 export class Options {
 
@@ -1181,7 +1183,7 @@ export class Compiler extends DiagnosticEmitter {
         );
       }
       module.addGlobal(internalName, nativeType, true, this.makeZero(type));
-      if (type.isManaged && !this.skippedAutoreleases.has(initExpr)) initExpr = this.makeRetain(initExpr);
+      if (type.isManaged && !this.skippedAutoreleases.has(initExpr)) initExpr = this.makeRetain(initExpr, type);
       this.currentBody.push(
         module.global_set(internalName, initExpr)
       );
@@ -1373,7 +1375,8 @@ export class Compiler extends DiagnosticEmitter {
           stmts.push(
             module.local_set(index,
               this.makeRetain(
-                module.local_get(index, type.toNativeType())
+                module.local_get(index, type.toNativeType()),
+                type
               )
             )
           );
@@ -1676,7 +1679,7 @@ export class Compiler extends DiagnosticEmitter {
       module.local_get(0, nativeThisType),
       nativeValueType, instance.memoryOffset
     );
-    if (type.isManaged) valueExpr = this.makeRetain(valueExpr);
+    if (type.isManaged) valueExpr = this.makeRetain(valueExpr, type);
     instance.getterRef = module.addFunction(instance.internalGetterName, nativeThisType, nativeValueType, null, valueExpr);
     if (instance.setterRef) {
       instance.set(CommonFlags.COMPILED);
@@ -1713,9 +1716,9 @@ export class Compiler extends DiagnosticEmitter {
           ),
           module.block(null, [
             module.drop(
-              this.makeRetain(module.local_get(1, nativeValueType))
+              this.makeRetain(module.local_get(1, nativeValueType), type)
             ),
-            this.makeRelease(module.local_get(2, nativeValueType))
+            this.makeRelease(module.local_get(2, nativeValueType), type)
           ])
         ),
         module.local_get(1, nativeValueType)
@@ -1940,6 +1943,7 @@ export class Compiler extends DiagnosticEmitter {
     var tableBase = this.options.tableBase;
     if (!tableBase) tableBase = 1; // leave first elem blank
     index = tableBase + functionTable.length;
+    assert(u32(index) < u32(2147483648));
     functionTable.push(instance);
     instance.functionTableIndex = index;
     return index;
@@ -2752,6 +2756,24 @@ export class Compiler extends DiagnosticEmitter {
     // Remember that this flow returns
     flow.set(FlowFlags.RETURNS | FlowFlags.TERMINATES);
 
+    // Prevent returning a closure in an exported function, since interop with closures is
+    // not yet supported
+    var returnSignature = returnType.signatureReference;
+    if (returnSignature !== null && flow.parentFunction.is(CommonFlags.EXPORT)) {
+      var returnValueLocalIndex = flow.getTempLocal(returnType).index;
+      var nativeReturnType = returnType.toNativeType();
+      expr = module.flatten([
+        module.local_set(returnValueLocalIndex, expr),
+        this.ifClosure(
+          module.local_get(returnValueLocalIndex, nativeReturnType),
+          this.makeAbort(null, statement), // TODO: throw
+          module.nop()
+        ),
+        module.local_get(returnValueLocalIndex, nativeReturnType)
+      ], nativeReturnType)
+    }
+
+
     // If the last statement anyway, make it the block's return value
     if (isLastInBody && expr != 0 && returnType != Type.void) {
       if (!stmts.length) return expr;
@@ -3106,7 +3128,7 @@ export class Compiler extends DiagnosticEmitter {
               module.local_set(local.index,
                 initAutoreleaseSkipped
                   ? initExpr
-                  : this.makeRetain(initExpr)
+                  : this.makeRetain(initExpr, type)
               )
             );
           } else {
@@ -3501,7 +3523,7 @@ export class Compiler extends DiagnosticEmitter {
       // check if that worked, and if it didn't, keep the reference alive
       if (!this.skippedAutoreleases.has(expr)) {
         let index = this.tryUndoAutorelease(expr, flow);
-        if (index == -1) expr = this.makeRetain(expr);
+        if (index == -1) expr = this.makeRetain(expr, returnType);
         this.skippedAutoreleases.add(expr);
       }
     }
@@ -3549,6 +3571,24 @@ export class Compiler extends DiagnosticEmitter {
           expr = this.makeRuntimeNonNullCheck(expr, fromType, reportNode);
         }
         fromType = fromType.nonNullableType;
+      }
+      var toSignature = toType.signatureReference;
+      var fromSignature = fromType.signatureReference;
+      if (toSignature !== null && fromSignature !== null && fromSignature.externalEquals(toSignature) && fromType.is(TypeFlags.IN_SCOPE_CLOSURE)) {
+        // When we convert from the closure type into a function pointer, we first
+        // update the local copy of the scope with the newest values
+        var tempResult = this.currentFlow.getTempLocal(fromType);
+        var convertExpr = module.block(null, [
+          module.local_set(
+            tempResult.index,
+            expr
+          ),
+          this.injectClosedLocals(tempResult),
+          this.getClosureReference(module.local_get(tempResult.index, fromType.toNativeType()))
+        ], toType.toNativeType());
+
+        //this.currentFlow.freeTempLocal(tempResult);
+        return convertExpr;
       }
       if (fromType.isAssignableTo(toType)) { // downcast or same
         assert(fromType.kind == toType.kind);
@@ -5645,7 +5685,7 @@ export class Compiler extends DiagnosticEmitter {
               if (!leftAutoreleaseSkipped) {
                 retainLeftInElse = true;
               } else {
-                rightExpr = this.makeRetain(rightExpr);
+                rightExpr = this.makeRetain(rightExpr, rightType);
                 rightAutoreleaseSkipped = true;
               }
             } else if (!(constraints & Constraints.WILL_RETAIN)) { // otherwise keep right alive a little longer
@@ -5656,7 +5696,8 @@ export class Compiler extends DiagnosticEmitter {
             if (leftAutoreleaseSkipped) { // left turned out to be true'ish and is dropped
               rightStmts.unshift(
                 this.makeRelease(
-                  module.local_get(temp.index, leftType.toNativeType())
+                  module.local_get(temp.index, leftType.toNativeType()),
+                  leftType
                 )
               );
             }
@@ -5669,7 +5710,8 @@ export class Compiler extends DiagnosticEmitter {
               rightExpr,
               retainLeftInElse
                 ? this.makeRetain(
-                    module.local_get(temp.index, leftType.toNativeType())
+                    module.local_get(temp.index, leftType.toNativeType()),
+                    leftType
                   )
                 : module.local_get(temp.index, leftType.toNativeType())
             );
@@ -5749,7 +5791,7 @@ export class Compiler extends DiagnosticEmitter {
               if (!leftAutoreleaseSkipped) {
                 retainLeftInThen = true;
               } else {
-                rightExpr = this.makeRetain(rightExpr);
+                rightExpr = this.makeRetain(rightExpr, rightType);
                 rightAutoreleaseSkipped = true;
               }
             } else if (!(constraints & Constraints.WILL_RETAIN)) { // otherwise keep right alive a little longer
@@ -5762,7 +5804,8 @@ export class Compiler extends DiagnosticEmitter {
               // once implicit conversion with strings is performed and left is "", so:
               rightStmts.unshift(
                 this.makeRelease(
-                  module.local_get(temp.index, leftType.toNativeType())
+                  module.local_get(temp.index, leftType.toNativeType()),
+                  leftType
                 )
               );
             }
@@ -5774,7 +5817,8 @@ export class Compiler extends DiagnosticEmitter {
               this.makeIsTrueish(leftExpr, leftType),
               retainLeftInThen
                 ? this.makeRetain(
-                    module.local_get(temp.index, leftType.toNativeType())
+                    module.local_get(temp.index, leftType.toNativeType()),
+                    leftType
                   )
                 : module.local_get(temp.index, leftType.toNativeType()),
               rightExpr
@@ -6003,6 +6047,14 @@ export class Compiler extends DiagnosticEmitter {
     switch (target.kind) {
       case ElementKind.LOCAL: {
         let local = <Local>target;
+        if (local.closureContextOffset > 0) {
+          // TODO: ability to update closed over locals
+          this.error(
+            DiagnosticCode.Not_implemented,
+            valueExpression.range
+          );
+          return module.unreachable();
+        }
         if (flow.isLocalFlag(local.index, LocalFlags.CONSTANT, true)) {
           this.error(
             DiagnosticCode.Cannot_assign_to_0_because_it_is_a_constant_or_a_read_only_property,
@@ -6191,6 +6243,7 @@ export class Compiler extends DiagnosticEmitter {
         valueExpr = this.makeReplace(
           valueExpr,
           module.local_get(localIndex, type.toNativeType()),
+          type,
           alreadyRetained
         );
         if (tee) { // local = REPLACE(local, value)
@@ -6203,7 +6256,7 @@ export class Compiler extends DiagnosticEmitter {
       } else {
         flow.unsetLocalFlag(localIndex, LocalFlags.CONDITIONALLY_RETAINED);
         flow.setLocalFlag(localIndex, LocalFlags.RETAINED);
-        if (!alreadyRetained) valueExpr = this.makeRetain(valueExpr);
+        if (!alreadyRetained) valueExpr = this.makeRetain(valueExpr, type);
         if (tee) { // local = __retain(value, local)
           this.currentType = type;
           return module.local_tee(localIndex, valueExpr);
@@ -6247,6 +6300,7 @@ export class Compiler extends DiagnosticEmitter {
         this.makeReplace(
           valueExpr,
           module.global_get(global.internalName, nativeType),
+          type,
           alreadyRetained
         )
       );
@@ -6320,6 +6374,7 @@ export class Compiler extends DiagnosticEmitter {
                 module.local_get(tempThis.index, nativeThisType),
                 nativeFieldType, field.memoryOffset
               ),
+              fieldType,
               alreadyRetained
             ),
             nativeFieldType, field.memoryOffset
@@ -6337,6 +6392,7 @@ export class Compiler extends DiagnosticEmitter {
               module.local_get(tempThis.index, nativeThisType),
               nativeFieldType, field.memoryOffset
             ),
+            fieldType,
             alreadyRetained
           ),
           nativeFieldType, field.memoryOffset
@@ -6370,6 +6426,30 @@ export class Compiler extends DiagnosticEmitter {
         );
       }
     }
+  }
+
+  private injectClosedLocals(
+    closureContextLocal: Local
+  ): ExpressionRef {
+    var module = this.module;
+    var type = closureContextLocal.type;
+    var locals = assert(type.locals);
+    let _values = Map_values(locals);
+    var exprs = new Array<ExpressionRef>(_values.length);
+    assert(type.is(TypeFlags.REFERENCE));
+    for (let i = 0, k = _values.length; i < k; ++i) {
+      let local = unchecked(_values[i]);
+      let nativeType = local.type.toNativeType();
+      assert(local.closureContextOffset > 0);
+      exprs[i] = module.store(
+        local.type.byteSize,
+        module.local_get(closureContextLocal.index, this.options.nativeSizeType),
+        module.local_get(local.index, nativeType),
+        nativeType,
+        local.closureContextOffset
+      );
+    }
+    return module.block(null, exprs);
   }
 
   /** Compiles a call expression according to the specified context. */
@@ -6472,7 +6552,36 @@ export class Compiler extends DiagnosticEmitter {
       // indirect call: index argument with signature (non-generic, can't be inlined)
       case ElementKind.LOCAL: {
         let local = <Local>target;
-        signature = local.type.signatureReference;
+        let nativeSizeType = this.options.nativeSizeType;
+        signature = assert(local.type.signatureReference);
+        if (local.type.is(TypeFlags.IN_SCOPE_CLOSURE)) {
+          if (this.currentFlow.parentFunction.parent.kind == ElementKind.FUNCTION) {
+            this.error(
+              DiagnosticCode.Not_implemented,
+              expression.expression.range
+            );
+            return module.unreachable();
+          }
+          // If we're calling a local we know to be a closure, then we must still be in the creator functions
+          // scope. Because of this, we should update the values of locals that are still available
+          return module.block(null, [
+            this.injectClosedLocals(local),
+            this.compileCallIndirect(
+              assert(signature),
+              module.load(
+                local.type.byteSize,
+                local.type.is(TypeFlags.SIGNED),
+                module.local_get(local.index, nativeSizeType),
+                this.options.nativeSizeType,
+                0
+              ),
+              expression.args,
+              expression,
+              module.local_get(local.index, nativeSizeType),
+              contextualType == Type.void
+            )
+          ], signature.returnType.toNativeType());
+        }
         if (signature) {
           if (local.is(CommonFlags.INLINED)) {
             indexArg = module.i32(i64_low(local.constantIntegerValue));
@@ -6577,14 +6686,47 @@ export class Compiler extends DiagnosticEmitter {
         return module.unreachable();
       }
     }
-    return this.compileCallIndirect(
-      assert(signature), // FIXME: bootstrap can't see this yet
-      indexArg,
-      expression.args,
-      expression,
-      0,
-      contextualType == Type.void
-    );
+    // Once we get here, we have a function reference. With the new scheme, this function
+    // could possibly be a closure. So here we check to see if it's a closure, then apply
+    // the appropriate call logic
+    signature = assert(signature) // FIXME: asc can't see this yet
+    var returnType = signature.returnType;
+    var tempFunctionReferenceLocal = this.currentFlow.getTempLocal(this.options.usizeType);
+    var usize = this.options.nativeSizeType;
+    return module.block(null, [
+      module.local_set(tempFunctionReferenceLocal.index, indexArg),
+      this.ifClosure(
+        module.local_get(tempFunctionReferenceLocal.index, usize),
+        this.compileCallIndirect( // If this is a closure
+          signature.toClosureSignature(),
+          module.block(null, [
+            module.load(
+              4,
+              true,
+              this.getClosurePtr(
+                module.local_get(tempFunctionReferenceLocal.index, usize),
+              ),
+              usize,
+              0
+            ),
+          ], this.options.nativeSizeType),
+          expression.args,
+          expression,
+          this.getClosurePtr(
+            module.local_get(tempFunctionReferenceLocal.index, usize),
+          ),
+          contextualType == Type.void
+        ),
+        this.compileCallIndirect( // If this function isn't a closure
+          signature,
+          module.local_get(tempFunctionReferenceLocal.index, usize),
+          expression.args,
+          expression,
+          0,
+          contextualType == Type.void
+        )
+      )
+    ], constraints & Constraints.WILL_DROP ? contextualType.toNativeType() : returnType.toNativeType())
   }
 
   private compileCallExpressionBuiltin(
@@ -6832,7 +6974,7 @@ export class Compiler extends DiagnosticEmitter {
       if (flow.isNonnull(paramExpr, paramType)) flow.setLocalFlag(argumentLocal.index, LocalFlags.NONNULL);
       // inlining is aware of skipped autoreleases:
       if (paramType.isManaged) {
-        if (!this.skippedAutoreleases.has(paramExpr)) paramExpr = this.makeRetain(paramExpr);
+        if (!this.skippedAutoreleases.has(paramExpr)) paramExpr = this.makeRetain(paramExpr, paramType);
         flow.setLocalFlag(argumentLocal.index, LocalFlags.RETAINED);
       }
       body.unshift(
@@ -6876,7 +7018,7 @@ export class Compiler extends DiagnosticEmitter {
       if (flow.isNonnull(initExpr, initType)) flow.setLocalFlag(argumentLocal.index, LocalFlags.NONNULL);
       if (initType.isManaged) {
         flow.setLocalFlag(argumentLocal.index, LocalFlags.RETAINED);
-        if (!this.skippedAutoreleases.has(initExpr)) initExpr = this.makeRetain(initExpr);
+        if (!this.skippedAutoreleases.has(initExpr)) initExpr = this.makeRetain(initExpr, initType);
       }
       body.push(
         module.local_set(argumentLocal.index, initExpr)
@@ -6900,7 +7042,7 @@ export class Compiler extends DiagnosticEmitter {
     this.currentType = returnType;
     if (returnType.isManaged) {
       if (immediatelyDropped) {
-        expr = this.makeRelease(expr);
+        expr = this.makeRelease(expr, returnType);
         this.currentType = Type.void;
       }
     }
@@ -7237,16 +7379,76 @@ export class Compiler extends DiagnosticEmitter {
   // <reference-counting>
 
   /** Makes a retain call, retaining the expression's value. */
-  makeRetain(expr: ExpressionRef): ExpressionRef {
+  makeRetain(
+    expr: ExpressionRef,
+    type: Type | null = null
+  ): ExpressionRef {
+    var module = this.module;
+
     var retainInstance = this.program.retainInstance;
     this.compileFunction(retainInstance);
-    return this.module.call(retainInstance.internalName, [ expr ], this.options.nativeSizeType);
+    if (type !== null && type.isFunctionIndex) {
+      var exprLocal = this.currentFlow.getTempLocal(type);
+      var exprLocalIndex = exprLocal.index;
+      var nativeType = type.toNativeType();
+      var usize = this.options.nativeSizeType;
+      var functionRetainCall = module.block(null, [
+        module.local_set(exprLocalIndex, expr),
+        module.drop(
+          module.call(
+            retainInstance.internalName,
+            [
+              this.ifClosure(
+                module.local_get(exprLocalIndex, nativeType),
+                this.getClosurePtr(module.local_get(exprLocalIndex, nativeType)),
+                usize == NativeType.I32 ? module.i32(0) : module.i64(0)
+              )
+            ],
+            usize
+          )
+        ),
+        module.local_get(exprLocalIndex, nativeType)
+      ], nativeType);
+
+      //this.currentFlow.freeTempLocal(exprLocal);
+
+      return functionRetainCall;
+    }
+
+    return module.call(retainInstance.internalName, [ expr ], this.options.nativeSizeType);
   }
 
   /** Makes a release call, releasing the expression's value. Changes the current type to void.*/
-  makeRelease(expr: ExpressionRef): ExpressionRef {
+  makeRelease(expr: ExpressionRef, type: Type | null = null): ExpressionRef {
+    var module = this.module;
+
     var releaseInstance = this.program.releaseInstance;
     this.compileFunction(releaseInstance);
+
+    if (type !== null && type.isFunctionIndex) {
+      var exprLocal = this.currentFlow.getTempLocal(type);
+      var exprLocalIndex = exprLocal.index;
+      var nativeType = type.toNativeType()
+      var functionReleaseCall = module.block(null, [
+        module.local_set(exprLocalIndex, expr),
+        module.call(
+          releaseInstance.internalName,
+          [
+            this.ifClosure(
+              module.local_get(exprLocalIndex, nativeType),
+              this.getClosurePtr(module.local_get(exprLocalIndex, nativeType)),
+              this.options.nativeSizeType == NativeType.I32 ? module.i32(0) : module.i64(0)
+            )
+          ],
+          NativeType.None
+        )
+      ], NativeType.None);
+
+      // TODO: fix a bug in which this free causes some overwrites
+      //this.currentFlow.freeTempLocal(exprLocal);
+
+      return functionReleaseCall;
+    }
     return this.module.call(releaseInstance.internalName, [ expr ], NativeType.None);
   }
 
@@ -7256,6 +7458,8 @@ export class Compiler extends DiagnosticEmitter {
     newExpr: ExpressionRef,
     /** Old value being replaced. */
     oldExpr: ExpressionRef,
+    /** Expression type. */
+    type: Type,
     /** Whether the new value is already retained. */
     alreadyRetained: bool = false,
   ): ExpressionRef {
@@ -7268,7 +7472,7 @@ export class Compiler extends DiagnosticEmitter {
       let temp = flow.getTempLocal(this.options.usizeType, findUsedLocals(oldExpr));
       let ret = module.block(null, [
         module.local_set(temp.index, newExpr),
-        this.makeRelease(oldExpr),
+        this.makeRelease(oldExpr, type),
         module.local_get(temp.index, nativeSizeType)
       ], nativeSizeType);
       flow.freeTempLocal(temp);
@@ -7289,9 +7493,9 @@ export class Compiler extends DiagnosticEmitter {
           ),
           module.block(null, [
             module.local_set(temp1.index,
-              this.makeRetain(module.local_get(temp1.index, nativeSizeType))
+              this.makeRetain(module.local_get(temp1.index, nativeSizeType), type)
             ),
-            this.makeRelease(module.local_get(temp2.index, nativeSizeType))
+            this.makeRelease(module.local_get(temp2.index, nativeSizeType), type)
           ])
         ),
         module.local_get(temp1.index, nativeSizeType)
@@ -7398,7 +7602,7 @@ export class Compiler extends DiagnosticEmitter {
       // If it worked, autorelease in `outerFlow` instead
       ? this.makeAutorelease(expr, type, outerFlow)
       // If it didn't work, extend the lifetime into `outerFlow`
-      : this.makeAutorelease(this.makeRetain(expr), type, outerFlow);
+      : this.makeAutorelease(this.makeRetain(expr, type), type, outerFlow);
   }
 
   /** Performs any queued autoreleases in the specified flow. */
@@ -7428,7 +7632,8 @@ export class Compiler extends DiagnosticEmitter {
             if (finalize) flow.unsetLocalFlag(localIndex, LocalFlags.ANY_RETAINED);
             stmts.push(
               this.makeRelease(
-                module.local_get(localIndex, local.type.toNativeType())
+                module.local_get(localIndex, local.type.toNativeType()),
+                local.type
               )
             );
           }
@@ -7523,7 +7728,8 @@ export class Compiler extends DiagnosticEmitter {
       flow.unsetLocalFlag(localIndex, LocalFlags.ANY_RETAINED);
       stmts.push(
         this.makeRelease(
-          module.local_get(localIndex, local.type.toNativeType())
+          module.local_get(localIndex, local.type.toNativeType()),
+          local.type
         )
       );
     }
@@ -7660,7 +7866,7 @@ export class Compiler extends DiagnosticEmitter {
           this.currentType = returnType;
           if (returnType.isManaged) {
             if (immediatelyDropped) {
-              expr = this.makeRelease(expr);
+              expr = this.makeRelease(expr, returnType);
               this.currentType = Type.void;
             } else if (!skipAutorelease) {
               expr = this.makeAutorelease(expr, returnType);
@@ -7686,7 +7892,7 @@ export class Compiler extends DiagnosticEmitter {
     this.currentType = returnType;
     if (returnType.isManaged) {
       if (immediatelyDropped) {
-        expr = this.makeRelease(expr);
+        expr = this.makeRelease(expr, returnType);
         this.currentType = Type.void;
       } else if (!skipAutorelease) {
         expr = this.makeAutorelease(expr, returnType);
@@ -7799,7 +8005,7 @@ export class Compiler extends DiagnosticEmitter {
     this.currentType = returnType;
     if (returnType.isManaged) {
       if (immediatelyDropped) {
-        expr = this.makeRelease(expr);
+        expr = this.makeRelease(expr, returnType);
         this.currentType = Type.void;
       } else {
         expr = this.makeAutorelease(expr, returnType);
@@ -7979,24 +8185,137 @@ export class Compiler extends DiagnosticEmitter {
         prototype.name,
         prototype,
         null,
-        signature,
+        prototype.hasNestedDefinition ? signature.toClosureSignature() : signature,
         contextualTypeArguments
       );
       if (!this.compileFunction(instance)) return this.module.unreachable();
-      this.currentType = contextualSignature.type;
-
     // otherwise compile like a normal function
     } else {
       instance = this.resolver.resolveFunction(prototype, null, contextualTypeArguments);
       if (!instance) return this.module.unreachable();
       this.compileFunction(instance);
-      this.currentType = instance.signature.type;
     }
 
+    // if this anonymous function turns out to be a non-closure, recompile a version
+    // of the function without context, deleting the previous function
+    if (instance.closedLocals.size == 0 && instance.prototype.hasNestedDefinition) {
+      this.module.removeFunction(instance.internalName);
+      instance = new Function(
+        instance.prototype.name + "~nonClosure",
+        instance.prototype,
+        null,
+        instance.signature.toAnonymousSignature(),
+        contextualTypeArguments
+      );
+      this.compileFunction(instance);
+    }
+    this.currentType = instance.signature.type;
+
     var index = this.ensureFunctionTableEntry(instance); // reports
-    return index < 0
-      ? this.module.unreachable()
-      : this.module.i32(index);
+
+    if(index < 0) return this.module.unreachable();
+
+    var usize = this.options.usizeType;
+    var nativeUsize = this.options.nativeSizeType;
+    var wasm64 = nativeUsize == NativeType.I64;
+
+    if(instance.closedLocals.size > 0) {
+      this.warning(
+        DiagnosticCode.Closure_support_is_experimental,
+        instance.prototype.declaration.range
+      )
+
+      // Append the appropriate signature and flags for this closure type, then set it to currentType
+      this.currentType = Type.closure32;
+      this.currentType.signatureReference = instance.signature;
+
+      // create a local which will hold our closure
+      var tempLocal = flow.getAutoreleaseLocal(this.currentType);
+      var tempLocalIndex = tempLocal.index;
+
+      // copied closed locals into type
+      this.currentType.locals = instance.closedLocals;
+
+      const closureSize = instance.nextGlobalClosureOffset;
+
+      var allocInstance = this.program.allocInstance;
+      this.compileFunction(allocInstance);
+
+      var closureExpr = this.module.flatten([
+        this.module.local_set( // Allocate memory for the closure
+          tempLocalIndex,
+          this.makeRetain(
+            this.module.call(allocInstance.internalName, [
+              wasm64 ? this.module.i64(closureSize) : this.module.i32(closureSize),
+              wasm64 ? this.module.i64(0) : this.module.i32(0)
+            ], nativeUsize)
+          )
+        ),
+        this.module.store( // Store the function pointer at the first index
+          4,
+          this.module.local_get(tempLocalIndex, nativeUsize),
+          wasm64 ? this.module.i64(index) : this.module.i32(index),
+          nativeUsize,
+          0
+        ),
+        this.module.local_get(tempLocalIndex, nativeUsize) // load the closure locals index
+      ], nativeUsize);
+
+      //flow.freeTempLocal(tempLocal);
+
+      return closureExpr;
+    }
+
+    return wasm64 ? this.module.i64(index) : this.module.i32(index);
+  }
+
+  private ifClosure(
+    indexExpr: ExpressionRef,
+    thenExpr: ExpressionRef,
+    elseExpr: ExpressionRef
+  ): ExpressionRef {
+    var module = this.module;
+    var wasm64 = this.options.nativeSizeType == NativeType.I64
+
+    return module.if(
+      module.binary(
+        wasm64 ? BinaryOp.EqI64 : BinaryOp.EqI32,
+        module.binary(
+          wasm64 ? BinaryOp.AndI64 : BinaryOp.AndI32,
+          indexExpr,
+          wasm64 ? module.i64(CLOSURE_TAG) : module.i32(CLOSURE_TAG)
+        ),
+        wasm64 ? module.i64(CLOSURE_TAG) : module.i32(CLOSURE_TAG)
+      ),
+      thenExpr,
+      elseExpr
+    )
+  }
+
+  private getClosurePtr(closureExpr: ExpressionRef): ExpressionRef {
+    var module = this.module;
+    var wasm64 = this.options.nativeSizeType == NativeType.I64
+
+    return module.binary(
+      wasm64 ? BinaryOp.ShlI64 : BinaryOp.ShlI32,
+      closureExpr,
+      wasm64 ? module.i64(4) : module.i32(4)
+    )
+  }
+
+  private getClosureReference(closureExpr: ExpressionRef): ExpressionRef {
+    var module = this.module;
+    var wasm64 = this.options.nativeSizeType == NativeType.I64
+
+    return module.binary(
+      wasm64 ? BinaryOp.OrI64 : BinaryOp.OrI32,
+      module.binary(
+        wasm64 ? BinaryOp.ShrI64 : BinaryOp.ShrI32,
+        closureExpr,
+        wasm64 ? module.i64(4) : module.i32(4)
+      ),
+      wasm64 ? module.i64(CLOSURE_TAG) : module.i32(CLOSURE_TAG)
+    )
   }
 
   /** Makes sure the enclosing source file of the specified expression has been compiled. */
@@ -8134,6 +8453,20 @@ export class Compiler extends DiagnosticEmitter {
         let local = <Local>target;
         let localType = local.type;
         assert(localType != Type.void);
+        var localClosureContextOffset = local.closureContextOffset;
+        if (localClosureContextOffset > 0) {
+          let contextLocal = assert(flow.lookupLocal(CommonNames.this_));
+
+          // TODO: replace this with a class field access, once we are able to construct the class before
+          // compiling
+          return module.load(
+            local.type.byteSize,
+            local.type.is(TypeFlags.SIGNED),
+            this.module.local_get(contextLocal.index, this.options.nativeSizeType),
+            local.type.toNativeType(),
+            localClosureContextOffset
+          );
+        }
         if (local.is(CommonFlags.INLINED)) {
           return this.compileInlineConstant(local, contextualType, constraints);
         }
@@ -8577,7 +8910,7 @@ export class Compiler extends DiagnosticEmitter {
             : module.i32(i64_low(bufferAddress))
         ], expression);
         this.currentType = arrayType;
-        expr = this.makeRetain(expr);
+        expr = this.makeRetain(expr, arrayType);
         if (arrayType.isManaged) {
           if (!(constraints & Constraints.WILL_RETAIN)) {
             expr = this.makeAutorelease(expr, arrayType);
@@ -8639,7 +8972,7 @@ export class Compiler extends DiagnosticEmitter {
       if (isManaged) {
         // value = __retain(value)
         if (!this.skippedAutoreleases.has(valueExpr)) {
-          valueExpr = this.makeRetain(valueExpr);
+          valueExpr = this.makeRetain(valueExpr, elementType);
         }
       }
       // store<T>(tempData, value, immOffset)
@@ -8796,7 +9129,7 @@ export class Compiler extends DiagnosticEmitter {
       if (isManaged) {
         // value = __retain(value)
         if (!this.skippedAutoreleases.has(valueExpr)) {
-          valueExpr = this.makeRetain(valueExpr);
+          valueExpr = this.makeRetain(valueExpr, elementType);
         }
       }
       // store<T>(tempThis, value, immOffset)
@@ -9388,18 +9721,18 @@ export class Compiler extends DiagnosticEmitter {
 
     if (ifThenAutoreleaseSkipped != ifElseAutoreleaseSkipped) { // unify to both skipped
       if (!ifThenAutoreleaseSkipped) {
-        ifThenExpr = this.makeRetain(ifThenExpr);
+        ifThenExpr = this.makeRetain(ifThenExpr, commonType);
         ifThenAutoreleaseSkipped = true;
       } else {
-        ifElseExpr = this.makeRetain(ifElseExpr);
+        ifElseExpr = this.makeRetain(ifElseExpr, commonType);
         ifElseAutoreleaseSkipped = true;
       }
     } else if (!ifThenAutoreleaseSkipped && commonType.isManaged) { // keep alive a little longer
       if (constraints & Constraints.WILL_RETAIN) { // try to undo both
         let ifThenIndex = this.tryUndoAutorelease(ifThenExpr, ifThenFlow);
-        if (ifThenIndex == -1) ifThenExpr = this.makeRetain(ifThenExpr);
+        if (ifThenIndex == -1) ifThenExpr = this.makeRetain(ifThenExpr, commonType);
         let ifElseIndex = this.tryUndoAutorelease(ifElseExpr, ifElseFlow);
-        if (ifElseIndex == -1) ifElseExpr = this.makeRetain(ifElseExpr);
+        if (ifElseIndex == -1) ifElseExpr = this.makeRetain(ifElseExpr, commonType);
         ifThenAutoreleaseSkipped = true;
         ifElseAutoreleaseSkipped = true;
       } else {
@@ -9634,7 +9967,6 @@ export class Compiler extends DiagnosticEmitter {
         break;
       }
       default: {
-        assert(false);
         return module.unreachable();
       }
     }
@@ -10470,7 +10802,7 @@ export class Compiler extends DiagnosticEmitter {
             : 1 + parameterIndex, // this is local 0
           nativeFieldType
         );
-        if (fieldType.isManaged) initExpr = this.makeRetain(initExpr);
+        if (fieldType.isManaged) initExpr = this.makeRetain(initExpr, fieldType);
 
       // fall back to use initializer if present
       } else if (initializerNode) {
@@ -10478,7 +10810,7 @@ export class Compiler extends DiagnosticEmitter {
           Constraints.CONV_IMPLICIT | Constraints.WILL_RETAIN
         );
         if (fieldType.isManaged && !this.skippedAutoreleases.has(initExpr)) {
-          initExpr = this.makeRetain(initExpr);
+          initExpr = this.makeRetain(initExpr, fieldType);
         }
 
       // otherwise initialize with zero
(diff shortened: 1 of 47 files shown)


MaxGraey:
you should refresh all tests due to new builtin global var

src/compiler.ts
```diff
@@ -6405,12 +6453,58 @@ export class Compiler extends DiagnosticEmitter {
       case ElementKind.LOCAL: {
         let local = <Local>target;
         signature = local.type.signatureReference;
+        if(local.type.is(TypeFlags.CLOSURE)) {
```
DuncanUszkay1, line 6456: When the closure is still in the local scope we'll have access to their closure class still. We use information in the closure class to copy in values from the local scope before the call

src/compiler.ts
```diff
@@ -6405,12 +6453,58 @@ export class Compiler extends DiagnosticEmitter {
...
+          break;
+        }
         if (signature) {
           if (local.is(CommonFlags.INLINED)) {
             indexArg = module.i32(i64_low(local.constantIntegerValue));
           } else {
             indexArg = module.local_get(local.index, NativeType.I32);
           }
+          // If the 16 is divisible by 16, then it must be a closure, so we call it as such
+          indexArg = module.if(
```
DuncanUszkay1, line 6483: Outside of the local scope, we only have function pointers, we don't know ahead of time whether or not it'll be a closure. So instead we check to see if it's divisible by 16, and if it is assume its a pointer. We don't know what locals this closure uses, but we don't need to since the local scope it was referencing is inaccessible to us here anyway. We just set the global closure and assume the function will know how to read from that.

src/compiler.ts
```diff
@@ -3460,6 +3468,18 @@ export class Compiler extends DiagnosticEmitter {
...
       }
+      // When we convert from the closure type into a function pointer, we first
+      // update the local copy of the scope with the newest values
+      if(fromType.is(TypeFlags.CLOSURE)) {
+        //Retain the Closure Class memory so it isn't freed when we change it into a function ptr
+        if (!this.skippedAutoreleases.has(expr)) {
+          let index = this.tryUndoAutorelease(expr, this.currentFlow);
+          if (index == -1) expr = this.makeRetain(expr);
+          this.skippedAutoreleases.add(expr);
+        }
```
DuncanUszkay1, line 3479: I'm not sure how we'll eventually free the closure given that we lose track of its size- that's another major missing piece

DuncanUszkay1:
Had a conversation IRL with @willemneal and Nikolay about this approach. Here are the outcomes:
- Change the current approach of setting a global into passing an object alongside the function
- Call anonymous functions using an invoke method, regardless of closure status, so that we can remove the 16 bit alignment hack

Seems like otherwise we're on the right track

dcodeIO:
> Change the current approach of setting a global into passing an object alongside the function

Would that mean changing the signature of the function? Asking because doing so can become a problem where normal functions and closures can be used interchangeably.

> Call anonymous functions using an invoke method

Can you elaborate how that'd look? Wondering if it implies that we need one invoke function per unique function signature.

> so that we can remove the 16 bit alignment hack

An alternative suggestion to avoid the hack has been to use fat pointers for function references, i.e. an i64 with N bits closure address and M bits function index and alignments bits truncated. Has the advantage that signatures remain unchanged (all function refs are i64, even if not a closure), but the disadvantage that i64 on the boundary is not well supported yet.

willemneal:
This would not change the signature of the function rather, any path where an anonymous function could be a closure it must become an empty one with an invoke method.  For example, a function that could return a closure or a normal anon function would then expect the output to be a closure and then when compiling a call site the invoke method would be used.

With this method you wouldn't need to use the original or your new suggested trick.

dcodeIO:
Can you give me an example how this mechanism would look like? Let's say we have

```ts
function doAdd(fn: (a: i32, b: i32) => i32, a: i32, b: i32): i32 {
  return fn(a, b); // may be a closure, or may not be one
}
```

DuncanUszkay1:
>Can you elaborate how that'd look? Wondering if it implies that we need one invoke function per unique function signature.

Only function signatures that are called indirectly, but yes that is true with what we've proposed.

Here's a code transformation that expresses it:
### Original
```typescript
function doAdd(fn: (a: i32, b: i32) => i32, a: i32, b: i32): i32 {
  return fn(a, b); // may be a closure, or may not be one
}

function namedAdder(a: i32, b: i32): i32 {
  return a + b;
}

i = 1
j = 2
doAdd((a: i32, b: i32) => { return a + b + i + j }, 3, 4)
doAdd((a: i32, b: i32) => { return a + b }, 3, 4)
doAdd(namedAdder, 3, 4)
```
### Translated
```typescript
fn anonymous_1(a: i32, b: i32, i: i32, j: i32): i32 {
  return a + b + i + j;
}

fn anonymous_2(a: i32, b: i32): i32 {
  return a + b;
}

class GeneratedClosure1 {
  f: usize //points to anonymous_1
  i: i32
  j: i32

  invoke(a: i32, b: i32) {
    changetype<(a: i32, b: i32) => i32>(self.f)(a,b,self.i,self.j)
  }
}

class EmptyClosure { // Non-closure represented by an empty closure
  f: usize //points to anonymous_2

  invoke(a: i32, b: i32) { //needs one such method per indirect call function signature
    changetype<(a: i32, b: i32) => i32>(self.f)(a,b)
  }
}

function doAdd(fn: (a: i32, b: i32) => i32, a: i32, b: i32): i32 {
  return fn.invoke(a, b); // may be a closure, or may not be one
}

i = 1;
j = 2;
doAdd(new GeneratedClosure1(anonymous_1,i,j), 3, 4)
doAdd(new EmptyClosure(anonymous_2), 3, 4)
doAdd(new EmptyClosure(namedAdder), 3, 4)
```

This isn't exactly what would be done because this isn't being implemented through a code transformation but should be functionally the same.

What do you think?

DuncanUszkay1:
Thinking more about this I'll probably need an interface or similar to make that `fn.invoke(a,b)` call work 🤔

MaxGraey:
@DuncanUszkay1 Does it mean that any function (not closure) should wrap by `EmptyClosure` object?

DuncanUszkay1:
>Does it mean that any function (not closure) should wrap by EmptyClosure object

Any function that can be called indirectly, yes. They'd all have to belong to an interface with the appropriate invoke signature

DuncanUszkay1:
If we decide not to go the invoke route, we could continue down the global closure route and have a stack of closure objects to handle nested calls (we could also explicitly disallow nested calls with a not implemented, but likely we would want that feature eventually). With that approach we'd still need to figure out a way to differentiate closures, for instance by the fat pointer method @dcodeIO mentioned.

MaxGraey:
> Any function that can be called indirectly, yes

This make things more complicated if we want efficiency. Like extra analysis which try determine could we simplify this close without free vars (captured variables) to simple indirect call which highly required for father optimization like directive and inlining which very important especailly for eliminate unnecessary retain / release calls for ARC

DuncanUszkay1:
That's a fair point- I think if we want to be focused on efficiency we need to step away from thinking about it as a code transformation and think about it on the WASM level. Thinking about it like this the fat pointers solution is definitely appealing- it ends up being similar in that we have nearly the same data structure except the operative parts (the function pointer and the optionally null reference to scope) live on the stack as an i64 rather than in linear memory.

>but the disadvantage that i64 on the boundary is not well supported yet.

I think either direction we're blocked by something- with the class approach we need an interface-like mechanism which afaik doesn't exist currently. 

@willemneal any thoughts on this? Also I don't have Nikolay's @ but would love to hear his opinion

src/compiler.ts
DuncanUszkay1: This has been changed, now we just check to see if the function reference has its MSB set, in which case we assume its a closure.

DuncanUszkay1:
Talked about this a bit on Slack- here's the result:

- A function will be identified as a closure when it has it's MSB set. Functions will remain 32 bits long
  - [ ] This limits the number of possible functionts to 2147483647
  - [x] When a closure is assigned to a function reference, we bitshift it right by 4 (which doesn't destroy any information because of 16 bit alignment), then set the MSB to 1
  - [x] When a function reference turns out to be a closure (MSB is 1) we bit shift left by 4 to get the closure pointer, then get the function reference (also the first field of the closure, so we don't need to know what specific class the closure has)
- [ ] We'll indirectly ref count closures through their function pointers
- [ ] Attempting to define a closure within a closure will result in a `Not Implemented` (for now)
- [ ] When calling indirectly, we will save the value of the current global context in a local so that we don't lose its value when making the call.

Some things that weren't mentioned that also need to be fixed:
- [ ] let statements can't be closed over

DuncanUszkay1:
Should let statements be able to be closed over? Seems like given that they're supposed to be bound to the block they live in maybe not- Right now it won't even resolve if you try to reference a let variable in a closure

src/compiler.ts
```diff
@@ -1827,6 +1827,7 @@ export class Compiler extends DiagnosticEmitter {
     var tableBase = this.options.tableBase;
     if (!tableBase) tableBase = 1; // leave first elem blank
     index = tableBase + functionTable.length;
+    assert(index >= 1 << 31)
```
MaxGraey, line 1830: I guess it should be `assert(index <= ((1 << 31) >>> 0));`

DuncanUszkay1: oh right, whoops.

DuncanUszkay1:
Discovered a flaw in how I read from the closure context- I don't have access to the class offsets during compilation because we are unaware of the closure before compilation. I had a system in place which made an assumption about where a field would be stored in the class by making some assumptions, but this won't work in a case where you use the same closed variable twice, i.e:
```typescript
function createClosure(arg: i32): () => i32 {
  var f = (): i32 => { return arg * arg; }
}
```
Going to look to see how hard it'd be to fix this without the ScopeAnalyzer, then if there isn't a way that makes sense maybe look over the analyzer again to see if we ought to get that in first

DuncanUszkay1:
Last push should've fixed it- the offset is now being stuck onto the closedLocal as it's created, so we can reuse them.

It does however bring to mind the scope analyzer and whether or not this should be blocked on that or not

MaxGraey:
How about inherit `ClosedLocal` from `Local`?

DuncanUszkay1:
Only reason I didn't do that initially is that the constructor of `Local` hardcodes its type to `LOCAL` whereas we need the type `CLOSEDLOCAL`- and passing in an `ElementKind` argument to the constructor didn't seem right. You're probably right though and it should extend Local.

Not super opinionated on that though, definitely something I can come back to once I start refactoring/reorganizing when all the functionality is there

DuncanUszkay1:
Working on issues related to inlining the creation of these closures with a following immediate call. Example:
```typescript
export function runInline(arg: i32): i32 {
  return ((): i32 => { return arg; } )();
}
```
This doesn't work and produces a huge binary in the current iteration- this is because the index argument expression actually includes the creation of the closure in this case- so the fact that I use the indexArg multiple times is definitely a problem. Before this was fine, since when the closure is assigned to a local before being called the indexArg expression is just loading from a local index. I can't make the assumption that the indexArg is safe to call multiple times going forward.

src/compiler.ts
```diff
@@ -6512,14 +6598,64 @@ export class Compiler extends DiagnosticEmitter {
...
-      indexArg,
-      expression.arguments,
-      expression,
-      0,
-      contextualType == Type.void
-    );
+
+    var returnType = signature.returnType;
+    var tempResultLocal = this.currentFlow.getTempLocal(returnType);
+    var tempFunctionReferenceLocal = this.currentFlow.getTempLocal(returnType);
```
DuncanUszkay1, line 6604: Added locals here to accomplish two things:
- tempFunctionReferenceLocal is used to store the result of the index argument expression- we need to make sure this is only executed once, so we store it to a local before using it multiple times in the following lines
- tempResultLocal: after we finish the `call_indirect` invokation, we need to change the global closure value back to what it was before the call began, and then we need to return whatever the call returned- since we need to access something that's below the top of the stack, afaik there is no way to do that without a local

I think there's probably a way to do this with just one local, but wanted to get my latest changes up on the PR before finishing up for the day- I'll probably look at this again tomorrow.

src/compiler.ts
```diff
@@ -6512,14 +6598,64 @@ export class Compiler extends DiagnosticEmitter {
...
+
+    var callExpr = module.block(null, [
+      module.local_set(
+        tempFunctionReferenceLocal.index,
+        indexArg
+      ),
+      module.push(
+        module.global_get(BuiltinNames.global_closure, NativeType.I32)
+      ),
```
DuncanUszkay1, line 6613: Here we push the original closure context onto the stack, so that we can access it again after the call is done

src/compiler.ts
```diff
@@ -6512,14 +6598,64 @@ export class Compiler extends DiagnosticEmitter {
...
+
+    var returnType = signature.returnType;
+    var tempResultLocal = this.currentFlow.getTempLocal(returnType);
+    var tempFunctionReferenceLocal = this.currentFlow.getTempLocal(returnType);
+
+    var callExpr = module.block(null, [
+      module.local_set(
+        tempFunctionReferenceLocal.index,
+        indexArg
+      ),
```
DuncanUszkay1, line 6610: Here we execute the expression and get the value of the index argument into a local so we can access freely later.

src/compiler.ts
```diff
@@ -6512,14 +6598,64 @@ export class Compiler extends DiagnosticEmitter {
...
+        module.if(
+          module.binary(
+            BinaryOp.EqI32,
+            module.binary(
+              BinaryOp.AndI32,
+              module.local_get(tempFunctionReferenceLocal.index, NativeType.I32),
+              module.i32(1 << 31)
+            ),
+            module.i32(1 << 31)
+          ),
```
DuncanUszkay1, line 6625: This is checking for our special closure pointer which is in this binary format:
[1000][all but the last 4 bits of the context address]

A normal function would never have that MSB set because we instituted a limit below 1 << 31.

src/compiler.ts
```diff
@@ -6512,14 +6598,64 @@ export class Compiler extends DiagnosticEmitter {
...
+              )
+            ),
+            module.load( //TODO: support 8 byte addrs
+              4,
+              true,
+              module.global_get(BuiltinNames.global_closure, NativeType.I32),
+              NativeType.I32,
+              0
+            )
+          ], this.options.nativeSizeType),
```
DuncanUszkay1, line 6642: If it's a closure, move the closure context into the global closure context, then access the first field, which is always the function index. Otherwise we just put the index argument as-is onto the stack

src/compiler.ts
```diff
@@ -6512,14 +6598,64 @@ export class Compiler extends DiagnosticEmitter {
...
+          module.local_get(tempFunctionReferenceLocal.index, NativeType.I32),
+        ),
+        expression.arguments,
+        expression,
+        0,
+        contextualType == Type.void
+      ),
+      module.local_set(tempResultLocal.index, module.pop(returnType.toNativeType())),
+      module.global_set(BuiltinNames.global_closure, module.pop(NativeType.I32)),
+      module.local_get(tempResultLocal.index, returnType.toNativeType())
```
DuncanUszkay1, line 6652: The call is now completed- we save the result into a local, then our old global closure context value will remain on the stack and we can put it back. Then we load up the result local to leave the result on the stack as expected of a function call

src/program.ts
```diff
@@ -668,6 +668,47 @@ export class Program extends DiagnosticEmitter {
...
+
+  makeNativeClassPrototype(name: string, members: DeclarationStatement[]): ClassPrototype | null {
+    return this.initializeClass(
+      this.makeNativeClassDeclaration(name, members),
+      this.nativeFile,
+      [],
+      []
+    )
+  }
+
```
DuncanUszkay1, line 711: These let us generate some boilerplate IR for the class which is required to compile it, pointing the source at 'native' rather than a real file like in a non-generated source file.

src/program.ts
```diff
@@ -3079,7 +3149,26 @@ export class Function extends TypedElement {
...
+          DiagnosticCode.Not_implemented,
+          this.identifierNode.range, this.identifierNode.text
+        );
+        return null;
+      }
+      if(parentResult.kind == ElementKind.LOCAL) {
+        let local = changetype<Local>(parentResult)
+        if(this.closedLocals.has(local.name)) return assert(this.closedLocals.get(local.name))
+        var closedLocal = local.close(this.nextGlobalClosureOffset);
+        this.nextGlobalClosureOffset += local.type.byteSize;
```
DuncanUszkay1, line 3166: the `nextGlobalClosureOffset` hack is here because we haven't generated the class yet at this point- so based on an assumption on how the fields will be ordered and packed, we determine the offset of each local. 

This could be replaced if we had a mechanism for determining where we had closures before we began compiling

src/compiler.ts
```diff
@@ -189,6 +189,8 @@ import {
   makeMap
 } from "./util";
 
+const CLOSURE_FLAG = 1 << 31;
```
MaxGraey, line 192:
```suggestion
const CLOSURE_TAG = 1 << 31;
```

src/compiler.ts
```diff
@@ -6991,16 +7127,74 @@ export class Compiler extends DiagnosticEmitter {
   // <reference-counting>
 
   /** Makes a retain call, retaining the expression's value. */
-  makeRetain(expr: ExpressionRef): ExpressionRef {
+  makeRetain(
+    expr: ExpressionRef,
+    type: Type | null = null
```
DuncanUszkay1, line 7132: RC is now dependent on type- here's the new process:
- If the expression represents a function reference, which are now managed, we check to see if it's a Closure using the MSB, then retain/release the pointer (which needs to be shifted by 4 bits to get the 16-bit aligned pointer). For retaining, we also need to set this value back to the local we are retaining, so we need to shift the bits/reset the MSB before returning
- Otherwise, if the type isn't passed in (calls which could not reference a closure) or if the type isn't a function (a type represented by a signature only- closures which are represented by signatures and class references don't count. They are normal pointers and don't need this special treatment), so we just retain/release as normal

src/compiler.ts
```diff
@@ -7718,9 +7919,109 @@ export class Compiler extends DiagnosticEmitter {
...
+
+    return module.binary(
+      BinaryOp.OrI32,
+      module.binary(
+        BinaryOp.ShrI32,
+        closureExpr,
+        module.i32(4)
+      ),
+      module.i32(CLOSURE_FLAG)
+    )
```
DuncanUszkay1, line 8024: Helper methods for dealing with the new function pointer format, applying the appropriate shifts and setting the MSB etc

DuncanUszkay1: Might belong elsewhere 🤷  maybe in `module.ts`? Though this isn't tied to binaryen very closely so maybe not..

DuncanUszkay1:
Next Steps:
- [ ] Fix local freeing bug
- [ ] Create a complete representative test suite for all these changes 😅 
- [ ] More reviews

DuncanUszkay1:
Working through some validator bugs- did not realize that it was turned off by default

DuncanUszkay1:
Getting this now: `Fatal: IR must be flat: run --flatten beforehand (instructions must only have constant expressions, local.get, or unreachable as children, in assert-nonnull/testFn)` I assume that's binaryen yelling at us.

src/compiler.ts
```diff
@@ -6512,14 +6600,56 @@ export class Compiler extends DiagnosticEmitter {
...
+          0,
+          contextualType == Type.void
+        )
+      ),
+      module.global_set(BuiltinNames.global_closure, module.pop(NativeType.I32)),
+      module.local_get(tempResultLocal.index, returnType.toNativeType())
+    ], returnType.toNativeType())
+
+    this.currentFlow.freeTempLocal(tempResultLocal);
+    this.currentFlow.freeTempLocal(tempFunctionReferenceLocal);
```
MaxGraey, line 6650: As I remember `freeTempLocal` should have LIFO order according it initialization. cc @dcodeIO

DuncanUszkay1:
Looks like an optimization is failing- going to try and isolate which one and figure out what's going on.

DuncanUszkay1:
Based on an inspection of binaryen it's one of these optimizations:
Screen Shot 2020-05-08 at 11 46 47 AM

MaxGraey:
it may just error soundness. That's PR should enhance validation I guess:
https://github.com/WebAssembly/binaryen/pull/2589

MaxGraey:
@DuncanUszkay1 btw are you using custom pass pipeline? Because it pretty differ from ours:
https://github.com/AssemblyScript/assemblyscript/blob/master/src/module.ts#L1473

DuncanUszkay1:
Must just be out of date, I'll rebase

DuncanUszkay1:
Rebased, and things are now compiling. Unfortunately I'm running into issues with Binaryen validation and optimization- validation is hanging which [I have an issue up for](https://github.com/WebAssembly/binaryen/issues/2845), and optimization is failing in a way I don't quite understand (some context on that in the previous issue as well)

For now I'm going to try and isolate the pass which is causing the optimization failure in case it makes the problem obvious.

src/compiler.ts
```diff
@@ -6290,6 +6339,28 @@ export class Compiler extends DiagnosticEmitter {
...
+    var module = this.module;
+    var locals = type.locals!;
+    let _values = Map_values(locals);
+    var exprs = new Array<ExpressionRef>(_values.length);
+    for (let i = 0, k = _values.length; i < k; ++i) {
+      let local = unchecked(_values[i]);
+      let closureClass= type.classReference!
+      exprs[i] = module.store(
+        local.type.byteSize,
+        expr,
```
dcodeIO, line 6355: This line appears to reuse the same `expr` multiple times instead of utilizing a temporary local. If IR nodes are not unique, this typically leads to all sorts of weird issues down the road (optimizing one modifies all with IR becoming invalid).

src/program.ts
dcodeIO: Also looks like this should take alignment into account to prevent unaligned loads and stores.

tests/compiler/closure.ts
```diff
@@ -21,4 +21,51 @@ function testLet(): (value: i32) => i32 {
...
+//function nestedExecutionTest(arg: i32): i32 {
+  //var x = 7;
+  //var f = complexCreateClosure(arg);
+  //var g = (fn: (x3: i32) => i32): i32 => {
+    //var first = fn(arg);
+    //return x;
+  //}
+  //return g(f);
+//}
+//nestedExecutionTest(1);
```
DuncanUszkay1, line 33: Accidentally commited, just wanted to get up what I had while I'm still trying to get tests working

src/compiler.ts
DuncanUszkay1: Looks like I missed this one when you told me about this earlier- I'll do another scan through to find any other ones I may have missed

src/compiler.ts
```diff
@@ -6343,16 +6343,18 @@ export class Compiler extends DiagnosticEmitter {
     expr: ExpressionRef,
     type: Type
   ): ExpressionRef {
+    let closureContextLocal = this.currentFlow.getTempLocal(Type.i32)
```
dcodeIO, line 6346: Looks like the local doesn't become free'd again.

DuncanUszkay1: Is freeing locals necessary for functionality? I was under the impression that it was just an optimization to use fewer locals- I figured for now I'd just omit them until I get this compiling/validating/optimized and return to it

DuncanUszkay1:
Another task that needs to be done once I get back to a stable functional state:
- [ ] Ensure all locals are freed appropriately

src/compiler.ts
dcodeIO: Not strictly for functionality, but other code does this so we don't end up with huge untouched output.

dcodeIO:
Perhaps one more general comment on the code. I didn't check in detail, but I spotted quite a few helper functions to do stuff after the fact of compiling an expression. Using helpers is typically very tricky, because one must make sure that temporary locals do not conflict in unexpected ways. For example, if one passes a pre-compiled expression to a helper, that expression might be compiled in a way that it is using temporary locals, which are now free'd, and adding another temporary that will be set out-of-line can overwrite a value within that expression out-of-line. Other code typically avoids helpers, unless it can't (the `makeXY` methods) but then usually does extra work to guarantee that expressions remain safe.

DuncanUszkay1:
>and adding another temporary that will be set out-of-line can overwrite a value within that expression out-of-line

Is this issue limited to helpers which create temporary locals?

src/compiler.ts
DuncanUszkay1: Agreed that this is needed before merging- but for now I think I'll wait until I get get into a stable state (compiles, validates, optimizes successfully) before looking at this again unless you think that we need it sooner

dcodeIO:
The most prominent example of what can go wrong is in `makeCallInline`, which needs to prepare arguments from right to left while keeping track of which locals are already used before creating the temporaries for the inlined function's parameters. Technically affects everything where compilation order doesn't match execution order, since `Flow`s observe in compilation order.

DuncanUszkay1:
>Technically affects everything where compilation order doesn't match execution order, since Flows observe in compilation order.

So even a method like this could cause this issue?
```typescript
private getClosurePtr(closureExpr: ExpressionRef): ExpressionRef {
    var module = this.module;
    return module.binary(
      BinaryOp.ShlI32,
      closureExpr,
      module.i32(4)     
  )
}
```
Here execution order doesn't follow compilation order, but there are no locals added in this method. I'm not sure I understand how this would cause an issue with locals

DuncanUszkay1:
- [ ] Refactor injectClosedLocals so that it compiles in execution order

dcodeIO:
Wrapping an expression with something that doesn't modify state is usually fine. One example that's especially problematic are the changes to `makeRetain` in this PR. Previously, it would just wrap with something that doesn't modify state, but now it utilizes temporaries including calling other helpers. Again, not checked in detail, but my intuition is that this will blow up.

DuncanUszkay1:
I'll look into it- that might've been what was causing my local freeing problems in the first place. I'll see if I can find another option
- [ ] Refactor makeRetain to prevent compilation order locals problem
- [ ] Refactor makeRetain to prevent compilation order locals problem

DuncanUszkay1:
What if instead of using `getTempLocal` we used `getPointerCalculationLocal` or something to that effect- it means that we possibly add an extra local into the untouched in any function that frees/retains a function reference, but that doesn't seem like an enormous cost, and also seems like if it does something completely wrong (like having two locals of the same type which are used in completely distinct blocks instead of just one) then optimization passes could fix that.

DuncanUszkay1:
Isolated the pass which causes the failure: it's `"rse"` which I believe corresponds to `RedundantSetEleminiation` in binaryen. Will see if I can resolve this exception with this info

DuncanUszkay1:
Actually it appears I was wrong, now I can only seem to get the first three passes to work, failing on Flatten.

DuncanUszkay1:
In either case the issue seems to be the same, it's receiving a type which is greater than `_last_value_type`, which appears to just mean it's a garbage value for the type enum (greater than the maximum).

DuncanUszkay1:
Alright fixed it, it was just a garbage type being passed in that wasn't being emitted into the stack IR or text format (hence it working) but was being read by the optimization walkers and I guess the validator.

DuncanUszkay1:
Working through the rest of the test failures now, which are more straightfoward

tests/compiler/closure.ts
```diff
@@ -13,12 +13,52 @@ function testVar(): (value: i32) => i32 {
...
-  return function inner(value: i32) {
-    return $local0; // closure
-  };
+//function testLet(): (value: i32) => i32 {
+  //let $local0 = 0;
+  //return function inner(value: i32) {
+    //return $local0; // closure
+  //};
+//}
+//testLet();
```
DuncanUszkay1, line 22: This still doesn't work, not sure if it's something that should work or not. Once I know I'll either fix this or move this into a separate test to test against stderr

tests/compiler/closure.ts
```diff
@@ -13,12 +13,52 @@ function testVar(): (value: i32) => i32 {
...
+  var baz = 4
+  var f = (x1: i32): i32 => { return foo + bar - baz; }
+  var g = (x2: i32): i32 => { return (bar - baz) + foo; }
+  foo = 7
+  bar = 11
+  return g;
+}
+function runClosure(closureToRun: (x3: i32) => i32): i32 {
+  return closureToRun(1);
+}
```
DuncanUszkay1, line 64: The beginnings of a closure test- probably need more cases here

DuncanUszkay1:
Got most test compilation done (having trouble getting through the entire test case because it takes forever to run, especially given the degree to which this PR blows up the untouched. Turned on ASC_TRACE and fixed the memory leaks, not detecting any right now.

dcodeIO:
> having trouble getting through the entire test case because it takes forever to run

Locally, running tests in parallel and disabling the diff algorithm typically helps: `npm run test:compiler -- --parallel --noDiff`

DuncanUszkay1:
Tests updated now

dcodeIO:
Is s-expression syntax intended? While that might look like valid .wat, it actually isn't in some cases, since Binaryen has custom pseudo-instructions like push/pop/tuple.xy etc. not part of the spec.

DuncanUszkay1:
Oh no that was just a mistake, left it in from earlier debugging. fixing now

tests/compiler/std/typedarray.json
```diff
@@ -1,7 +1,6 @@
 {
   "asc_flags": [
     "--runtime half",
-    "--explicitStart",
-    "--use ASC_RTRACE=1"
+    "--explicitStart"
```
dcodeIO, line 4: `std/typedarray` is one of the more interesting tests, but it's missing RTRACE now

DuncanUszkay1: Oh, weird. Did not mean to do that.

MaxGraey:
Great. Now we need do something with `__global_closure`. It seems read and write to this variable prevent code reduction and father optimizations

DuncanUszkay1:
- [ ] Try to prevent _global_closure from blocking optimizations. Maybe push it into the If statement? I also wonder if binaryen is smart enough to tell that each value swap into global closure is followed by switching back, so it doesn't do anything if the code in between doesn't do anything.

dcodeIO:
**Test / Features** is most likely failing because these fixtures are not updated by default. Can be activated via `export ASC_FEATURES=*` or the equivalent to set an env var in your OS, and then updating fixtures as usual. Recommending to use node 14.

**Test / Boostrap** is most likely failing due to logical binary expression weirdness or other AS limitations when compiling the compiler using itself. For instance `if (someObj && someBool)` does not compile yet due to type mismatch, and must be written as `if (someObj !== null && someBool)` to have uniform types. Tracking issue is https://github.com/AssemblyScript/assemblyscript/issues/1136.

DuncanUszkay1:
I'm trying some signature manipulation stuff out- basically replicating the `this` injection pattern to pass in the global closure. Should remove all the global closure problems

DuncanUszkay1:
Actually since nothing is ever an instance method and a closure, I think I'm going to move forward by just reusing `this` entirely. Later on we can separate them if we need to but I can't think of a reason you'd ever have both off the top of my head 🤔

DuncanUszkay1:
With the latest push the globals are gone, and instead we're using the `this` parameter in all anonymous functions- even when they aren't closures. In the non-closure case we just pass in a null reference, which should get optimized away since it isn't being used.

Still failing to get significant gains on optimization though- next I'm going to be looking at fixing issues with arc optimization by moving the retain/releases out of if statements

DuncanUszkay1:
Just cleaning up some changes right now which would move retain/release statements outside of if statements so they can be optimized, and also fixing a bug with non-closures. Going to get these out EOD

MaxGraey:
Could you sync with upstream plz? Recent changes in compiler improved `$~argumentsLength` elimination which could help you better understand generated wasm code quality

tests/compiler/function-expression.optimized.wat
```diff
@@ -1,5 +1,99 @@
...
+  nop
+ )
+ (func $start:function-expression~anonymous|2~nonClosure (result i32)
+  i32.const 1
+ )
+ (func $start:function-expression~anonymous|3~nonClosure (param $0 i32) (param $1 i32) (result i32)
+  local.get $0
+  local.get $1
+  i32.add
+ )
```
DuncanUszkay1, line 28: Need to look into why this doesn't get optimized away anymore 🤔

src/compiler.ts
DuncanUszkay1: Updated process:
- Get either the closure pointer if the MSB set, or 0 otherwise.
- Release/Retain that result

src/compiler.ts
```diff
@@ -7233,16 +7346,71 @@ export class Compiler extends DiagnosticEmitter {
...
+              this.ifClosure(
+                module.local_get(exprLocal.index, type.toNativeType()),
+                this.getClosurePtr(module.local_get(exprLocal.index, type.toNativeType())),
+                module.i32(0)
+              )
+            ],
+            this.options.nativeSizeType
+          )
+        ),
+        module.local_get(exprLocal.index, type.toNativeType())
```
DuncanUszkay1, line 7374: From my understanding after talking to dcode: retains don't change the pointer, but it is expected that a retainment is an assignment of the form `x = retain(x)`. Before this change this wasn't a problem- because what was passed into retain was always exactly `x`. Now, what is passed into retain could be `x << 4` or `0` or `x`, so we just drop the retain result and do a get on the local.

Seems like that ought to be the same here, but I'll have to make sure that this isn't preventing ARC optimization

src/compiler.ts
```diff
@@ -7975,24 +8148,139 @@ export class Compiler extends DiagnosticEmitter {
...
-      this.currentType = instance.signature.type;
     }
 
+    // if this anonymous function turns out to be a non-closure, recompile a version
+    // of the function without context, and ignore the other index. Since the index is ignore
+    // the closure version of this function will never be used, and will be optimized out
+    if (instance.closedLocals.size == 0 && instance.prototype.isAnonymous) {
+      this.module.removeFunction(instance.internalName)
+      instance = new Function(
+        instance.prototype.name + "~nonClosure",
```
DuncanUszkay1, line 8166: Should be a way to use the name it already had, but was running into some issues. These names aren't exported, so it should be okay

src/compiler.ts
dcodeIO: The important aspect is that the optimize-arc pass sees that "`x` holds a reference" on `x = __retain(...)` (only), so it can eliminate the `__retain` when it sees a matching `__release(x)`. It completely ignores all other `__retain`s, and one must be very careful in these cases because the optimizer will still try to find another matching `x = __retain(...)`, potentially performing elimination of unbalanced branches or otherwise, leading to leaky or broken final code.

DuncanUszkay1: Not super adept with binaryen but would it be possible to change what we consider the `full retain pattern` to `local.set(x, block(drop(_retain(..)), local.get(x))` to match what we have here? Seems like we'd just need to change the binaryen pass right? Since functionally these two approaches are the same given that `_retain()` is a no-op, we just need to change what our binaryen pass it looking for?

Seems like it would just be some minor changes to https://github.com/WebAssembly/binaryen/blob/e4d1e203c8bf8577e645dbf7ab265d1ec1c68bee/src/passes/PostAssemblyScript.cpp#L98

If not, we might have to reconsider the current strategy of using a retain(0) in anonymous function cases, since we're unable to reconstruct the function reference for the writeback

src/program.ts
```diff
@@ -3398,7 +3474,29 @@ export class Function extends TypedElement {
...
+        return null;
+      }
+      if(parentResult.kind == ElementKind.LOCAL) {
+        let local = changetype<Local>(parentResult)
+        if(this.closedLocals.has(local.name)) return assert(this.closedLocals.get(local.name))
+        let mask = local.type.byteSize - 1;
+        let memoryOffset = this.nextGlobalClosureOffset;
+        if (memoryOffset & mask) memoryOffset = (memoryOffset | mask) + 1;
+        var closedLocal = local.close(memoryOffset);
+        this.nextGlobalClosureOffset += this.nextGlobalClosureOffset;
```
DuncanUszkay1, line 3494: Replicating the logic used to determine field offsets at proper byte alignment

src/program.ts
DuncanUszkay1: Added in some alignment logic that we use when assigning memory offsets for fields- is this what you meant?

src/compiler.ts
dcodeIO: Wondering if this can be `makeRetain(expr: ExpressionRef, isFunction: bool = false)`, so we end up with less code changes?

src/compiler.ts
dcodeIO: As a last resort, perhaps. I still hope that there a ways to make it work without looking for more complicated patterns, though. Something along the lines of the following maybe

```ts
var x = __retain(
  isClosure(ptr)
    ? ptr << 4
    : 0
);
...
__release(x);
```

```ts
var x;
if (isClosure(ptr)) {
  x = __retain(ptr << 4);
} else {
  x = __retain(0);
}
...
__release(x);
```

src/program.ts
```diff
@@ -3398,7 +3474,29 @@ export class Function extends TypedElement {
   lookup(name: string): Element | null {
     var locals = this.localsByName;
     if (locals.has(name)) return assert(locals.get(name));
-    return this.parent.lookup(name);
+    var parentResult = this.parent.lookup(name);
+    if(parentResult && this.parent.kind == ElementKind.FUNCTION) {
```
dcodeIO, line 3478:
```suggestion
    if (parentResult !== null && this.parent.kind == ElementKind.FUNCTION) {
```

src/program.ts
```diff
@@ -3398,7 +3474,29 @@ export class Function extends TypedElement {
...
+      var parentFunction = <Function>this.parent;
+      if(parentFunction.closedLocals.size > 0) { //TODO allow nested closure definitions
```
dcodeIO, line 3480:
```suggestion
      if (parentFunction.closedLocals.size > 0) { //TODO allow nested closure definitions
```

And perhaps cache `.closedLocals` in a local here since it is used again later.

src/program.ts
```diff
@@ -3398,7 +3474,29 @@ export class Function extends TypedElement {
...
+        this.program.error(
+          DiagnosticCode.Not_implemented,
+          this.identifierNode.range, this.identifierNode.text
+        );
+        return null;
+      }
+      if(parentResult.kind == ElementKind.LOCAL) {
```
dcodeIO, line 3487:
```suggestion
      if (parentResult.kind == ElementKind.LOCAL) {
```

src/program.ts
```diff
@@ -3398,7 +3474,29 @@ export class Function extends TypedElement {
...
+        let local = changetype<Local>(parentResult)
+        if(this.closedLocals.has(local.name)) return assert(this.closedLocals.get(local.name))
```
dcodeIO, line 3489:
```suggestion
        if (this.closedLocals.has(local.name)) return assert(this.closedLocals.get(local.name));
```

src/program.ts
```diff
@@ -3398,7 +3474,29 @@ export class Function extends TypedElement {
...
+        let mask = local.type.byteSize - 1;
+        let memoryOffset = this.nextGlobalClosureOffset;
+        if (memoryOffset & mask) memoryOffset = (memoryOffset | mask) + 1;
+        var closedLocal = local.close(memoryOffset);
+        this.nextGlobalClosureOffset += this.nextGlobalClosureOffset;
+        this.closedLocals.set(local.name, closedLocal)
```
dcodeIO, line 3495:
```suggestion
        this.closedLocals.set(local.name, closedLocal);
```

src/resolver.ts
```diff
@@ -2774,6 +2775,7 @@ export class Resolver extends DiagnosticEmitter {
     var signature = new Signature(this.program, parameterTypes, returnType, thisType);
     signature.parameterNames = parameterNames;
     signature.requiredParameters = requiredParameters;
+    if (prototype.isAnonymous) signature = signature.toClosureSignature()
```
dcodeIO, line 2778:
```suggestion
    if (prototype.isAnonymous) signature = signature.toClosureSignature();
```

src/types.ts
```diff
@@ -177,6 +184,10 @@ export class Type {
     return classReference !== null && classReference.hasDecorator(DecoratorFlags.UNMANAGED);
   }
 
+  get isFunction(): bool {
+    return this.signatureReference !== null && this.classReference == null
```
dcodeIO, line 188:
```suggestion
    return this.signatureReference !== null && !this.classReference;
```

src/types.ts
```diff
@@ -654,6 +662,20 @@ export class Signature {
...
+    // check `this` type
+    var thisThisType = this.thisType;
+    var otherThisType = other.thisType;
+    if (thisThisType !== null) {
+      if (otherThisType === null || !thisThisType.equals(otherThisType)) return false;
+    } else if (otherThisType) {
+      return false;
+    }
+
+    return this.externalEquals(other)
```
dcodeIO, line 676:
```suggestion
    return this.externalEquals(other);
```

src/types.ts
```diff
@@ -720,6 +742,20 @@ export class Signature {
     return sb.join("");
   }
 
+  toClosureSignature(): Signature {
+    var closureSignature = this.clone()
+    closureSignature.thisType = this.program.options.usizeType
+    return closureSignature
```
dcodeIO, line 748:
```suggestion
    var closureSignature = this.clone();
    closureSignature.thisType = this.program.options.usizeType;
    return closureSignature;
```

src/types.ts
```diff
@@ -720,6 +742,20 @@ export class Signature {
...
+  }
+
+  // Reverses toClosureSignature, for when we recompile a function with the context argument
+  // Not convinced this is the right way to go about getting the original unmodified signature, but it works
+  toAnonymousSignature(): Signature {
+    var normalSignature = this.clone()
```
dcodeIO, line 754:
```suggestion
    var normalSignature = this.clone();
```

src/types.ts
```diff
@@ -738,6 +774,7 @@ export class Signature {
       }
       clone.parameterNames = cloneParameterNames;
     }
+    clone.requiredParameters = this.requiredParameters
```
dcodeIO, line 777:
```suggestion
    clone.requiredParameters = this.requiredParameters;
```

src/program.ts
dcodeIO: Yes, looks good this way. That makes sure that the loads are fast (unaligned ones are rather unpredictable, often much slower).

src/program.ts
```diff
@@ -671,6 +671,48 @@ export class Program extends DiagnosticEmitter {
     );
   }
 
+  // TODO: make these more generic, purpose built for closures atm
+  makeNativeMember(
```
dcodeIO, line 675: Do these helpers exist to construct actual source-level classes backing closure contexts here? If so, that seems like unnecessary overhead since an actual class doesn't give us anything, like a user interacting with it or proper diagnostics. Suggesting to use a compiler-level simple class instead holding the necessary state, completely independent of sources.

src/program.ts
```diff
@@ -3149,6 +3192,28 @@ export class Local extends VariableLikeElement {
...
     this.setType(type);
   }
+
+  close(offset: usize): ClosedLocal {
+    return new ClosedLocal(this, offset);
+  }
+}
+
+/** a closed-over local variable. */
+export class ClosedLocal extends VariableLikeElement {
```
dcodeIO, line 3202: Having a separate class here makes me wonder if we lose the ability to record local state for closed vars in flows. Like, flows observe whether a small integer local needs to be wrapped on the next operation requiring it or can remain unwrapped to avoid unnecessary operations, or whether it is known-to-be retained and stuff like that. What happens there with `ClosedLocal`s?

src/compiler.ts
```diff
@@ -1933,6 +1937,7 @@ export class Compiler extends DiagnosticEmitter {
     var tableBase = this.options.tableBase;
     if (!tableBase) tableBase = 1; // leave first elem blank
     index = tableBase + functionTable.length;
+    assert(u32(index) < u32(2147483648));
```
dcodeIO, line 1940:
```suggestion
    assert(index >= 0);
```

src/compiler.ts
```diff
@@ -3541,6 +3546,24 @@ export class Compiler extends DiagnosticEmitter {
         }
         fromType = fromType.nonNullableType;
       }
+      var toSignature = toType.signatureReference
+      var fromSignature = fromType.signatureReference
+      if(toSignature && fromSignature && fromSignature.externalEquals(toSignature) && fromType.is(TypeFlags.CLOSURE)) {
```
dcodeIO, line 3551:
```suggestion
      if (toSignature !== null && fromSignature !== null && fromSignature.externalEquals(toSignature) && fromType.is(TypeFlags.CLOSURE)) {
```

src/compiler.ts
```diff
@@ -3541,6 +3546,24 @@ export class Compiler extends DiagnosticEmitter {
         }
         fromType = fromType.nonNullableType;
       }
+      var toSignature = toType.signatureReference
+      var fromSignature = fromType.signatureReference
```
dcodeIO, line 3550:
```suggestion
      var toSignature = toType.signatureReference;
      var fromSignature = fromType.signatureReference;
```

src/compiler.ts
```diff
@@ -6366,6 +6406,28 @@ export class Compiler extends DiagnosticEmitter {
...
+      let local = unchecked(_values[i]);
+      let closureClass= type.classReference!
+      exprs[i] = module.store(
+        local.type.byteSize,
+        module.local_get(closureContextLocal.index, NativeType.I32),
+        module.local_get(local.index, local.type.toNativeType()),
+        local.type.toNativeType(),
+        closureClass.offsetof(local.name)
+      )
+    }
```
dcodeIO, line 6427:
```suggestion
    var locals = assert(type.locals);
    let _values = Map_values(locals);
    var exprs = new Array<ExpressionRef>(_values.length);
    assert(type.is(TypeFlags.REFERENCE));
    var closureClass = assert(type.classReference);
    for (let i = 0, k = _values.length; i < k; ++i) {
      let local = unchecked(_values[i]);
      let nativeType = local.type.toNativeType();
      exprs[i] = module.store(
        local.type.byteSize,
        module.local_get(closureContextLocal.index, NativeType.I32),
        module.local_get(local.index, nativeType),
        nativeType,
        closureClass.offsetof(local.name)
      )
    }
```

The difference of using an `assert` vs a `!` is that the former triggers an assertion when compiling the compiler to JS as well, not only when compiling to Wasm. I typically use `!` if I am 100% certain that something will never be false-ish anyway, and an `assert` if I'd like to perform an actual check.

Added the REFERENCE assert because a type can have a `classReference` but is not actually a class, theoretically, for instance it is planned that `i32` will eventually have the `I32` wrapper as its `classReference`, but remains a basic type.

src/compiler.ts
```diff
@@ -6366,6 +6406,28 @@ export class Compiler extends DiagnosticEmitter {
...
+    return module.block(null, exprs)
```
dcodeIO, line 6428:
```suggestion
    return module.block(null, exprs);
```

src/compiler.ts
```diff
@@ -6466,7 +6528,28 @@ export class Compiler extends DiagnosticEmitter {
       // indirect call: index argument with signature (non-generic, can't be inlined)
       case ElementKind.LOCAL: {
         let local = <Local>target;
-        signature = local.type.signatureReference;
+        signature = local.type.signatureReference!;
```
dcodeIO, line 6531: Similar to my comment above

src/compiler.ts
```diff
@@ -6466,7 +6528,28 @@ export class Compiler extends DiagnosticEmitter {
...
+        if(local.type.is(TypeFlags.CLOSURE)) {
```
dcodeIO, line 6532:
```suggestion
        if (local.type.is(TypeFlags.CLOSURE)) {
```

src/compiler.ts
```diff
@@ -6466,7 +6528,28 @@ export class Compiler extends DiagnosticEmitter {
...
+          // If we're calling a local we know to be a closure, then we must still be in the creator functions
+          // scope. Because of this, we should update the values of locals that are still available
+          return module.block(null, [
+            this.injectClosedLocals(local),
+            this.compileCallIndirect(
+              assert(signature),
+              module.load(
+                local.type.byteSize,
+                local.type.is(TypeFlags.SIGNED),
+                module.local_get(local.index, this.options.nativeSizeType),
```
dcodeIO, line 6542: Can cache `this.options.nativeSizeType` here

src/compiler.ts
```diff
@@ -6466,7 +6528,28 @@ export class Compiler extends DiagnosticEmitter {
...
+                this.options.nativeSizeType,
+                0
+              ),
+              expression.arguments,
+              expression,
+              module.local_get(local.index, this.options.nativeSizeType),
+              contextualType == Type.void
+            )
+          ], signature.returnType.toNativeType())
```
dcodeIO, line 6551:
```suggestion
          ], signature.returnType.toNativeType());
```

src/compiler.ts
```diff
@@ -6241,6 +6278,7 @@ export class Compiler extends DiagnosticEmitter {
         this.makeReplace(
           valueExpr,
           module.global_get(global.internalName, nativeType),
+          type,
           alreadyRetained
         )
```
dcodeIO, line 6283: D'oh, I also missed a `;` here

src/compiler.ts
```diff
@@ -6571,14 +6654,45 @@ export class Compiler extends DiagnosticEmitter {
...
+        this.compileCallIndirect( //If this function isn't a closure
+          assert(signature), // FIXME: asc can't see this yet
+          module.local_get(tempFunctionReferenceLocal.index, NativeType.I32),
+          expression.arguments,
+          expression,
+          0,
+          contextualType == Type.void
+        )
+      )
+    ], constraints & Constraints.WILL_DROP ? contextualType.toNativeType() : returnType.toNativeType())
```
dcodeIO, line 6695: Can you explain the relation of the block type to `WILL_DROP` here? Typically, `WILL_DROP` is used in something like `i++` where the compiler can then compile without teeing the value, making the return type `Type.void`, and I would have expected a `void` around here as well.

src/compiler.ts
```diff
@@ -7233,16 +7347,71 @@ export class Compiler extends DiagnosticEmitter {
...
+    expr: ExpressionRef,
+    type: Type | null = null
+  ): ExpressionRef {
+    var module = this.module;
+
     var retainInstance = this.program.retainInstance;
     this.compileFunction(retainInstance);
-    return this.module.call(retainInstance.internalName, [ expr ], this.options.nativeSizeType);
+    if (type && type.isFunction) {
+      var exprLocal = this.currentFlow.getTempLocal(type)
```
dcodeIO, line 7359:
```suggestion
      var exprLocal = this.currentFlow.getTempLocal(type);
```

dcodeIO, line 7359:
```suggestion
    if (type !== null && type.isFunction) {
      var exprLocal = this.currentFlow.getTempLocal(type);
```

jtenner:
Hey. I see a lot of code style clashes in this post. They tend to take up a lot of time. Would it be possible to open a pull request and use `prettier` to help with code formatting issues?

src/compiler.ts
```diff
@@ -7233,16 +7347,71 @@ export class Compiler extends DiagnosticEmitter {
...
+            ],
+            this.options.nativeSizeType
+          )
+        ),
+        module.local_get(exprLocal.index, type.toNativeType())
+      ], type.toNativeType())
+
+      //this.currentFlow.freeTempLocal(exprLocal)
+
+      return functionRetainCall
```
dcodeIO, line 7380:
```suggestion
      ], type.toNativeType());

      //this.currentFlow.freeTempLocal(exprLocal);

      return functionRetainCall;
```

src/compiler.ts
```diff
@@ -7233,16 +7347,71 @@ export class Compiler extends DiagnosticEmitter {
...
+    }
+
+    return module.call(retainInstance.internalName, [ expr ], this.options.nativeSizeType)
```
dcodeIO, line 7383:
```suggestion
    return module.call(retainInstance.internalName, [ expr ], this.options.nativeSizeType);
```

src/compiler.ts
```diff
@@ -7233,16 +7347,71 @@ export class Compiler extends DiagnosticEmitter {
...
   /** Makes a release call, releasing the expression's value. Changes the current type to void.*/
-  makeRelease(expr: ExpressionRef): ExpressionRef {
+  makeRelease(expr: ExpressionRef, type: Type | null = null): ExpressionRef {
+    var module = this.module;
+
     var releaseInstance = this.program.releaseInstance;
     this.compileFunction(releaseInstance);
+
+    if (type && type.isFunction) {
+      var exprLocal = this.currentFlow.getTempLocal(type)
```
dcodeIO, line 7394:
```suggestion
    if (type !== null && type.isFunction) {
      var exprLocal = this.currentFlow.getTempLocal(type);
```

src/compiler.ts
```diff
@@ -7233,16 +7347,71 @@ export class Compiler extends DiagnosticEmitter {
...
+            )
+          ],
+          NativeType.None
+        )
+      ], NativeType.None)
+
+      // TODO: fix a bug in which this free causes some overwrites
+      //this.currentFlow.freeTempLocal(exprLocal)
+
+      return functionReleaseCall
```
dcodeIO, line 7413:
```suggestion
      ], NativeType.None);

      // TODO: fix a bug in which this free causes some overwrites
      //this.currentFlow.freeTempLocal(exprLocal);

      return functionReleaseCall;
```

dcodeIO:
@jtenner Seems like a good thing to do, yeah. We should probably combine this with the replacement of `tslint` with `eslint`, and add the necessary checks on top. I mentioned this as a necessary change a few times before, but haven't gotten around to it yet (secretly hoping that someone more familiar with eslint jumps in).

jtenner:
@dcodeIO please ignore code style problems, as bad looking as they are, until you are ready to pull a request involving prettier *first*. Then we can deal with eslint later.  Small pull requests, one at a time :).

tests/compiler/function-expression.optimized.wat
dcodeIO: Typically `call`ed somewhere directly, or becomes inserted into the table and there is at least one `call_indirect` expression somewhere that Binaryen cannot precompute. Exporting the table also prevents it, but I think that's not what's happening here.

src/compiler.ts
dcodeIO: That's probably because the old function with the same name is already in `Program#instancesByName`. A better fix would be to not reconstruct the `Function` there, but to modify the existing one if necessary.

src/program.ts
dcodeIO: This is looking fine

src/compiler.ts
```diff
@@ -7975,24 +8148,138 @@ export class Compiler extends DiagnosticEmitter {
...
+  private ifClosure(
+    indexExpr: ExpressionRef,
+    thenExpr: ExpressionRef,
+    elseExpr: ExpressionRef
+  ): ExpressionRef {
+    var module = this.module;
+
+    return module.if(
+      module.binary(
+        BinaryOp.EqI32,
```
dcodeIO, line 8248: `indexExpr` is of type `usize` here, so requires separate instructions for 32-bit and, eventually, 64-bit Wasm. Ofc this isn't a thing as of today, but other code already takes this into account so we don't have to rewrite half the compiler and go bug hunting for an extended period of time once 64-bit Wasm hits. Typically checked via `this.options.isWasm64`, sometimes also by comparing `nativeSizeType == NativeType.I64` where code uses that already.

src/compiler.ts
```diff
@@ -8206,6 +8493,20 @@ export class Compiler extends DiagnosticEmitter {
...
       }
+      case ElementKind.CLOSEDLOCAL: {
+        let closedLocal = <ClosedLocal>target;
+        let contextLocal = assert(flow.lookupLocal(CommonNames.this_));
+
+        // TODO: replace this with a class field access, once we are able to construct the class before
+        // compiling
+        return module.load(
+          closedLocal.type.byteSize,
+          true,
```
dcodeIO, line 8504: This appears to always perform a signed load potentially leading to wrong values when the type is unsigned, i.e. an `i32.load8_s` expands to a different value than `i32.load8_u` if msb is set.

dcodeIO:
I certainly hope this isn't too overwhelming already. Have been looking at lines so far mostly, not so much at concepts and logic. Regarding the style issues, the patterns i've seen so far are

* Non-boolean logical expression as `if` conditions. This isn't overly important anymore due to recent improvements on this front, but since all other code now does it explicitly, I guess we can also stick with the clear intents.
* Rarely caching loads into locals, which we typically do. For instance, instead of repeatingly accessing `this.xy` or `obj.yz`, the value is often pulled into a local in the hope that an optimizer or VM can do some register magic. It's always pulled into a local if the value can theoretically be `null`, avoiding rechecking later on. It is typically not pulled into a local if the value is used only once.
* Using `!` where using an `assert` or `if` instead would help to see the same error when compiling the compiler to JS instead of to Wasm. `!` is almost always the wrong choice in AS, as it indicates a missing branch, *except* if one is 100% certain that a value can never be `null` so we don't need an `assert` and there is no better syntax alternative (i.e. TS compatibility stuff).
* Missing whitespace before `if`, after `//` of comments etc.
* Missing semicolons

DuncanUszkay1:
Yeah I'll run prettier after accepting these changes, should prevent needing those comments going forward.

src/compiler.ts
DuncanUszkay1: Well we would still need to alter most of the makeRetain statements to include this new isFunction parameter, so it won't decrease the diff that way. Unless I'm missing something and it's impossible for these retain statements to be targeting a closure (and there are a lot of them, so I should probably look through again), in which case we can just omit the type on those ones, which would reduce the diff

tests/compiler/function-expression.optimized.wat
DuncanUszkay1: Ah right.. I suppose there's no way for binaryen to precompute `call_indirect` if it's calling a function pointer stored in linear memory

src/compiler.ts
DuncanUszkay1: That could work but might be tricky to implement, because now we need to keep track of `x` and `ptr` so we know what to release later. I'll look into it

src/program.ts
DuncanUszkay1: By compiler-level simple class do you mean just constructing a class manually? We certainly could, just means we don't get to reuse all the struct creation logic which determines offsets and such that exists within compileClass. But if we're worried about the overhead I will look into it

src/program.ts
DuncanUszkay1: Could probably just get rid of it and just make a property on local, then it shouldn't be a problem

src/compiler.ts
DuncanUszkay1: Why would that work? Index is a u32 so I'd expect it to always be `>= 0`

MaxGraey: I suggest leave this as is as more clear check or leave some comment about signed overflow

src/compiler.ts
DuncanUszkay1: When the result of the call was immediately discarded, then the contextualType would be void, so I take that as the native type. Could probably just replace `contextualType.toNativeType` just with `Type.void`. Otherwise I just used the call return type.

torch2424:
@DuncanUszkay1 Let me know what I can help with here! Perhaps I can write some tests? Or help with like style reviews? 

I did not oriignally know how large this PR was :upside_down_face: so it may take me a little while to catch up on how the closures actually work and things. :smile:

DuncanUszkay1:
@torch2424 Tests would be amazing- I want to get as many problems out into the open as possible. There's already some tests that have been added but I'm sure more edge cases can be found

What I'm going to be working on today:
- [ ] Fix everything mentioned in dcodes post
- [ ] Look into modifying functions instead of recompiling
- [ ] Ensure 64 bit compatibility
- [ ] Fix issue with signed loads
- [ ] Refactor ClosedLocal out of existence

torch2424:
@DuncanUszkay1 

> Tests would be amazing- I want to get as many problems out into the open as possible. There's already some tests that have been added but I'm sure more edge cases can be found.

Awesome! That's something easy I can definitely tackle that I am sure would help :smile: I'll start with that. I'll write some maybe EOD today, more likely tomorrow (as this is my when my PRs for work are blocked project haha!)

> One thing I could extract is adding the type argument to makeRetain, that's definitely gumming up the diff and could be safely merged to master much earlier because it would be a no-op

So I don't have enough context on this, but I do think keeping the runtime changes to make this possible in a seperate PR, and the actual code generation in a different PR, I think would make things A LOT easier to look through, and make everyone a lot more confident in their reviews and merging :smile: I'll let @dcodeIO make the call here though :+1:

MaxGraey:
Btw is it possible marking closure names instead non-closure function names? 
Like use for closures 
`<...name...>~closure~anonymous|0` 
instead `~nonClosure` suffix for non-closures:
`<...name...>~anonymous|0~nonClosure`

DuncanUszkay1:
>Btw is it possible marking closure names instead non-closure function names?

Yep, though as @dcodeIO pointed out, it might be better if I can just edit the function rather than recompiling and there won't be any need to change the name.

If that doesn't end up being the case then I'll change it so that closures have special names rather than non-closures

DuncanUszkay1:
One thing that would be greatly appreciated is if someone could try compiling this in 64 bit address mode and let me know if that works. Going to move onto other tasks for now

dcodeIO:
Attempting that at the current state would yield a broken Wasm binary unfortunately, because the respective instructions ultimately consuming the pointer do not recognize 64-bit inputs yet. It's also likely that there is other code unrelated to your changes not doing it right already, but it's still a good thing to at least try early so an eventual Wasm64 implementation will be easier.

src/compiler.ts
DuncanUszkay1: The challenge with that approach is that all local indexes become invalidated, as they are reduced by 1. Not sure if I'm able to make changes like that after compilation

DuncanUszkay1:
PR which extracts the additional argument added to makeRetain/makeRelease: https://github.com/AssemblyScript/assemblyscript/pull/1287/files

If we can get that into master it should make this diff easier to read, hopefully we can since the PR makes no functional changes

src/program.ts
DuncanUszkay1: No longer need these, class generation has been removed

tests/compiler/closure.ts
DuncanUszkay1: Moving it into limitations for now, since it doesn't seem necessary for an mvp

DuncanUszkay1:
The diff is slowly decreasing in size- if we can get https://github.com/AssemblyScript/assemblyscript/pull/1287 merged then the diff goes down to just 501 lines (excluding compiled test changes)

torch2424:
@DuncanUszkay1 As promised! Wrote a nice commons JS closures pattern test. But can you give me contributor access to your AssemblyScript fork? 

https://github.com/DuncanUszkay1/assemblyscript/tree/closures

In the meantime, heres the test I wrote. The comments explain everything weird I'm trying to do haha! :joy: 

```typescript
// Common use cases / concepts for closures, as covered in articles like:
// https://medium.com/@dis_is_patrick/practical-uses-for-closures-c65640ae7304
// https://stackoverflow.com/questions/2728278/what-is-a-practical-use-for-a-closure-in-javascript
// https://softwareengineering.stackexchange.com/questions/285941/why-would-a-program-use-a-closure
// https://medium.com/@prashantramnyc/what-is-an-iife-in-javascript-24baf0febf08

// Currently, IIFEs and simple Function Factories work
// But my advanced Function Factory Pub Sub, and weird function namespacing does not
// Due to runtime and/or compilation errors :)

// IIFE (Immediately Invoked function expressions)
// Used for encapsulating data usually

// Simple IIFE
let myData = ((): boolean => {
  return true;
})();

// Constructor IIFE?
// Don't know why someone wouldn't just use their class, but yeah

class IIFEReturn {
  myBool: boolean
  myFunc: (x: i32) => i32
}

let myInstanceThing = ((): IIFEReturn => {
  return {
    myBool: true,
    myFunc: (x: i32) => {
      return x + 1;
    }
  }
})();

assert(myInstanceThing.myBool == true);
assert(myInstanceThing.myFunc(24) == 25);

// Function Factories
// Closures that create specific functions

// Simple function that will change depending on input
type generatedFunc = () => i32;
let myFactory = (x: i32): generatedFunc => {
  let myFunc = (): i32 => {
    return 24 + x;
  }

  return myFunc;
}

let generatedPlusOne: generatedFunc = myFactory(1);
let generatedPlusTwo: generatedFunc = myFactory(2);

// For some reason, couldn't do
// Cannot invoke an expression whose type lacks a call signature. Type 'closure-common-js-patterns/myFactory' has no compatible call signatures.
// assert(myFactory(1)() == 25);
assert(generatedPlusOne() == 25);
assert(generatedPlusTwo() == 26);

// I often will use this for like Pub/Sub stuff

// Commenting all of this out, as it doesn't work yet due to runtime, or compilation errors :)

/*
type SubFunc = () => void;
type UnSubFunc = () => void;
let subscriptions = new Array<SubFunc>();
let globalSubVar: i32 = 0;

function subscribe(funcToCallOnPub: SubFunc): UnSubFunc {
  subscriptions.push(funcToCallOnPub);
  return (): void => {
    let funcIndex = subscriptions.indexOf(funcToCallOnPub);
    subscriptions.splice(funcIndex, 1);
  }
}

function publish(): void {
  for(let i = 0; i < subscriptions.length; i++) {
    // Can't call directly? Get a Type error
    // ERROR TS2757: Type '() => void' has no call signatures.
    // Noticed some other weird type errors if I don't declare the function type
    // But I also am meh at typescripte signatures haha!
    // subscriptions[i]();

    let subFunc = subscriptions[i];
    subFunc();
  }
}

let plusOne = (): void => {
  globalSubVar += 1;
}

let plusTwo = (): void => {
  globalSubVar += 1;
}


let unsubPlusOne: () => void = subscribe(plusOne);
let unsubPlusTwo: () => void = subscribe(plusTwo);

assert(globalSubVar == 0);
assert(subscriptions.length == 2);

publish();

assert(globalSubVar == 3);
assert(subscriptions.length == 2);

unsubPlusOne();

assert(globalSubVar == 3);
assert(subscriptions.length == 1);

publish();

assert(globalSubVar == 5);
assert(subscriptions.length == 1);

unsubPlusTwo();

assert(globalSubVar == 5);
assert(subscriptions.length == 0);

publish();

assert(globalSubVar == 5);
assert(subscriptions.length == 0);

// Namespacing private functions
// Again, kind of weird, they should probably just make a class, but it's another interesting test

class Chunk {
  totalSize: i32;
  usedSize: i32;
  write: (size: i32) => void;
}

let getChunk = (): Chunk => {
  let chunk: Chunk = {
    totalSize: 1024,
    usedSize: 0,
    write: (x: i32): void => {}
  };
  
  let growChunk = (): void => {
    chunk.totalSize += 1024;
  }

  let allocateForChunk = (amount: i32): void => {
    if (chunk.usedSize + amount <= chunk.totalSize) {
      chunk.usedSize += amount;
    } else {
      // growChunk(chunk);
      // allocateForChunk(chunk, amount);
    }
  }

  chunk.write = (x: i32) => {
    allocateForChunk(x);
  }

  return chunk;
  
}

let myChunk: Chunk = getChunk();

assert(myChunk.totalSize == 1024);
assert(myChunk.usedSize == 0);

myChunk.write(1025);

assert(myChunk.totalSize == 2048);
assert(myChunk.usedSize == 1025);
*/
```

DuncanUszkay1:
Appreciate the additional test cases! 

As for this issue:
```typescript
// Cannot invoke an expression whose type lacks a call signature. Type 'closure-common-js-patterns/myFactory' has no compatible call signatures.
// assert(myFactory(1)() == 25);
```
This appears to not be limited to closures, but a problem with calling functions in general which is present on master. [I've cut a ticket](https://github.com/AssemblyScript/assemblyscript/issues/1289).

DuncanUszkay1:
Isolated a problem from one of your test cases, made a minimal repro:
```typescript
class Foo {
  field: i32;
}

var compClosuresSameScope = (): void => {
  var foo: Foo = {
    field: 8
  }
  var withFoo1 = (): void => {
    foo.field += 1;
  }

  var withFoo2 = (): void => {
    withFoo1();
  }
}
```
I think the problem here is that the field being closed over (`foo`) is still in scope when calling `withFoo1` within `withFoo2`. It tries to read in the value of `foo` in from a local variable, but it's assuming we're still in `compClosuresSameScope` not `withFoo2`. Need to think a bit on how to resolve this

torch2424:
@DuncanUszkay1 Thanks for the invite to your repo! Let me know if you want me to open a PR with my test file :smile: 

> This appears to not be limited to closures, but a problem with calling functions in general which is present on master. I've cut a ticket.

Awesome! Thanks for opening that issue! Agree that it probably shouldn't be in this PR if it's a general function calling issue :smile: 

> I think the problem here is that the field being closed over (foo) is still in scope when calling withFoo1 within withFoo2. It tries to read in the value of foo in from a local variable, but it's assuming we're still in compClosuresSameScope not withFoo2, which has a separate set of locals. Need to think a bit on how to resolve this. Update: I've written out a fix for this locally, we just need to lookup the locals we're injecting/calling rather than assuming that they're present locally and load them from context as necessary. I think given that this is fixable, it might make sense to just block this with unimplemented for now and list it as a limitation we can fix later in interest of limiting the diff.

Thanks for isolating that! :smile: I appreciate you taking the time to do that. Agreed, anything that isn't a fundamental design thing like we discussed in the meeting, I think it would be cool to just open a ticket for, that way we can start playing with closures, and keep the diff small(er). Since it seems like you already know the fix, I'm sure we could do a patch after this lands as well easily :smile: 

---

So! Honestly, that was all the tests I could think of. I actually tend not to use closures much outside of the tests that are already written, and the patterns I showed in my test. Let me know what I can help with next, would be more than happy to keep helping here :smile:

DuncanUszkay1:
Added in a warning flag to indicate that closure support is experimental:
```
WARNING TS6190: Closure support is experimental.

     return ((): i32 => { return arg + foo + bar; } )();
             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   in closure.ts(41,11)
```

dcodeIO:
Looking at the bootstrap test, there appear to be these two new warnings:

```
WARNING AS201: Conversion from type 'usize' to 'u32' will require an explicit cast when switching between 32/64-bit.

         local.closureContextOffset
         ~~~~~~~~~~~~~~~~~~~~~~~~~~
 in src/compiler.ts(6457,9)

WARNING AS201: Conversion from type 'usize' to 'u32' will require an explicit cast when switching between 32/64-bit.

             localClosureContextOffset
             ~~~~~~~~~~~~~~~~~~~~~~~~~
 in src/compiler.ts(8480,13)
```

DuncanUszkay1:
Warnings should be resolved now

MaxGraey:
@DuncanUszkay1 We seem to have talked about this. What about replacing marking `~nonClosure` fuffixes to `~closure` for closure functions?

DuncanUszkay1:
I don't think we really came to a conclusion on the recompilation step, but I suppose I may as well throw that in for now

dcodeIO:
Can we close this PR in favor of the more recent https://github.com/AssemblyScript/assemblyscript/pull/1308 covering the official beta branch?

Issue #383, opened 2018-12-19
[Feature Request] Support for @destructor


jtenner:
I have a use case for creating objects in AssemblyScript that would benefit from a callback being called when they are garbage collected or destroyed. Of course javascript doesn't actually have destructors, and you can't tell when an object is being collected. However, if I could keep track of which objects were deleted, I could use a callback to communicate with my module consumer to notify them an internal object was collected.

This is just a simple example, utilizing the decorator syntax:

```ts
declare function notify_collected(ref: vec_3): void;

class vec_3 {
  x: f64;
  y: f64;
  z: f64;

  @destructor
  protected __some_func_name() {
    notify_collected(this);
  }
}
```

I don't know if this idea is good, or worth pursuing, but I just wanted to share it somewhere constructive to get feedback.  Thanks for all your hard work on this amazing project!

dcodeIO:
Should be doable, yeah, and there are certainly use cases where it makes sense to communicate destruction of a managed object to the host.

willemneal:
There seems to be more general need for a this in the new runtime.  For example, a class which holds references to unmanaged classes and uses the destructor method to manually free them.

jtenner:
Currently my use case is solved by using rtrace.

onionhammer:
I think this is still a great feature request, even if your specific use case is taken care of

jtenner:
I still have software that would benefit from a destructor. But I have ways of working around not having it.

jedisct1:
Old issue, but still relevant :)

My use case for destructors is WASI APIs, or more generally any external API that obtains opaque handles, that eventually have to be closed.

The current workaround is to use unmanaged objects and expect applications to explicitly destroy them, but this is not super convenient.

dcodeIO:
Created a little experiment linked above that might help to inform future design.

willemneal:
@jedisct1 It might also be nice to have something like `with` in python when using external APIs so that they will be cleaned up.

dcodeIO:
Closing this issue as part of [2020 vacuum](https://github.com/AssemblyScript/assemblyscript/issues/1277) because destructors turned out to be not feasible very likely, but I opened a [new issue](https://github.com/AssemblyScript/assemblyscript/issues/1299) tracking the road ahead.

PR #1256, opened 2020-05-07
Add GC finalization hook

dcodeIO:
Related: https://github.com/AssemblyScript/assemblyscript/issues/383

This is a quick experiment for a finalization hook when managed objects become collected. A lot can go wrong here, especially when finalization messes with MM/GC state before freeing, so this is as low-level and unsafe as it gets. Doesn't bother with registering destructors (needs compiler support or a virtual destructor table, and suffers from the same challenges) but might be a starting point for anyone interested in playing with finalization in general / inform future design. Wondering if a mechanism along the lines of JS weakref can be implemented on top.


tests/compiler/rt/finalize.ts
```diff
@@ -0,0 +1,55 @@
...
+    var map = this.map;
+    if (map.has(ptr)) {
+      let value = map.get(ptr);
+      this.fn(value);
+      map.delete(ptr);
+    }
+  }
+}
+
+var registry = new FinalizationRegistry(value => {
```
jtenner, line 23: So this is how we should hook into object finalization? Any tips or important things we should be aware of? Also, where should this go in the documentation?

dcodeIO: That's really just an initial idea along the lines of [JS weakrefs](https://github.com/tc39/proposal-weakrefs). The advantage this has over destructors is that the finalization function isn't called with the actual object being free'd, which would be super unsafe, but with the registered value. Nothing set in stone yet.

dcodeIO:
Keeping this PR open as part of [2020 vacuum](https://github.com/AssemblyScript/assemblyscript/issues/1277) as it hasn't yet received enough feedback to make a proper decision just yet. Also considering, in case it appears useful, to merge just the `__finalize` runtime hook as a low-level building block to encourage experimentation. It's opt-in anyway and doesn't affect code not using it.

dcodeIO:
Thinking about it a little more, since it doesn't affect existing code and is opt-in, let's just merge this. Shouldn't be a burden due to its simplicity, and enables everyone to experiment with finalization to aid future design of a proper API. Updated the test to check just `__finalize` without any design assumptions.

jtenner:
How do we use this officially? Is there anything in the docs about it?

dcodeIO:
An example of how to use it is [the test file](https://github.com/AssemblyScript/assemblyscript/blob/master/tests/compiler/rt/finalize.ts). Essentially, one has to add

```ts
@global function __finalize(ptr: usize): void {
  ...
}
```

as an additional runtime hook, which is then called with the pointer to the object being free'd right before it is. Not the final API of course (that'd be something more along the lines of `FinalizationRegistry`), just the integration point. Not sure if it deserves to be included in the docs just yet, but perhaps we can mention it [within RT internals](https://github.com/AssemblyScript/assemblyscript/tree/master/std/assembly/rt).

jtenner:
This is amazing. Thank you.

PR #1259, opened 2020-05-10
Refactor static properties / use correct this arguments

dcodeIO:
This PR fixes https://github.com/AssemblyScript/assemblyscript/issues/1252 by refactoring static property prototypes to have a singleton instance. Previously, the compiler would instead resolve the respective getter/setter, skipping over the static prototype, while it now behaves more like with functions. Note that this still does not apply to instance properties, since properties can never be generic anyway so these are resolved together with the class, which then doesn't contain property prototypes anymore.

While on it, I also made compilation of `this` arguments use the correct type, which is either the implicit or explicit `this` type of a function or getter/setter, or the class type for fields. Noticed that this was still TODO for properties, but then also implemented it for everything else. Has a couple implications on null checking, where some property accesses would not have complained previously, but now do (see the related compiler source changes, like `namespace.members ? namespace.members.x : y`, necessary to fix bootstrapping).

### DIFF
--- a/cli/asc.json
+++ b/cli/asc.json
@@ -243,7 +243,6 @@
     "category": "Binaryen",
     "description": "Skips validating the module using Binaryen.",
     "type": "b",
-    "alias": "c",
     "default": false
   },
 
--- a/src/compiler.ts
+++ b/src/compiler.ts
@@ -806,7 +806,7 @@ export class Compiler extends DiagnosticEmitter {
         break;
       }
       case ElementKind.PROPERTY_PROTOTYPE: {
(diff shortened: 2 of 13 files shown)


dcodeIO:
Thinking about this more, it might be even better to make properties behave exactly like functions, always using an intermediary prototype, since currently it's still somewhat inconsistent. For example, if a method is not used, it won't ever be resolved, hence essentially ignored, while properties will always be resolved along the class. I don't expect a large compile-time win from this, but extending tree-shaking to everything appears to be worth.

PR #1265, opened 2020-05-14
update options definition

jtenner:
While we are here, should we check the d.ts files are correct? Are there any other options for generating these dts files because they tend to get outdated, and are subject to developer error very often.


dcodeIO:
Thanks! The only option to guarantee types are in sync is transpiling from TS I guess, but I'd still favor to keep this internal stuff simple.

jtenner:
I guess my argument was that it currently isn't simple and tends to be error prone, but these are of course my opinions. If I find any more problems, I'll be sure to file more pull requests then! 😊

Thanks.

PR #1269, opened 2020-05-14
Fix env imports problem

jtenner:
I made the changes locally to my d.ts file and verified that the result is stable. Here's the output of the new Imports type.

image

Anything need to be changed or made clearer?

### DIFF
--- a/lib/loader/index.d.ts
+++ b/lib/loader/index.d.ts
@@ -5,9 +5,13 @@ export interface ResultObject {
   instance: WebAssembly.Instance;
 }
 
-/** WebAssembly imports with two levels of nesting. */
-export type Imports = {
-  [key: string]: object,
+/** Helper type for WebAssembly imports that defines two levels of nesting. */
+type ImportsHelper = {
+  [key: string]: any;
+}
+
+/** WebAssembly imports with an optional env object and two levels of nesting. */
+export type Imports = ImportsHelper & {
   env?: {
     memory?: WebAssembly.Memory;
     table?: WebAssembly.Table;
--- a/lib/loader/index.d.ts
+++ b/lib/loader/index.d.ts
@@ -5,13 +5,10 @@ export interface ResultObject {
   instance: WebAssembly.Instance;
 }
 
-/** Helper type for WebAssembly imports that defines two levels of nesting. */
-type ImportsHelper = {
-  [key: string]: any;
-}
 
 /** WebAssembly imports with an optional env object and two levels of nesting. */
-export type Imports = ImportsHelper & {
+export type Imports = {
+  [key: string]: any;
   env?: {
     memory?: WebAssembly.Memory;
     table?: WebAssembly.Table;


lib/loader/index.d.ts
```diff
@@ -5,9 +5,13 @@ export interface ResultObject {
   instance: WebAssembly.Instance;
 }
 
-/** WebAssembly imports with two levels of nesting. */
-export type Imports = {
-  [key: string]: object,
+/** Helper type for WebAssembly imports that defines two levels of nesting. */
+type ImportsHelper = {
+  [key: string]: any;
```
dcodeIO, line 10: How does this differ from just changing `object` to `any` below?

jtenner: It differes because when I want to use the typescript types, I get a tsc compiler error that the `env` object type is not assignable to `object`.

This makes the types more correct and explicit.

jtenner: Technically the type should be `WebAssembly.Memory | WebAssembly.Table | number | string | (...args: number[]) => void | number`

Is that something worth changing? Or could it just be `any`? As far as I can tell, you can use `anyref` imports, so it really could be `any`.

jtenner: Do you think it would be better to just make it `any`?

dcodeIO: My question was rather about how using `ImportsHelper` differs from changing the `[key: string]: object` index signature below to the equivalent without adding a helper class.

Regarding the type: Not sure what's best, but making things too restrictive leads to problems when specifying the `binaryen` object from binaryen.js as an import for example, as it does not fulfill the criteria.

jtenner: I switched to an inline `[key: string]: any` in the current pull request. Does this satisfy your requirements?

jtenner: And honestly, I'm not sure I understand the binaryen problem. All I can replicate locally is the typescript error that throws every time I want to use the loader in typescript.

dcodeIO: I think you are misunderstanding. Was just curious if there is a difference :)

dcodeIO: Regarding the Binaryen problem: I noticed this in `tests/bootstrap/index.ts`, which uses `{ binaryen }` as the imports object. This didn't type-check before I made the changes to `[key: string]: object`, but then lead to the errors your PR now tries to fix. All kinda unfortunate.

jtenner: There isn't a difference, not effectively anyway. I often like to separate out my types so that I can avoid operator precedence problems. In this case, the second commit I applied to this pull request is exactly equivalent, and accurately represents what an import can be: `any`.

dcodeIO:
Perhaps it makes sense to add a test, trying all sorts of valid imports. Maybe, once that's in place, an opportunity to concertize the types more pops up?

One of the tests would be an imports object of `{ binaryen: require("binaryen") }`, and another one what this PR is trying to fix, plus a few other variants.

jtenner:
Right. Let me see if this errors first.

jtenner:
image

It appears that this works just fine, because it's defined as `any`.

Am I missing something?

MaxGraey:
what about add `binaryen.d.ts` and revert back my old changes which exposed `Record` wrappers?

jtenner:
If anything, we can pull this request now, and file an issue to write a test later. I can get around to that later.

dcodeIO:
That's expected, yeah. It works with `: object` and `: any`, but will most likely not work with `WebAssembly.Memory | WebAssembly.Table | number | string | (...args: number[]) => void | number`, while other things obviously don't work with `: object` (hence this PR). What I'm essentially suggesting is trying to find the most useful variant that is not `: any` working with everything.

jtenner:
I don't think you can use anything other than `any` here. This is because we don't know what `anyref` objects will be passed in via the loader. For instance, it could be something created without the `Object` prototype via `Object.create(null)`, and you would have to somehow account for this.

If you really truly want to be pedantic about what can be passed to the loader, I can try to make a helper type.

```ts
// must work with anyref
type WebAssemblyImportCallback = (...args: any[]) => any;
type WebAssemblyImport =
  | number
  | string
  | Array<any>
  | WebAssembly.Table
  | WebAssembly.Memory
  | object;
```

Again, this is error prone, and being specific makes assumptions that might change in the future. `any` is the accurate type here.

jtenner:
I also want to add a reference for what the WebAssembly type looks like in lib.dom.ts

```ts
    var Table: {
        prototype: Table;
        new(descriptor: TableDescriptor): Table;
    };
    declare var Function: FunctionConstructor;
    var Memory: {
        prototype: Memory;
        new(descriptor: MemoryDescriptor): Memory;
    };
    var Global: {
        prototype: Global;
        new(descriptor: GlobalDescriptor, v?: any): Global;
    };
    type ExportValue = Function | Global | Memory | Table;
    type ImportValue = ExportValue | number;
    type ModuleImports = Record<string, ImportValue>;
    // this is what we are interested in, and what max had originally
    type Imports = Record<string, ModuleImports>;
```

FWIW I think these types are incorrect too, because we need to support `anyref` anyway.

PR #1273, opened 2020-05-14
Add provided set to cli parser

jtenner:
This will be useful in the following ways:

1. Aspect and other projects that bootstrap the compiler can reuse the cli parser and reduce code size
2. We can reuse this for when I get back around to asconfig
3. Maybe other uses

This is a very small change too. Should `provided` be a `Set<string>`, or `string[]`?

### DIFF
--- a/cli/util/options.js
+++ b/cli/util/options.js
@@ -21,6 +21,7 @@ function parse(argv, config) {
   var unknown = [];
   var arguments = [];
   var trailing = [];
+  var provided = new Set();
 
   // make an alias map and initialize defaults
   var aliases = {};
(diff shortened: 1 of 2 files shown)


cli/util/options.js
```diff
@@ -53,6 +54,7 @@ function parse(argv, config) {
       else { arguments.push(arg); continue; } // argument
     }
     if (option) {
+      provided.add(key);
```
dcodeIO, line 57: Isn't the contents of `provided` equivalent to `Object.keys(options)`?

jtenner: Oops. I put it in the wrong place, I think. Let me check.

jtenner: It should go into the boolean block, and in the nested if statement below. There.

PR #1276, opened 2020-05-15
Experiment: Memory/thread-safe runtime

dcodeIO:
Another experiment, this time to make the runtime memory and thread-safe. So far this should be memory-safe in that multiple modules can share the memory allocator and garbage collector, but there's no locking in place yet. As-is, stuff will explode.

This will most likely not be a viable approach in the long term due to all the locking overhead it ultimately introduces, but perhaps it can serve as the start of a filler implementation until there's something better. Regarding something better, I imagine a global allocator distributing pages to threads which then manage their memory with TLSF or otherwise. We also all know that manual locking is error-prone, and something more aligned with JS, like backed by workers and postMessage, would be much safer.


dcodeIO:
Keeping this PR open as part of [2020 vacuum](https://github.com/AssemblyScript/assemblyscript/issues/1277), as it is fairly fresh.

HerrCai0907:
clean outdated draft PR

PR #1280, opened 2020-05-16
Document null pointer behavior of the loader helpers

dcodeIO:
A while back it has been suggested to make the loader helpers check for zeroes implicitly, raising an error. While that seems good to do initially, it leads to a situation where users will check for zeroes manually to avoid throwing, and the helper will check for zero once again in case it has to throw. To avoid redundant checks like these, this PR instead documents the behavior more clearly, expecting users to check for zeroes where necessary. In the worst case there'll be one check in user code this way, while in the best case there's no check at all.

See: https://github.com/AssemblyScript/assemblyscript/pull/381

### DIFF
--- a/lib/loader/README.md
+++ b/lib/loader/README.md
@@ -43,7 +43,7 @@ Besides demangling classes exported from your entry file to a handy object struc
   ```
 
 * **__getString**(ptr: `number`): `string`<br />
-  Copies a string's value from the module's memory.
+  Copies a string's value from the module's memory. `ptr` must not be zero.
 
   ```ts
   var str = module.__getString(ptr);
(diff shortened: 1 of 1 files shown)


jtenner:
As-pect wraps the getString function, checking for 0 values, and returning a default value instead of the string when its 0.

Would it make sense to let getString return null instead?

dcodeIO:
This is similar in nature, in that the helper will always have to include a check for `0` even if it isn't necessary (type is `string`, not `string | null`) or has been checked in user code already. It's somewhat a dilemma between efficiency and convenience I guess, and I typically chose the former since the latter can be achieved by wrapping an efficient helper with a convenient function, while a convenient helper cannot be made efficient.

jtenner:
I like this reasoning. Thanks for the explaination.

PR #1283, opened 2020-05-18
Refactor compiler sources to adhere to TS strict mode

dcodeIO:
As the title says, this PR enables TypeScript's `--strict` mode (strict null checks, strict checking of function types, strict checking of property initialization in classes) in compiler sources by means of refactoring quite a bit of code. Also found a few potentially problematic places where `range`s were missing.

Helps https://github.com/AssemblyScript/assemblyscript/pull/1155 in particular in that the compiler sources don't need a heap of definitive assignment assertions anymore, where my expectation is that these will yield runtime checks ultimately, which we now avoid.

Technically this is a breaking change for internal compiler API users due to changed parameter orders and similar, but not for compiler users, so I'm not sure if we should make this a new major release just for that, or if it's ok to assume that internal API compatibility is something we can't guarantee between minor versions just yet?

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -50,6 +50,8 @@ jobs:
       run: npm ci --no-audit
     - name: Clean distribution files
       run: npm run clean
+    - name: Check sources
+      run: npm run check
     - name: Test sources
       run: |
         if [[ `node bin/asc --version` != *"-dev" ]]; then
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,22 +5,28 @@
   "requires": true,
   "dependencies": {
     "@babel/code-frame": {
-      "version": "7.5.5",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.5.5.tgz",
-      "integrity": "sha512-27d4lZoomVyo51VegxI20xZPuSHusqbQag/ztrBC7wegWoQ1nLREPVSKSW8byhTlzTKyNE4ifaTA6lCp7JjpFw==",
+      "version": "7.8.3",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.8.3.tgz",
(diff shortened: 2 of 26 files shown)


dcodeIO:
The next thing I wanted to look into was the same for stdlib, but the `Map` and `Set` constructors that reuse `this.clear()` for initialization leave me puzzled. Reusing this part of code makes total sense as it avoids duplication, but expressing this in an AOT-safe way [isn't possible](http://www.staging-typescript.org/play?#code/MYGwhgzhAEBiD29oG8BQ0PQA4Cd5YC5oIAXHASwDsBzAblXU2HktJwFdgT4cAKAShSNMmEgAtyEAHRVyJAfREBfBiNnzBaESPGSpufNAC80AESnFmFUqA), unless using a definitive assignment assertion, leaving us with:

* Accept the runtime check on every access of a DA property (quite a bit of overhead)
* Make it an unsafe feature with an implicit null-store instead (some overhead)
* Duplicate code (meh)
* `changetype` the hell out of it (unrealistic in user code)
* Diverge from TypeScript and traverse into functions within the constructor (¯\\(ツ)/¯, new set of problems)

dcodeIO:
One alternative I see is to allow uninitialized fields if their default value is a zero anyway, and do a `memory.fill` upfront to guarantee that this is correct. For `Map` this means that

```ts
export class Map<K,V> {
  private buckets: ArrayBuffer | null;
  private bucketsMask: u32;
  private entries: ArrayBuffer | null;
  private entriesCapacity: i32;
  private entriesOffset: i32;
  private entriesCount: i32;
```

is fine since every field can use a default value of zero. In particular this works for `Map` because `this.buckets` is accessed as `changetype<usize>(this.buckets)` exclusively anyway, and we know it is not `null`, so there is at least a straight-forward workaround in cases like these. Still very close to the 4th option above hmm.

dcodeIO:
Merging to get it out of the way, as it is likely that future work will conflict significantly.

Issue #1286, opened 2020-05-22
Add message to NotImplemented error


DuncanUszkay1:
Sometimes when we raise a `Not Implemented` error in the compiler it isn't clear just from the highlighted text what isn't implemented. If we add an optional clarification message, it could help. 

i.e. `Not Implemented: Closures`

dcodeIO:
Fixed by https://github.com/AssemblyScript/assemblyscript/pull/1295

PR #1294, opened 2020-05-26
Add argument to not implemented error message

DuncanUszkay1:
Issue: https://github.com/AssemblyScript/assemblyscript/issues/1286

Right now `Not Implemented` accepts no arguments, but it'd be nice if it also included a string which stated what exactly isn't implemented for cases when the range alone doesn't suffice.

Here I've just added the argument for the preexisting message, however I could also make an additional diagnostic message which has the argument so that we don't have to go fix every use of not implemented. The only issue with that is that we'd need a different diagnostic code, and based on the fact that they've been manually specified I'm assuming there's some special meaning behind the selected codes.

Without passing the argument: `ERROR AS100: Not implemented: {0}`
When passing the argument: `ERROR AS100: Not implemented: Try Statements`

So we'd likely need to go update all the current Not implemented error calls for this to go in.

Would love to have this for the closures PR so that I can make the Not implemented errors more clear

Thoughts?

### DIFF
--- a/src/builtins.ts
+++ b/src/builtins.ts
@@ -2838,7 +2838,7 @@ function builtin_assert(ctx: BuiltinContext): ExpressionRef {
     }
   }
   compiler.error(
-    DiagnosticCode.Not_implemented,
+    DiagnosticCode.Not_implemented_0,
     ctx.reportNode.typeArgumentsRange
   );
   return abort;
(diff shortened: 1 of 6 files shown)


dcodeIO:
> So we'd likely need to go update all the current Not implemented error calls for this to go in.

I can take a look at that.

DuncanUszkay1:
Should I leave this one to you then?

Issue #1282, opened 2020-05-18
AssemblyScript 0.10.0 : Imports that specify a file extension no longer work


torch2424:
Noticed this behavior after upgrading some more projects to `0.10.0`. I would use syntax like:

```typescript
import { exportedFunc } from "./export.ts";
import {add} from "my-other-package";

export function hello(): i32 {
  return add(exportedFunc(), exportedFunc);
}
```

However, imports with a file extension, (or external packages that import with a file extension), no longer can be found, and you will get errors like:

![Screenshot from 2020-05-18 14-48-37](https://user-images.githubusercontent.com/1448289/82263194-39b9d900-9917-11ea-8c30-affbb3230bae.png)
![Screenshot from 2020-05-18 14-47-55](https://user-images.githubusercontent.com/1448289/82263197-3a526f80-9917-11ea-8d87-d86086e4085a.png)

Made a quick little test case at: https://github.com/torch2424/as-playground/tree/master/import-file-extension

I feel this is probably due how we can now specify the file extension in the CLI itself :smile: 

And if you point me to where in the project this could probably be fixed, would be happy to open a PR for this. Thanks! :smile: :+1: 

cc @MaxGraey @dcodeIO

dcodeIO:
Appears TypeScript doesn't support this as well, and has a special error for it:

```ts
"An import path cannot end with a '{0}' extension. Consider importing '{1}' instead.": {
  "category": "Error",
  "code": 2691
},
```

What do you think of leaving it as an error to avoid ambiguity (i.e. there is an `export.ts` and an `export.ts.ts`), but improve the error message?

torch2424:
@dcodeIO Ah! If typescript doesn't support it, let's go ahead and just improve the error here :smile: Thank you!

PR #1307, opened 2020-05-28
Improve error message if an import cannot be found

dcodeIO:
This PR fixes https://github.com/AssemblyScript/assemblyscript/issues/1282 by adding the paths searched to the error message when an imported file cannot be found, for example when trying to import a file by erroneously specifying a file extension, looking like:

```
ERROR: Import 'tests/compiler/404.ts' not found.
  No such file 'path\to\tests\compiler\404.ts.ts'
  No such file 'path\to\tests\compiler\404.ts\index.ts'
  No such file 'path\to\tests\compiler\404.ts.d.ts'
```

Doesn't implement a `.ts` specific error message because that might be intended, but the problem should be clear from the paths searched. cc @torch2424

### DIFF
--- a/cli/asc.js
+++ b/cli/asc.js
@@ -207,20 +207,20 @@ exports.main = function main(argv, options, callback) {
   // Check for unknown arguments
   if (opts.unknown.length) {
     opts.unknown.forEach(arg => {
-      stderr.write(colorsUtil.stderr.yellow("WARN: ") + "Unknown option '" + arg + "'" + EOL);
+      stderr.write(colorsUtil.stderr.yellow("WARNING ") + "Unknown option '" + arg + "'" + EOL);
     });
   }
 
   // Check for trailing arguments
   if (opts.trailing.length) {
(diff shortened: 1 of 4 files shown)


dcodeIO:
Oh, right, good catch. Will fix :)

dcodeIO:
Now the error looks like

```
ERROR: Import path 'assembly/404.as' in 'assembly/index' not found.
  No such file 'path\to\assembly\404.as.as'
  No such file 'path\to\assembly\404.as\index.as'
  No such file 'path\to\assembly\404.as.d.as'
```

on

```ts
import "./404.as";
```

using `--extension .as`.

dcodeIO:
Last commit goes a little further by moving the error reporting for files not being found to the compiler, by accepting a `null` text in `parseFile`. Has the advantage that the respective code region can be highlighted in a diagnostic, which is something I always wanted to do. A typical error now looks like `ERROR TS6054: File '404.ts.ts' not found.`. Doesn't display search paths, but by highlighting the code location this should actually be even clearer.

Issue #1306, opened 2020-05-28
Investigate switching all of AS to ESM


dcodeIO:
While the CommonJS and ESM compatibility story in node appears to be a disaster still, ESM is getting more and more traction and since AS requires relatively modern environments anyway, we should aim at making the switch rather sooner than later.

Related issues: https://github.com/AssemblyScript/assemblyscript/issues/1207

michailVestnik:
At the moment, this is still a very relevant question!
Deno supports TypeScript out of the box and a simpler and more concise solution for working with Web Assembly!
Is it possible to have two versions of AS, one for Node.js, and another for Deno?
where can I keep up-to-date information on this issue?

vamsivishnu:
The vision is to be able to build/run Deno binaries from TypeScript and AssemblyScript source files, arbitrarily importing modules, supported natively in Deno. Really looking forward to seeing that vision realized.

vjpr:
`npx asinit .` generates a commonjs Node package.

What is the status of Node.js ESM support? I expect there would need to be an [`--experimental-loader=xxx`](https://nodejs.org/api/esm.html#esm_resolve_specifier_context_defaultresolve) created. Has this been done?

RefinedSoftwareLLC:
This might be related: 
[https://github.com/WebAssembly/binaryen/pull/4280](https://github.com/WebAssembly/binaryen/pull/4280)
[https://github.com/AssemblyScript/binaryen.js/releases/tag/v102.0.0-nightly.20211028-esm](https://github.com/AssemblyScript/binaryen.js/releases/tag/v102.0.0-nightly.20211028-esm)

PR #1313, opened 2020-06-03
Convert loader to ESM

dcodeIO:
This PR converts the loader to ESM format so it is more easily usable both on the web and under node.js. I stumbled over this while working on the new website and its editor, where I need a little hack to make the loader work. Of course that'd be easily fixable by appending an AMD/CommonJS shim, which I considered first, but with https://github.com/AssemblyScript/assemblyscript/issues/1306 being a thing now (see for context) I figured that I can as well initiate the switch to ESM at the loader.

Thoughts? :)


dcodeIO:
Well, seems TypeScript isn't happy with that in the bootstrap test. Also quickly looked up how to make a hybrid module, but that requires either transpiling, which is silly, or keeping the base code in CommonJS format, which isn't helpful in the browser.

dcodeIO:
[insert random swearwords / honest opinion here]

Looks like the only thing that'll actually work in all circumstances is to transpile, but pulling in potentially heavy dependencies like babel just for that feels wrong as well :(

jtenner:
Why not go the whole nine yards and convert the loader to typescript? :)

dcodeIO:
That's indeed another option given the circumstances. What I'm trying to achieve currently is that we have future proof sources in ESM, and would make the code base directly use the sources so making changes is immediately in sync without a build step, and contributors don't get confused by multiple files that look like sources. Then, if we absolutely need a build step, automate it on publish.

dcodeIO:
Latest commit now gives this a shot using [esm2umd](https://github.com/dcodeIO/esm2umd) as a prepublish step, making the loader default to ESM with UMD as a fallback. `ts-node` doesn't like this, unfortunately, perhaps because we are stuck on v6, so I converted the bootstrap test to ESM as well.

.eslintrc.js
```diff
@@ -87,13 +87,18 @@ module.exports = {
...
       },
       files: [
-        "**/*.js"
+        "**/*.js",
+        "**/*.mjs",
+        "**/*.cjs"
       ],
       rules: {
         // Node's support for ESM is still not great, but this rule is likely
         // to become activated once compatibility doesn't suck anymore.
```
jtenner, line 96: Maybe it would be a good idea to remove this language from our source?

dcodeIO:
Superseded by https://github.com/AssemblyScript/assemblyscript/pull/1513

PR #1314, opened 2020-06-05
Modernize READMEs

dcodeIO:
With the new website now being live, and the contributors and sponsors component in place, I figured that I can as well generate our own SVGs to display on the README. Fixes the OSC image being out of date, and is guaranteed to be in sync with what the website shows. Also simplified a few other READMEs.

[This is how it looks now.](https://github.com/AssemblyScript/assemblyscript/blob/readme/README.md)

Other READMEs now [look like this](https://github.com/AssemblyScript/assemblyscript/blob/readme/lib/loader/README.md).


dcodeIO:
Merging since no objections, but feel free to suggest further changes if there's something that can be improved :)

PR #1317, opened 2020-06-08
Fixed compiler constant hint in typescript typedefs

gzuidhof:
From what I can tell, it's 0 for WASM32, not JS. 

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -22,6 +22,7 @@ under the licensing terms detailed in LICENSE:
 * Duncan Uszkay <duncan.uszkay@shopify.com>
 * Surma <surma@surma.dev>
 * Julien Letellier <letellier.julien@gmail.com>
+* Guido Zuidhof <me@guido.io>
 
 Portions of this software are derived from third-party works licensed under
 the following terms:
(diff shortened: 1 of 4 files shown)


willemneal:
Perhaps it might be worth setting those values in the enum so that it's more declarative.

gzuidhof:
I agree that's a good idea, I will edit this PR.

dcodeIO:
What do you guys think about making the inverse change, i.e. actually make it JS=0, Wasm32=1 and Wasm64? This way, `!ASC_TARGET` would represent JS (not asc) and `ASC_TARGET` would be any valid asc target with the option to check which one it is.

gzuidhof:
I think that makes more sense, but it is definitely a breaking change for anybody depending on this value.

gzuidhof:
When not using the portability library, ASC_TARGET is the only value I have to set globally prior to running my program with a JS target. If the JS value is 0 instead I do not have to explicitly set it (I can use `!global.ASC_TARGET`)

dcodeIO:
Alright, I'm OK with this change as well. From my perspective there'll never be a better time to change the values than now so we break as few projects as possible, which will only become more over time. In particular the largest consumer of the portable lib is the compiler itself, and afaik some benchmarks use it as well but iirc not using `ASC_TARGET` for it.

Going to merge this PR as it fixes the current behavior, and suggesting to make the switch a new PR that we can pull on the next major release.

PR #1318, opened 2020-06-09
Make JS Target value 0 in enum and ASC_VERSION

gzuidhof:
As discussed in #1317


dcodeIO:
Thanks!

PR #1319, opened 2020-06-09
Rename anyref to externref to match proposal change

jayphelps:
Note: this is a breaking change, but only for those who have opted-in to the `reference-types` experimental feature. It and anyref usage are also undocumented AFAIK.

The reference type proposal recently had some changes, including the removal of all subtyping of reference types for now, as well as renaming anyref to externref so that a future type that actually supports subtyping (a real anyref) could be added back later--maybe.

I wanted to open this PR as a starting point for the work, but also because I believe some dialog needs to be had because this change impacts more than just renaming thing for AS and there is not yet a PR in binaryen to change those things. For example, now ref.null requires an [immediate](https://en.wikichip.org/wiki/immediate_value) to know whether it's a null for func or null for extern but this is not yet done in Binaryen.

I thought it might be important to discuss this ASAP since the proposal ship seems to be sailing to stage 4 soon.

- type checking assumes subtyping exists in several places [e.g. truthy](https://github.com/AssemblyScript/assemblyscript/blob/6adbbd9a62411d91d48ba27ffdf6fd63b0f02631/src/compiler.ts#L10436)
- [`makeZero()`](https://github.com/AssemblyScript/assemblyscript/blob/6adbbd9a62411d91d48ba27ffdf6fd63b0f02631/src/compiler.ts#L10334), for initialization assumes subtyping
- [tests](https://github.com/AssemblyScript/assemblyscript/blob/a27cc547c4929027527f01a405ef933348f36694/tests/compiler/features/reference-types.ts#L31-L44
) assume subtyping.

Proposal update: https://github.com/WebAssembly/reference-types/pull/87
Binaryen update: https://github.com/WebAssembly/binaryen/pull/2900
⯈
⯈
⯈


jayphelps:
@torch2424 not sure of the process to add things to the agenda. Can this get added to one of the upcoming meetings?

torch2424:
@jayphelps Oops! Just saw this, it slipped through my github notifications.

I saw you commented on the weekly meeting, that's totally how you add something to the agenda! Looking forward to discussing this then! :smile: :tada:

jayphelps:
Thanks!

dcodeIO:
As discussed at the last WG meeting, I decided to go ahead and raise some concerns here: https://github.com/WebAssembly/reference-types/issues/61#issuecomment-646470170

MaxGraey:
@jayphelps could we move forward and finish started changes in this PR? Currently we can't update to latest binaryen without this changes)

jayphelps:
The delay was because Binaryen had not yet adapted the rest of the changes to the proposal, which I believe impact AS too. We could update just the name, anyref to externref, but we would then be relying on part of the old version of r proposal and part of the new version intermixed, I think. Maybe that’s not a big deal, for now? After all, it’s pretty unclear if anyone else at all is using anyref in AS right now and if they are, I don’t think they’ve been vocal. Thoughts?

dcodeIO:
Given that we are blocked on this, and that reftypes is an experimental feature anyway, just renaming for now to reflect the state in Binaryen and looking into everything else later sounds reasonable to me.

jayphelps:
Okiedoke, I’ll move forward. Looks like we might have some test issues, looking into it.

dcodeIO:
Testing with https://nodejs.org/download/v8-canary/v15.0.0-v8-canary202006192e28363093/, the experimental wasm flags appear to be

```
  --experimental-wasm-eh (enable prototype exception handling opcodes for wasm)
        type: bool  default: false
  --experimental-wasm-simd (enable prototype SIMD opcodes for wasm)
        type: bool  default: false
  --experimental-wasm-return-call (enable prototype return call opcodes for wasm)
        type: bool  default: false
  --experimental-wasm-compilation-hints (enable prototype compilation hints section for wasm)
        type: bool  default: false
  --experimental-wasm-gc (enable prototype garbage collection for wasm)
        type: bool  default: false
  --experimental-wasm-typed-funcref (enable prototype typed function references for wasm)
        type: bool  default: false
  --experimental-wasm-reftypes (enable prototype reference type opcodes for wasm)
        type: bool  default: false
  --experimental-wasm-bigint (enable prototype JS BigInt support for wasm)
        type: bool  default: false
  --experimental-wasm-mv (enable prototype multi-value support for wasm)
        type: bool  default: false
  --experimental-wasm-threads (enable prototype thread opcodes for wasm)
        type: bool  default: false
  --experimental-wasm-type-reflection (enable prototype wasm type reflection in JS for wasm)
        type: bool  default: false
  --experimental-wasm-bulk-memory (enable prototype bulk memory opcodes for wasm)
        type: bool  default: true
```

The error seen in the CI log is

```
Error: unrecognized flag --experimental-wasm-externref
Try --help for options
- compile untouched
  SUCCESS (154.718 ms)

- compare fixture
  SUCCESS (0.939 ms)

- compile optimized
  SUCCESS (368.300 ms)

- instantiate untouched
  [call preInstantiate]
  ---
  CompileError: WebAssembly.Module(): invalid local type @+16
```

so I guess that'd have to be `--experimental-wasm-reftypes` in `tests/features.json` instead?

jayphelps:
How did you get it to change the node V8 version? I changed the test.yml file but it didn’t seem to change which version the test ran with, according to the logs.

dcodeIO:
Hmm, the change to test.yml looks correct to me. Perhaps CI got confused somehow and installed v14 from master's test.yml, or there is something wrong with that exact version (maybe try `v15.0.0-v8-canary202006192e28363093`). But also just guessing.

.github/workflows/test.yml
```diff
@@ -115,8 +115,7 @@ jobs:
     - uses: dcodeIO/setup-node-nvm@master
       with:
         node-mirror: https://nodejs.org/download/v8-canary/
-        # FIXME: newer node-v8 builds are currently broken
-        node-version: "14.0.0-v8-canary201911242015a12d82"
+        node-version: "v15.0.0-v8-canary202007077c53168ead"
```
MaxGraey, line 118: According to [this](https://github.com/pkgjs/nv/blob/master/index.js#L33) it seems version should declare as `15.0.0-v8-canary202007077c53168ead`

jayphelps: Lol thank you! I promise I specifically removed the v-before it when I copy and pasted, doubling checking that the v shouldn’t be included.... then must have some how either not saved or reverted it. 🤦‍♂️ Having another set of eyes wins again.

tests/features.json
```diff
@@ -22,7 +22,7 @@
       "--enable reference-types"
     ],
     "v8_flags": [
-      "--experimental-wasm-anyref"
+      "--experimental-wasm-externref"
```
MaxGraey, line 25:
```suggestion
      "--experimental-wasm-reftypes"
```

jayphelps:
Damn. I didn’t think about the fact that, while Binaryen and AS both just renamed it to externref, V8 actually implemented all the other changes around removing us typing/null stuff. That means as-is this PR can’t be merged.

Options I can think of:

- Remove anyref/externref support entirely for now
- Temporarily remove the ability to support null for variables of type externref
- Remove the test, meaning the PR will pass but if someone actually uses null for externref’s in V8 it will fail validation because it’s technically malformed.
- Someone updates Binaryen for the other ref types changes, then updates AS
- We wait on this PR and updating Binaryen for other reasons won’t be possible.

I’m inclined to say if updating Binaryen now is needed for other reasons, anyref support should just be removed for now. It was experimental, and this PR would have been a breaking change anyway.

Thoughts?

MaxGraey:
Personally I'm ok with but with extra `FIXME` comments:
> Remove the test code that relies on null, meaning the PR will pass but if someone actually uses null for externref’s in V8 it will fail validation because it’s technically malformed.

cc @dcodeIO

dcodeIO:
Commenting out the conflicting tests with a comment sounds good to me. Perhaps we should also comment out codegen related to `ref.null` and `ref.is_null`, and make these `Not implemented` error for now so there are at least no validation failure surprises included in emitted binaries until we know more?

src/compiler.ts
```diff
@@ -10763,7 +10763,8 @@ export class Compiler extends DiagnosticEmitter {
       case TypeKind.EXTERNREF: {
         // TODO: non-null object might still be considered falseish
         // i.e. a ref to Boolean(false), Number(0), String("") etc.
-        return module.unary(UnaryOp.EqzI32, module.ref_is_null(expr));
+        // TODO: return module.unary(UnaryOp.EqzI32, module.ref_is_null(expr));
+        assert(false, 'Truthy checks for externref are not yet supported');
```
dcodeIO, line 10767: Looks like these should be diagnostics, i.e. `DiagnosticCode.Not_implemented_0` with message `"ref.is_null"`. Unfortunately `makeIsTrueish` and others don't take the AST node as an argument yet (typically as `reportNode: Node`), which would have to be added in order to obtain the diagnostic range.

jayphelps:
Ok I changed all null ref stuff to be no longer implemented, and commented out the related tests. All green, ready for final review I believe.

src/compiler.ts
jayphelps: >Unfortunately makeIsTrueish and others don't take the AST node as an argument yet 
Yep that's why I punted on it. How should I proceed?

jayphelps: ohh I think you mean that I should pass in the appropriate node as another argument of `reportNode: Node` just for reporting purposes. That's hairy but makes sense. Will do. Lmk if that's wrong.

dcodeIO: Suggesting to add a `reportNode: Node` parameter to the affected functions, and provide it where these are called. The diagnostics would then look like:

```ts
this.error(
  DiagnosticCode.Not_implemented_0,
  reportNode.range,
  "ref.is_null" // respectively "ref.null", i.e. the exact feature missing
);
return module.unreachable();
```

replacing the `assert`s.

jayphelps: Done. please review.

src/compiler.ts
```diff
@@ -8271,7 +8276,12 @@ export class Compiler extends DiagnosticEmitter {
             this.currentType = signatureReference.type.asNullable();
             return module.i32(0);
           }
-          return module.ref_null();
+          // TODO: return null ref for externref or funcref
+          this.error(
+            DiagnosticCode.Not_implemented_0,
+            expression.range,
+            "null references"
+          );
```
dcodeIO, line 8284:
```suggestion
            "ref.null"
          );
          return module.unreachable();
```

src/compiler.ts
```diff
@@ -10752,10 +10770,15 @@ export class Compiler extends DiagnosticEmitter {
...
+      case TypeKind.EXTERNREF: {
         // TODO: non-null object might still be considered falseish
         // i.e. a ref to Boolean(false), Number(0), String("") etc.
-        return module.unary(UnaryOp.EqzI32, module.ref_is_null(expr));
+        // TODO: return module.unary(UnaryOp.EqzI32, module.ref_is_null(expr));
+        this.error(
+          DiagnosticCode.Not_implemented_0,
+          reportNode.range,
+          "ref.is_null"
+        );
```
dcodeIO, line 10781:
```suggestion
        );
        return module.unreachable();
```

src/compiler.ts
```diff
@@ -8800,7 +8811,7 @@ export class Compiler extends DiagnosticEmitter {
         }
         values[i] = expr;
       } else {
-        values[i] = this.makeZero(elementType);
+        values[i] = this.makeZero(elementType, expression);
```
dcodeIO, line 8814: Not directly related to this PR, but looks like we are missing a node represent an omitted expression so we can reference a better code location in diagnostics. I guess we should tackle that in a separate PR and merge it into this one so diagnostics aren't overly surprising.

dcodeIO: Changed my mind. Going to merge as-is and look into omitted nodes in a follow-up.

dcodeIO:
Thanks for looking into this! :)

PR #1321, opened 2020-06-09
Remember failed globals so we don't hit assertions

dcodeIO:
An alternative to fix https://github.com/AssemblyScript/assemblyscript/pull/1205 by remembering globals that failed to compile so we properly get a `false` upon subsequent `compileGlobal` invocations and don't run into `assert`s in the first place, as per my latest comment on the above PR.

@DuncanUszkay1: Can you confirm that this solves the issue as well?

If so, also fixes https://github.com/AssemblyScript/assemblyscript/issues/1165

Now also: fixes https://github.com/AssemblyScript/assemblyscript/issues/1186, fixes https://github.com/AssemblyScript/assemblyscript/pull/1190

### DIFF
--- a/src/common.ts
+++ b/src/common.ts
@@ -62,21 +62,23 @@ export enum CommonFlags {
   RESOLVED = 1 << 21,
   /** Is compiled. */
   COMPILED = 1 << 22,
+  /** Did error. */
+  ERRORED = 1 << 23,
   /** Has a constant value and is therefore inlined. */
-  INLINED = 1 << 23,
+  INLINED = 1 << 24,
   /** Is scoped. */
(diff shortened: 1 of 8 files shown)


dcodeIO:
Last commit is an attempt to also fix cases like

```ts
let a = (a = 4, 3)
```

on top of this, by reusing the set of pending globals in compilation to check for this condition. I'm not particularly proud of the dummy local that must be added for this, but it fixes the problem so I guess it's fine.

dcodeIO:
Merging to get it out of the way, but might not catch all cases yet. Let me know if there's something missing, can add it in a follow-up.

Issue #1367, opened 2020-06-30
Compile-time data segment import


lexaknyazev:
My application is compiled with `--runtime none` and uses only primitive numbers and intrinsics for calculations. The WebAssembly Memory object is used for exchanging data with JS host via low-level load/store ops on a fixed layout.

When an algorithm needs some pre-computed values, it's convenient to put them into some fixed memory segment on the first run like:

```ts
var firstRun = true;

export function foo(): void {
  if (firstRun) {
    // put pre-packed data in memory
    store<i64>(0, 0x0123456789ABCDEF);
    store<i64>(8, 0xFEDCBA9876543210);
    // ...
    firstRun = false;
  }

  // use 
  var v = load<u16>(byteOffset);
}
```

This approach works fine so far.

With the recently added `memory.data`, it's possible to avoid an extra branch with a bunch of store ops and to start an execution with the needed values already in place. However, there are two concerns for my use case. 

- `memory.data` returns a pointer that depends (in this case, as runtime isn't used) only on compiler's options (`--lowMemoryLimit` and `--memoryBase` specifically). There does not seem to be any way to enforce an offset of 0 (to replicate the snippet above) when using this function. Arguably, it's possible to partly workaround this limitation by setting the base memory to 8 and adding the same immediate offset to all relevant load ops. While the latter works, it looks fragile.
- When the size of such data is large, it's sub-optimal to keep it in the source code, so a new compiler option like `--importData file.bin` (with a configurable offset) would be useful here.

Thanks!

dcodeIO:
Importing data from a file sounds good to me. Suggesting to put the contents of the file at offset `--memoryBase`, so one can have both reserved memory plus additional static data from files afterwards, or omit one of the two. Would that be sufficient, or is there a need for multiple files or more granular control over offsets?

lexaknyazev:
As long as the spec supports only one memory object, single file is enough. For more complex cases, one could construct such a file from multiple sources using a build script.

> Suggesting to put the contents of the file at offset `--memoryBase`

Sounds fine. When `--memoryBase` is omitted (defaulting to 0), would it mean that data section imported from a file starts at 0?

dcodeIO:
> Sounds fine. When --memoryBase is omitted (defaulting to 0), would it mean that data section imported from a file starts at 0?

[...]

PR #1370, opened 2020-07-01
Add memoryData option to import a file as static memory

dcodeIO:
An implementation of the suggestion in https://github.com/AssemblyScript/assemblyscript/issues/1367, loading a file specified via `--memoryData` into static memory at memoryBase.

Open questions:
* Do we need constants like `ASC_MEMORY_DATA_OFFSET` and `ASC_MEMORY_DATA_LENGTH` to obtain the exact range or is expecting it at `ASC_MEMORY_BASE` with a known length enough?
* Currently does not attempt to align the loaded data, but places it right at memoryBase. If `--memoryBase` is omitted, it starts at 0, otherwise exactly at `--memoryBase` putting alignment under the user's control.
* Should we change related constants like `ASC_MEMORY_BASE` and `ASC_LOW_MEMORY_LIMIT` to `usize` (is `i32` currently)? Would be a breaking change, so might make sense to do this later.
* Doesn't play well with asconfig merging, in that exactly one memory segment can be imported somewhere, but not multiple from different configs. Given how low level this is I guess this is fine? A mechanism for multiple data files would be more complicated to use, since a user would have to keep track somehow. cc @jtenner 


torch2424:
So I've been doing this kind of behavior in Node scripts for a few projects of mine. Normally for multiple files though. 

 I was thinking of maybe doing a PR on the compiler for something like https://github.com/rollup/rollup-plugin-url or https://github.com/parcel-bundler/parcel/issues/1080#issuecomment-557240449 , where you can import a static file using an ES6 import as `memory.data` :smile:

Also, @tschneidereit mentioned to me that it may also be a good idea to take a look at:  [import assertions](https://github.com/tc39/proposal-import-assertions) ( and the [companion proposal for JSON modules](http://github.com/tc39/proposal-json-modules) ) . As we could then define our own behavior per type :smile: 

Though, I think this would maybe require some discussion. As I could imagine some assets would want special handlers (like JSON), and others may want to just be binary blobs (Like Images).

PR #1383, opened 2020-07-09
Fixed Incorrect Callback Return in CLI Module Validation

torch2424:
relates to https://github.com/DuncanUszkay1/assemblyscript/pull/14

Please see the PR above for the context. But TL;DR I was working on closures, and the tests would hang when failing when validating the module. I think this return is meant to happen outside of the measure. 

With this fix, tests will not hang whenever a module cannot be validated correctly :smile: :+1: 

cc @dcodeIO @MaxGraey

### DIFF
--- a/cli/asc.js
+++ b/cli/asc.js
@@ -653,12 +653,16 @@ exports.main = function main(argv, options, callback) {
   // Validate the module if requested
   if (!args.noValidate) {
     stats.validateCount++;
+    let callbackErrorResponse;
     stats.validateTime += measure(() => {
       if (!module.validate()) {
         module.dispose();
-        return callback(Error("Validate error"));
+        callbackErrorResponse = callback(Error("Validate error"));
       }
     });
+    if (callbackErrorResponse) {
+      return callbackErrorResponse;
+    }
   }
 
   // Set Binaryen-specific options
--- a/cli/asc.js
+++ b/cli/asc.js
@@ -653,15 +653,13 @@ exports.main = function main(argv, options, callback) {
   // Validate the module if requested
   if (!args.noValidate) {
     stats.validateCount++;
-    let callbackErrorResponse;
+    let isValid;
     stats.validateTime += measure(() => {
-      if (!module.validate()) {
-        module.dispose();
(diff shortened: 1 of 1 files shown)


cli/asc.js
```diff
@@ -653,12 +653,16 @@ exports.main = function main(argv, options, callback) {
   // Validate the module if requested
   if (!args.noValidate) {
     stats.validateCount++;
+    let callbackErrorResponse;
```
dcodeIO, line 656: Good catch! Perhaps let's make this an `isValid` var and throw the error `if (!isValid)`?

torch2424: That sounds good to me! :smile:

torch2424: Oh wait, if I'm not mistaken we expect the callback to return a status code? 

So it would be something like:

```
// Validate the module if requested
  if (!args.noValidate) {
    stats.validateCount++;
    let isValid;
    stats.validateTime += measure(() => {
      if (!module.validate()) {
        module.dispose();
        isValid = callback(Error("Validate error")) === 0;
      }
    });
    if (!isValid) {
      return 1;
    }
  }
```

dcodeIO: Something like the following maybe

```ts
if (!args.noValidate) {
  stats.validateCount++;
  let isValid;
  stats.validateTime += measure(() => {
    isValid = module.validate();
  });
  if (!isValid) {
    module.dispose();
    return callback(Error("validate error"));
  }
}
```

dcodeIO: Callback there is typically [the default](https://github.com/AssemblyScript/assemblyscript/blob/1d86695dcd27777271ff2ea25dbbc6b7892ce327/cli/asc.js#L220), unless one uses the CLI API. Returns `1` when an error is provided. The returned value becomes the exit code of asc.

torch2424:
@dcodeIO Made requested changes! And I am pretty sure they work, as I forced some compiler errors (which I think would break validation), but it didn't hang :smile: 

Screenshot from 2020-07-10 15-28-18

dcodeIO:
Great, thanks!

PR #1393, opened 2020-07-16
Closures: Added tests from the current closures PR

torch2424:
relates to #1308 
relates to https://github.com/AssemblyScript/working-group/issues/41

As discussed in the working group meeting, this moves the Tests that are being added / modified in the current AssemblyScript Closures PR #1308 , and skips the relevant errors in these closures tests. This is being done, as these tests represent our current goals of what closures should support, while also, allowing us to start bringing closures into main/master in smaller more manageable PRs :smile: 

**NOTE:** Closures will need additional tests later down the line. But again, just trying to start merging in what we currently have in parts that won't affect the project as a whole or break things :smile: 

cc @DuncanUszkay1 :smile: 

# Screenshots

Screenshot from 2020-07-15 17-41-40
Screenshot from 2020-07-15 17-39-36


tests/compiler/closure-passing-functions.ts
```diff
@@ -0,0 +1,12 @@
...
+}
+function apply_to_pair(x: i32, y: i32, fn: (x:i32, y:i32) => i32): i32 { 
+  return fn(x,y);
+}
+
+let addResult = add(1, 1);
+let applyResult = apply_to_pair(1, 2, add);
+
+assert(addResult == 2);
+assert(applyResult == 3);
```
DuncanUszkay1, line 12: I don't see the closure in this test 🤔  what is the purpose of this test?

torch2424: Ah! Good catch, yeah we added this test because of our current implementation in #1308 had a problem with this test. But, this was before we fixed the CI, so this test probably isn't needed. I'll remove it :smile:

torch2424:
@dcodeIO @DuncanUszkay1 This should be good to go now! Tests are passing and everything :smile: :+1:

dcodeIO:
Thanks Aaron! Suggesting to leave this PR open until we have an implementation to test positive against. Still good to have these tests pinned here so we know what we are looking for initially :)

torch2424:
@dcodeIO You are welcome! :smile: 

> Suggesting to leave this PR open until we have an implementation to test positive against. Still good to have these tests pinned here so we know what we are looking for initially :)

So normally I would agree, but since the PR is going to be really really large. I think it'd probably good to merge these, and then unskip / update them in our final closures PR? If you are worried that people will think that closures will work, I can add a note comment on the top of the file? :smile: 

I also think this method would be a better, that way we can implement closures themselves in smaller pieces and get tests passing one or a couple at a time :smile: 

Unless you would prefer to just merge it in an all-in-one PR?

HerrCai0907:
clean outdated draft PR

PR #1395, opened 2020-07-16
Update Binaryen and other dependencies

dcodeIO:
With https://github.com/AssemblyScript/assemblyscript/pull/1319 now sorted, we can update Binaryen to latest again. Also updates other dependencies.

Fixes https://github.com/AssemblyScript/assemblyscript/pull/1381, but should be mostly the same. As usual, doing this on my end due to intangible package-lock differences between OSes.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,27 +5,27 @@
   "requires": true,
   "dependencies": {
     "@babel/code-frame": {
-      "version": "7.10.1",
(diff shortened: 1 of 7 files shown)


dcodeIO:
Merging as tests are passing and the optimizer changes closely resemble those in https://github.com/AssemblyScript/assemblyscript/pull/1381. @MaxGraey If something seems off, let me know :)

PR #1404, opened 2020-07-21
Merge oldest parent first

willemneal:
This adds more tests and changes the order of merges so that cli > target > options > parent.target ...

Now adding `expected.json`, which will be checked if `--showConfig` is passed to the test runner.

Changed `addDefaults` to return a new options object instead of editing the passed one.

### DIFF
--- a/cli/asc.js
+++ b/cli/asc.js
@@ -280,11 +280,30 @@ exports.main = function main(argv, options, callback) {
   const seenAsconfig = new Set();
   seenAsconfig.add(path.join(baseDir, args.config));
 
-  while (asconfig) {
-    // merge target first, then merge options, then merge extended asconfigs
-    if (asconfig.targets && asconfig.targets[target]) {
-      args = optionsUtil.merge(exports.options, asconfig.targets[target], args);
+  // Keep array of configs
+  const configs = [asconfig];
+
+  // First find all parent configs and add them each to the front
+  while (asconfig.extends) {
+    asconfigDir = path.isAbsolute(asconfig.extends)
+      // absolute extension path means we know the exact directory and location
+      ? path.dirname(asconfig.extends)
+      // relative means we need to calculate a relative asconfigDir
+      : path.join(asconfigDir, path.dirname(asconfig.extends));
+    const fileName = path.basename(asconfig.extends);
+    const filePath = path.join(asconfigDir, fileName);
+    if (seenAsconfig.has(filePath)) {
+      asconfig = null;
+    } else {
+      seenAsconfig.add(filePath);
+      asconfig = getAsconfig(fileName, asconfigDir, readFile);
+      configs.unshift(asconfig);
     }
+  }
+  let entries = new Set();
+  asconfig = configs.shift();
+  while (asconfig) {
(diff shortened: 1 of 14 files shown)


cli/asc.js
```diff
@@ -298,43 +317,33 @@ exports.main = function main(argv, options, callback) {
           return p;
         });
       }
-      args = optionsUtil.merge(exports.options, args, asconfig.options);
+      args = optionsUtil.merge(exports.options, asconfig.options, args);
```
jtenner, line 320: This is the wrong order. The asconfig options need to be overridden by cli args.

The options.merge function puts the parent as the third parameter, preferring the second argument.

Did I miss something or mis-implement this algorithm?

willemneal: You're right.  Just added more tests and sure enough it ignores the CLI args.  The issue was that the defaults for the args was being set before the config properties were being added.  So if you treat `args` as the current config it will prioritize the defaults.  I just changed it to not add the defaults right away so we can get just the args passed and things seem to work.

dcodeIO:
The fixes to `util/options` look fine, but I'm wondering if most of the state keeping in `cli/asc` can be avoided. For instance, the intended order for the merge algorithm (preferring current over parent) was to start with parsed CLI args, and then subsequently merge the next set of options taking precedence over what's still to merge, i.e.

```
current = parse(...)
while (asConfig = getNextAsconfig()) {
  if (asConfig.target) {
    current = merge(current, asConfig.target)
  }
  current = merge(current, asConfig.options)
}
current = addDefaults(current)
// use current
```

Can you elaborate why this isn't working?

willemneal:
We need to split up the CLI args because the defaults in `args` are needed, e.g. `args.target`, `args.config`.  I took your advice and removed the configs array and returned to @jtenner's original design.

cli/asc.js
```diff
@@ -193,8 +193,9 @@ exports.main = function main(argv, options, callback) {
   if (!stdout) throw Error("'options.stdout' must be specified");
   if (!stderr) throw Error("'options.stderr' must be specified");
 
-  const opts = optionsUtil.parse(argv, exports.options);
-  let args = opts.options;
+  const opts = optionsUtil.parse(argv, exports.options, false);
+  let args = optionsUtil.addDefaults(exports.options, opts.options);
```
dcodeIO, line 197: This is happening too early, and is instead intended to be performed as the last step once CLI args and asconfigs have been fully processed. For that, we just need to apply the defaults for `target` and `config` early, independently of populating their values. Can either hardcode these or pull them from the config.

willemneal: I changed addDefaults so that it returns a new object.  This way `opts.options` is untouched.  Then at the end of the config it is merged back into args.  It was the only way I found that fixed the issues your having in #1406.

dcodeIO: Hmm, I suspect that the underlying issue is that the relative path transform is applied to imported file names, in turn stripping the `.` from `./a`, resulting in the packages test trying to find `a`, which resolves differently. While the changes in ordering here apparently prevent this, somehow, that doesn't look like a proper fix.

PR #1408, opened 2020-07-23
Add tests

willemneal:


tests/asconfig/target/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "private": true,
   "scripts": {
-    "test": "node ../index.js"
+    "test": "node ../index.js --target && node ../index.js --showConfig --runtime none"
```
dcodeIO, line 4: Is the `--target` without a target intended?

dcodeIO:
Thanks, looks good! Just one question above.

tests/asconfig/target/package.json
willemneal: Just fixed it!

cli/asc.js
```diff
@@ -303,7 +303,7 @@ exports.main = function main(argv, options, callback) {
     // Append entries
     if (asconfig.entries) {
       for (let entry of asconfig.entries) {
-        argv.push(optionsUtil.resolvePath(entry, asconfigDir));
+        argv.push(path.relative(baseDir, optionsUtil.resolvePath(entry, asconfigDir)));
```
dcodeIO, line 306: Why is this necessary when `asconfigDir` is already relative to `baseDir`?

dcodeIO:
Hmm, merging https://github.com/AssemblyScript/assemblyscript/pull/1411 into here looks like a mistake?

cli/asc.js
willemneal: `resolvePath` returns an absolute path.  This was mainly to output a relative path when `--showConfig` is passed.

dcodeIO: It returns an absolute path only if `entry` is already absolute, where I guess keeping it that way makes sense for debugging purposes (so one knows about where the entry comes from by looking at the asconfigs), or node resolution is used, which is debatable. In the latter case, the place to change it would be inside of `resolvePath` where `require.resolve` is called, though.

dcodeIO: Oh, wait, for entries node resolution isn't used anyway, so the `relative` here seems redundant.

willemneal:
I had forgotten to add the `node_modules` in the test for node resolution.

cli/asc.js
willemneal: It can be used to get the asconfigDir, which returns an absolute path.

dcodeIO: What do you mean?

willemneal: When the extends field the file could be found with node resolution, which returns an absolute path, which means `path.join(entry, asconfigDir)` would return an absolute path.  But I just really want to get this finished so if you don't like it then I can take it out.

dcodeIO: Thanks for clarifying. I'm not necessarily opposed to changing node resolution to relative, but if we do so I think we should do it in `resolvePath` where `require.resolve` is used (and the path returned as absolute), so we don't break explicitly specified absolute paths given as the value of `extends`. Like, if a user specifies something absolute explicitly, I'd prefer to not mangle it but keep it as-is.

dcodeIO: Changed it in #1406 accordingly.

willemneal: very true.

willemneal: Just looked at your changes to #1406, and I think you should roll it back and just keep everything as is.  The issue was that I wanted the path relative to `baseDir` in asc.js, but in resolvePath, `baseDir` is passed in as `asconfigDir.` So we could adapt that to always pass in top level `baseDir` or just forget about it all together.

dcodeIO: Ok, let's remove the `path.relative` here then.

dcodeIO:
As noted above, #1411, which is unrelated, somehow made it into this PR. With the original base branch now merged, I suggest to rebase this PR cleanly on master.

willemneal:
I think I'll just close this and reopen it to master.

Issue #1386, opened 2020-07-10
"No exported member" when Importing an exported "import * as"


torch2424:
I know the title is a bit confusing, but the best way to show off this bug is by just showing the pattern in code :smile: 

in `tests/compiler/reexport.ts` we `export {}` an `import * as`

https://github.com/torch2424/assemblyscript/blob/export-star-as/tests/compiler/reexport.ts#L28

Then, if we try to import that export in other AS code, it will erorr:

https://github.com/torch2424/assemblyscript/blob/export-star-as/tests/compiler/rereexport.ts#L1

For example:

![Screenshot from 2020-07-10 13-07-10](https://user-images.githubusercontent.com/1448289/87194900-4a9b2100-c2ae-11ea-937a-d12631a4c56f.png)

It is worth noting that if we `export {exportstar}` along with the other exports, the final wasm module will correctly have the right exports. Seems to just not be working between AS files :smile: 

cc @dcodeIO @MaxGraey :smile:

dcodeIO:
Hmm, in the example above (and the test), there seems to be a typo `exportstar` vs `exportstart`. Does it still fail when fixing that?

torch2424:
@dcodeIO Oops haha! :joy: 

But unfortunately no, that doesn't fix things :cry: 

![Screenshot from 2020-07-10 15-21-51](https://user-images.githubusercontent.com/1448289/87207612-25fc7480-c2c1-11ea-9b63-d4b2ca48346a.png)

torch2424:
I did some research on this, it seems like the `export *` doesn't have a `foreignPath`, therefore it can't find the exports from the original file? :thinking:

dcodeIO:
A `foreignPath` is only set for `export ... from "foreignPath"`, so on an `export { X }` without a `from` it's expected to be `null`. Most likely a problem with the algorithm itself, which is kinda unwieldy due to going through multiple stages. Always a lot of fun to dive back into it :)

torch2424:
Ahhh thanks for the insight :smile: If you don't mind, can you send me a link to where the algorithm is in the code? :smile:

dcodeIO:
Sure: The linking starts around [here](https://github.com/AssemblyScript/assemblyscript/blob/master/src/program.ts#L1004), after the queued collections have been populated during initialization of the original elements (in their respective file). Afterwards the graph of what re-exports what from where is known, but I guess there's something missing during traversal.

torch2424:
[...]

PR #1415, opened 2020-07-25
Keep resolving queued imports while there is progress

dcodeIO:
Potentially fixes https://github.com/AssemblyScript/assemblyscript/issues/1386. Looks like this is an ordering issue where imports depend upon other imports to be resolved first. In the specific case

```ts
// index.ts
import { one } from "./two"; // 1: is attempted first and...
one.one;
```

```ts
// two.ts
import * as one from "./one"; // 3: has not been processed yet, hence fails
export { one }; // 2: is found, but...
```

```ts
// one.ts
export var one = 1;
```

Instead of giving up early, the algorithm now repeats while it makes progress.

@torch2424 Can you confirm that this fixes the issue?


src/glue/js/node.d.ts
```diff
@@ -5,3 +5,6 @@
 
 declare const global: Record<string,unknown>;
 declare function require(name: string): unknown;
+declare namespace console {
+  function log(...args: unknown[]): void;
```
torch2424, line 9: Nice, thanks for adding this for console log debuggers like me :joy:

src/program.ts
```diff
@@ -1020,50 +1020,69 @@ export class Program extends DiagnosticEmitter {
...
+            queuedImports.splice(i, 1);
+            madeProgress = true;
+          } else {
+            ++i;
+            assert(false); // already reported by the parser not finding the file
+          }
         }
       }
-    }
+      if (!madeProgress) {
```
torch2424, line 1071: Ahhhhh! I had a feeling it may have been an ordering issue where one import had to resolve before others :smile: :+1:

tests/compiler/rereexport.ts
```diff
@@ -6,3 +6,6 @@ export {
   default,
   default as renamed_default
 } from "./reexport";
+
+import { exportstar } from "./reexport";
```
torch2424, line 10: Thanks for adding a test as well! :+1:

torch2424:
Yooooo thanks for merging / doing this @dcodeIO ! :smile: :+1:

PR #1419, opened 2020-07-29
Add tests for asconfig and windows

willemneal:
This tests checks if an entry file from another project imports from a relative path it is found on windows.

- [ x ] I've read the contributing guidelines

### DIFF
--- a/tests/asconfig/entry-points/assembly/data.ts
+++ b/tests/asconfig/entry-points/assembly/data.ts
@@ -1,0 +1,2 @@
+
+export let forty_one = 41;
--- a/tests/asconfig/entry-points/assembly/globalTwo.ts
+++ b/tests/asconfig/entry-points/assembly/globalTwo.ts
@@ -1,0 +1,5 @@
+// @ts-ignore
+@global
+  const ONE = 1;
+
+assert(ONE);
--- a/tests/asconfig/entry-points/assembly/globals.ts
+++ b/tests/asconfig/entry-points/assembly/globals.ts
@@ -1,3 +1,7 @@
+import { forty_one } from "./data";
+import "./globalTwo";
+
 // @ts-ignore: decorator
-@global const answerToLife = 42;
-assert(answerToLife);
+@global function answerToLife(): i32 { return forty_one + ONE; }
+
+assert(answerToLife());
--- a/tests/asconfig/entry-points/nested/assembly/index.ts
+++ b/tests/asconfig/entry-points/nested/assembly/index.ts
@@ -1,1 +1,1 @@
-assert(answerToLife == 42);
+assert(answerToLife() == 42);
--- a/tests/asconfig/entry-points/node-resolution/assembly/index.ts
+++ b/tests/asconfig/entry-points/node-resolution/assembly/index.ts
@@ -1,1 +1,1 @@
-assert(answerToLife == 42);
+assert(answerToLife() == 42);
--- a/tests/asconfig/entry-points/node-resolution/node_modules/entry-points/asconfig.json
+++ b/tests/asconfig/entry-points/node-resolution/node_modules/entry-points/asconfig.json
@@ -1,3 +1,3 @@
 {
-  "entries": ["assembly/globals.ts"]
+  "extends": "../../../asconfig.json"
 }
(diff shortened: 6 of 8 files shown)


willemneal:
@dcodeIO As I had suspected the issue is here:
```ts

> @ test D:\a\assemblyscript\assemblyscript\tests\asconfig\entry-points\node-resolution
> node ../../index.js

[2 parse error(s)]
ERROR TS6054: File 'globalTwo.ts' not found.

 import "./globalTwo";
        ~~~~~~~~~~~~~
 in ..\assembly\globals.ts(2,8)
```

dcodeIO:
Hmm, that's a strange error, especially because Windows doesn't really care about capitalization in file names. Would rather expect *nix to fail if it was that.

cli/asc.js
```diff
@@ -303,7 +303,7 @@ exports.main = function main(argv, options, callback) {
     // Append entries
     if (asconfig.entries) {
       for (let entry of asconfig.entries) {
-        argv.push(optionsUtil.resolvePath(entry, asconfigDir));
+        argv.push(path.relative(baseDir, optionsUtil.resolvePath(entry, asconfigDir)));
```
dcodeIO, line 306: Given `optionsUtil.resolvePath` https://github.com/AssemblyScript/assemblyscript/blob/348c42b7466054a793fde80a8100eaf9623fff36/cli/util/options.js#L241-L247

I wonder why this is necessary given that `asconfigDir` is *supposed* to be relative to `baseDir` (or absolute) already. If `entry` is relative, we should get `path.join(asconfigDir, entry)` as the result with `asconfigDir = path.join(baseDir, somePath)` (or absolute). As such I'm puzzled how this fixes the problem and would suspect that the problem might be deeper?

willemneal: I'm not sure.  Since I don't have windows, how it handles paths is weird to me.  In other node projects I've found that using all unix paths works well since node handles the conversion.

willemneal:
@dcodeIO could we make this an issue and just merge the patch for now?

dcodeIO:
I'd prefer to find out what's actually wrong there.

Tried this locally now, and what I can tell so far is that the problem happens in `test:asconfig/entry-points/nested`, with

* entry = `assembly/globals.ts`
* asconfigDir = `C:\path\to\assemblyscript\tests\asconfig\entry-points`
* baseDir = `.`

and the output of `optionsUtil.resolvePath(entry, asconfigDir)` being `C:\path\to\assemblyscript\tests\asconfig\entry-points\assembly\globals.ts`, ultimately resulting in `ERROR TS6054: File 'globalTwo.ts' not found.`.

Looks like this has something to do with absolute paths coming from the `require.resolve` branch in `resolvePath`, but might even be general for all absolute paths.

dcodeIO:
Diving a little deeper into it, the problem appears to be the call to `readFile` in `getAsconfig` reading a different file on Windows than on Linux, even though `file=asconfig.json` and `baseDir=.` are the same. On Windows, it loads `entry-points/node-resolution/asconfig.json`, while on Linux it loads `entry-points/nested/asconfig.json`. In particular, it appears that both commands are operating on a different `process.cwd()`, indicated by the directory used to look up `asconfig.json`.

willemneal:
How could `process.cwd()` be different?

dcodeIO:
That's the question. On Windows, it is `C:\path\to\assemblyscript\tests\asconfig\entry-points` while on Linux it is `/path/to/assemblyscript/tests/asconfig/entry-points/nested` when executing `test:nested`.

dcodeIO:
False alarm, I was executing a different test.

dcodeIO:
I see the problem now. At

https://github.com/AssemblyScript/assemblyscript/blob/348c42b7466054a793fde80a8100eaf9623fff36/cli/asc.js#L661

`sourcePath` is `..\assembly\globals.ts` on Windows, but internal

https://github.com/AssemblyScript/assemblyscript/blob/348c42b7466054a793fde80a8100eaf9623fff36/src/util/path.ts#L20

where the path ultimately ends at is supposed to be dealing with Unix-like paths, in turn normalizing differently. Adding a `path = path.replaceAll("\\", "/");` to `normalizePath` fixes the problem.

cli/asc.js
dcodeIO, line 306:
```suggestion
        argv.push(optionsUtil.resolvePath(entry, asconfigDir));
```

dcodeIO:
Thanks! :)

PR #1422, opened 2020-08-01
Imported & Exported ES6 Namespaces now contained properties that would have been queuedExports

### LINKED ISSUE #1417: Import * as not working if the file has an export from
relates to #1386 
relates to #1415 

So! My apologies for not noticing this in #1415 . I just made sure the code worked for hello world on my end, and not the more advanced tests :cry: 

So I started upgrading to 0.14.6 , and I found a new bug, which I showed in the branch: `export-star-as-with-export-from-bug`. 

I made the two changes to the AS tests:

* Added an export from: https://github.com/AssemblyScript/assemblyscript/blob/export-star-as-with-export-from-bug/tests/compiler/exports.ts#L60
* Added a new import * : https://github.com/AssemblyScript/assemblyscript/blob/export-star-as-with-export-from-bug/tests/compiler/reexport.ts#L31

And got the compiler error:

Screenshot from 2020-07-28 18-30-57

Essentially, all of the exports defined in the file work fine, but external exports still don't seem to be resolving.

cc @dcodeIO . I assume this would be more just keep iterating on the imports/exports resolution? :smile: I'll try at it tomorrow since it's the end of my work day haha!

torch2424:
closes #1417 
relates to #1386
relates to #1415

This completely fixes how `queuedExports` are added to imported namespaces `import * as ImportedNamespace`. This creates a map of all importedNamespaces, and then as queued exports are resolve,d they are assigned to the appropriate namespace. It also ensures globals are correctly overriden when we finally generate the code for the said namespace.

And in this process I found serveral more bugs in general in just how namespaces are both imported and exported. And this fixes alllll of that :smile: :tada: 

@dcodeIO for the review :smile: :+1: 

# Screenshots of this working

Screenshot from 2020-07-31 17-11-40
Screenshot from 2020-07-31 17-03-58
Screenshot from 2020-07-31 17-03-42


src/compiler.ts
```diff
@@ -815,6 +815,13 @@ export class Compiler extends DiagnosticEmitter {
         if (!classInstance.type.isUnmanaged) {
           let module = this.module;
           let internalName = classInstance.internalName;
+
+          // We remove the global before adding it, as this ensures
+          // we are not adding the same global twice. This is needed,
+          // As we can export a class from a file, and in that sample file,
+          // Export a namspace (import * as), that exports the same class.
+          // And there is no module.hasGlobal(), and this works fine :)
+          module.removeGlobal(internalName);
```
dcodeIO, line 824: Iiuc the problem here is that the same class can be exported using two different names, with the simplest case being

```ts
export class Foo {}
export { Foo as Bar }
```

so we get the exports

```
Foo = index/Foo
Bar = index/Foo
```

That correct?

Regarding removing and re-adding the global, this is a little dangerous because Binaryen APIs typically error when operating on something that doesn't exist. I can confirm that it doesn't error on `BinaryenRemoveGlobal`, but this might be an oversight with a slight chance to change in the future. As an alternative, I'd suggest to introduce a new `Compiler#exportedClasses: Set<Class>` and only add the global if it isn't present in the set yet.

src/program.ts
```diff
@@ -1062,6 +1067,19 @@ export class Program extends DiagnosticEmitter {
             );
             queuedImports.splice(i, 1);
             madeProgress = true;
+
+            // Also, make sure to record the namespace in our importNamespaces for later
+            if (importNamespacesMap.has(foreignFile)) {
+              let importNamespaceLocalFileMap = importNamespacesMap.get(foreignFile) as Map<File, Array<string>>;
+              let importNamespaceNames = importNamespaceLocalFileMap.get(localFile) as Array<string>;
```
dcodeIO, line 1074: This line can `abort`s if `localFile` is not yet in the map, needs another `has` check.

In general the mechanism here appears reasonable, but I'd suspect that it can be simplified a bit. For instance, we reference the imported namespace above via `foreignFile.asImportedNamespace` and when we have references there might be an opportunity to link related imported namespaces together so whenever we add to one of them we can add to all of them. Like `Namespace#aliasNamespaces: Set<Namespace>` and upon `Namespace#add` add to aliases as well? Just an idea so far, didn't dive deep :)

src/program.ts
```diff
@@ -1306,6 +1344,29 @@ export class Program extends DiagnosticEmitter {
     }
   }
 
+  private addExportToImportNameSpace(
+    importNamespacesMap: Map<File, Map<File, Array<string>>>,
```
dcodeIO, line 1348: Looks like we can avoid one lookup here by directly providing `Map<File, Array<string>>` as the argument, because earlier code already looked that up and knows that `file` is the key to the outer map iiuc. Depends a bit on the previous comment though, i.e. perhaps we can eliminate this helper with another approach to linking alias namespaces?

dcodeIO:
Thanks for investigating :) Added a few comments above, but I must admit that it's always a challenge to dive into this stuff because the linking is kinda abstract, so not sure how much sense the `aliasNamespace` suggestion makes. Perhaps you have a better idea how to make this more efficient without introducing new maps and somehow link sibling namespaces to make lookups cheap?

src/compiler.ts
torch2424: > Iiuc the problem here is that the same class can be exported using two different names

Ah not quite. The problem is if we have the code:

**one.ts**
```
export class MyClass {}
```

**two.ts**
```
export {MyClass} from "./one.ts"
import * as MyNamespace from "./one.ts"
export {MyNamespace}
```

When we traverse this part of the code. The first export of MyClass will add it as a global. But then we we export the namespace, this will go and traverse the name space members, which also contains MyClass, and thus will try to add it as a global. And then Binaryen will error out because it's already been created as a global, thus creating a conflict (when really it's the exact same thing).

> Regarding removing and re-adding the global, this is a little dangerous because Binaryen APIs typically error when operating on something that doesn't exist. I can confirm that it doesn't error on BinaryenRemoveGlobal, but this might be an oversight with a slight chance to change in the future. As an alternative, I'd suggest to introduce a new Compiler#exportedClasses: Set<Class> and only add the global if it isn't present in the set yet.

Oh yeah I'm totally down to not do this in my hack-y way and that totally makes sense haha! :joy: 

And just to be clear, I assume `Compiler#exportedClasses: Set<Class>` is referring to `src/compiler.ts`, https://github.com/AssemblyScript/assemblyscript/blob/master/src/compiler.ts#L327 and not something in Binaryen correct? :smile:

src/program.ts
torch2424: > This line can aborts if localFile is not yet in the map, needs another has check.

Ah yeah, I totally need to add the `has` check, was a bas assumption on my part, nice catch! :smile: 

> In general the mechanism here appears reasonable, but I'd suspect that it can be simplified a bit. For instance, we reference the imported namespace above via foreignFile.asImportedNamespace and when we have references there might be an opportunity to link related imported namespaces together so whenever we add to one of them we can add to all of them. Like Namespace#aliasNamespaces: Set<Namespace> and upon Namespace#add add to aliases as well? Just an idea so far, didn't dive deep :)

Ah, so to be honest, I'm a little lost in that explanation :upside_down_face: But I do agree it could be done better! I was already having a hard time understanding what was going on, so I was more trying to just at least not make a hack haha! :joy: But if I'm not mistaken, it seems like you want to change the map from:

foreignFile -> localFile -> NamespaceNames

to 

foreignFile -> localFile

And then the localFile would then introduce something to know "Hey, for this foreignfile, these namespaces of mine need to be updated"? :thinking:

src/program.ts
torch2424: Ah, yep, Totally we can just pass the smaller map, but yeah, let's continue on the other comment :smile:

torch2424:
@dcodeIO 

> Thanks for investigating :)

You are welcome! I am thoroughly surprised I was able to figure this out actually haha :joy: 

> Added a few comments above, but I must admit that it's always a challenge to dive into this stuff because the linking is kinda abstract, so not sure how much sense the aliasNamespace suggestion makes. Perhaps you have a better idea how to make this more efficient without introducing new maps and somehow link sibling namespaces to make lookups cheap?

So yes! I replied to your comments. Before I update the PR I'll wait to see what your replies are so I have a better idea on what to try to implement. 

For the `hasGlobal` fix, If my comment was correct in that it's a set on our compiler class, than we should be good to go there.

For the Namespace set, yeah I just wanted you to either say I was understanding what you meant, and/or if my whole add functionality to the file class to allow adding exports later to a namespace makes sense.

Either way! Thanks for the review, super excited to get this in :smile:

src/compiler.ts
dcodeIO: Right, just a set on the `Compiler` class to keep track of the classes already handled, so the implicit id global is only added if it doesn't yet exist :). Also wondering if anyone ever used the implicit id, as it's really just a convenience shortcut so one doesn't have to

```ts
export const MyClassID = idof<MyClass>;
```

when already

```ts
export { MyClass };
```

so externally one can do `exports.MyClass` to obtain the id instead of the manual `exports.MyClassID`.

src/program.ts
dcodeIO: Yeah, that's all a bit confusing. What I'm suggesting is that when we convert a file to a namespace in `File#asImportedNamespace`, we can keep track of all the namespaces we made from that same file, and implicity add new elements to all the converted namespaces when adding an element to the file in `File#add`. Let's say we have a `Set` `File#asImportedNamespaces` containing all the namespaces we made from that file, then instead of adding everything manually, adding any element to that `File` would also populate all the `asImportedNamespaces` automatically, in turn greatly simplifying this.

src/compiler.ts
torch2424: > Also wondering if anyone ever used the implicit id, as it's really just a convenience shortcut so one doesn't have to

I'm not too sure as I've really just been focused on the importing/exporting between files here :joy: But it does seem like a nice-to-have/nice-to-try later on :smile:

src/program.ts
torch2424: Ahhhhhh! That makes perfect sense, cool! I'll do that now! :smile:

torch2424:
@dcodeIO Made Requested changes, and tests are passing on my end! This is good to go! :smile_cat: 

Screenshot from 2020-08-04 11-57-53

src/compiler.ts
```diff
@@ -367,6 +367,8 @@ export class Compiler extends DiagnosticEmitter {
   virtualCalls: Set<Function> = new Set();
   /** Elements currently undergoing compilation. */
   pendingElements: Set<Element> = new Set();
+  /** Classes that are already added as globals */
+  globalClasses: Set<Class> = new Set();
```
dcodeIO, line 371: Comment and name here are somewhat ambiguous. Can easily be misinterpreted as if this has something to do with the global scope or elements `@global` to the program. What we are tracking here in particular is "classes, that are module exports, already processed". Perhaps one way to model this in a more general way is to make the map contain all the module exports already processed, including functions and globals, and name the set `doneModuleExports` of type `Element`? While we really only check classes currently, we'd still get a somewhat complete concept out of it this way :)

src/compiler.ts
```diff
@@ -812,11 +814,18 @@ export class Compiler extends DiagnosticEmitter {
       case ElementKind.CLASS: {
         let classInstance = <Class>element;
         // make the class name itself represent its runtime id
-        if (!classInstance.type.isUnmanaged) {
+        if (!classInstance.type.isUnmanaged && !this.globalClasses.has(classInstance)) {
```
dcodeIO, line 817: There can be multiple exports with different names of the same class, sharing a single global. As such, this seems not quite correct, in that it also skips `addGlobalExport` but should only skip the `addGlobal` if the class has already been processed.

src/program.ts
```diff
@@ -2893,6 +2893,11 @@ export class File extends Element {
   exportsStar: File[] | null = null;
   /** Top-level start function of this file. */
   startFunction!: Function;
+  /** 
+   * Array of all the names space made from this file 
+   * E.g import * as MyNamespace from "./this-file"
+   */
```
dcodeIO, line 2899:
```suggestion
  /** Array of `import * as X` alias namespaces of this file. */
```

src/program.ts
```diff
@@ -2962,6 +2967,14 @@ export class File extends Element {
...
     exports.set(name, element);
     if (this.source.sourceKind == SourceKind.LIBRARY_ENTRY) this.program.ensureGlobal(name, element);
+
+    // Also, add to the namespaces that caputure our exports
+    for(let i = 0; i < this.importedNamespacesFromThisFile.length; i++) {
+      let ns = this.importedNamespacesFromThisFile[i];
+      if (!ns.lookup(name)) {
+        ns.add(name, element);
+      }
+    }
```
dcodeIO, line 2977:
```suggestion
    // Also add to all the `import * as X` alias namespaces
    let aliasNamespaces = this.importedNamespacesFromThisFile;
    for(let i = 0, k = aliasNamespaces .length; i < k; i++) {
      aliasNamespaces[i].add(name, element);
    }
```

dcodeIO: Just came up with the `aliasNamespaces` term while looking at this section, perhaps that name is a little more convenient than `importedNamespacesFromThisFile` in general? :)

src/program.ts
```diff
@@ -2987,12 +3000,21 @@ export class File extends Element {
   }
 
   /** Creates an imported namespace from this file. */
-  asImportedNamespace(name: string, parent: Element, localIdentifier: IdentifierExpression): Namespace {
+  asImportedNamespace(
```
dcodeIO, line 3003: If you think that `aliasNamespaces` is good, we may also rename this one to `asAliasNamespace`.

src/compiler.ts
```diff
@@ -812,11 +814,18 @@ export class Compiler extends DiagnosticEmitter {
...
           let module = this.module;
           let internalName = classInstance.internalName;
+
+          // We remove the global before adding it, as this ensures
+          // we are not adding the same global twice. This is needed,
+          // As we can export a class from a file, and in that sample file,
+          // Export a namspace (import * as), that exports the same class.
+          // And there is no module.hasGlobal(), and this works fine :)
```
dcodeIO, line 825: Comment is outdated now

src/program.ts
```diff
@@ -2987,12 +3000,21 @@ export class File extends Element {
...
     var declaration = this.program.makeNativeNamespaceDeclaration(name);
     declaration.name = localIdentifier;
     var ns = new Namespace(name, parent, declaration);
     ns.set(CommonFlags.SCOPED);
     this.copyExportsToNamespace(ns);
+    // NOTE: Some exports are still queued, and are not added here.
+    // But will be added to the namespace when the exports are being resolved.
+    // We are adding this namespaces to our this files imported namespaces so
+    // we can resolve when we add our exports
+    this.importedNamespacesFromThisFile.push(ns);
```
dcodeIO, line 3017:
```suggestion
    // NOTE: Some exports are still queued, and can't yet be added here,
    // so we remember all the alias namespaces and add to them as well
    // when adding an element to the file.
    this.importedNamespacesFromThisFile.push(ns);
```

Indeed, this appears to be the most efficient way to handle it, thanks for updating! 👍

dcodeIO:
Thanks! 👍 A few more comments above (sorry). I like how the mechanism works now!

torch2424:
Made requested changes @dcodeIO ! :smile: :+1: 

Screenshot from 2020-08-04 18-28-47

src/program.ts
```diff
@@ -2962,6 +2964,14 @@ export class File extends Element {
     if (!exports) this.exports = exports = new Map();
     exports.set(name, element);
     if (this.source.sourceKind == SourceKind.LIBRARY_ENTRY) this.program.ensureGlobal(name, element);
+
+    // Also, add to the namespaces that caputure our exports
```
dcodeIO, line 2968:
```suggestion
    // Also, add to the namespaces that capture our exports
```

src/program.ts
```diff
@@ -2962,6 +2964,14 @@ export class File extends Element {
...
+    for(let i = 0; i < this.aliasNamespaces.length; i++) {
+      let ns = this.aliasNamespaces[i];
+      if (!ns.lookup(name)) {
+        ns.add(name, element);
+      }
```
dcodeIO, line 2973: Did you find that the additional check is necessary here? I guess it shouldn't, because `File#add` would already terminate if there is a duplicate, and nothing else should be adding to the alias namespaces? Or is there something?

torch2424: Oh no I just put it there because I thought I needed it :joy: I'll remove it :smile:

torch2424:
Made requested changes! This is good to go @dcodeIO :smile: :+1: 

Screenshot from 2020-08-05 09-22-12

dcodeIO:
Looks great, thank you!

torch2424:
Yooooooooo! :smile: :+1: Thanks for the review @dcodeIO :rocket:

PR #1448, opened 2020-08-27
Add CLI option: allow specifying --tsdModuleName for .d.ts

WietseWind:
[X] I've read the contributing guidelines

Currently, if the `-d` argument is added to the `asc` CLI command specifying a location to store a generated `.d.ts` (Typescript definition for the AssemblyScript being built) looks like:

```
declare module ASModule {
// ...
}
export default ASModule;
```

The module name `ASModule` is hardcoded:
https://github.com/AssemblyScript/assemblyscript/blob/master/src/definitions.ts#L363

It would be nice if the module name could be specified by eg. a CLI argument, which is why I added:
`--tsdModuleName`

When specified and matching `/^[a-z0-9_]$/i` instead of `ASModule` the specified value will be used in the generated .d.ts file:

```
declare module MyCustomModule {
// ...
}
export default MyCustomModule;
```

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -26,6 +26,7 @@ under the licensing terms detailed in LICENSE:
 * ncave <777696+ncave@users.noreply.github.com>
 * Andrew Davis <pulpdrew@gmail.com>
 * Maël Nison <nison.mael@gmail.com>
+* Wietse Wind <w.wind@ipublications.net>
 
 Portions of this software are derived from third-party works licensed under
 the following terms:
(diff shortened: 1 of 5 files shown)


dcodeIO:
The usage here is

```ts
import NotASModule from "pathToModule";
```

implicitly renaming the default export to whatever identifier is used, even if it's `ASModule` internally. As such seems unnecessary, or is there a use case where this doesn't just work?

PR #1459, opened 2020-09-06
Rename index arg

DuncanUszkay1:
Before we had first class functions, we used this `indexArg` local in `compileCallIndirect`. From my understanding, this name no longer makes sense because it is not in fact an index into the function table but actually a pointer into a block of memory which contains that index.

I picked the name `functionInstance` but I'm not terribly attached to it, I just think that `indexArg` is the wrong name (Unless I'm misreading the code of course)

This was tripping me up when I was reviewing the code so I figured I'd submit a patch 🤷 


DuncanUszkay1:
cc @dcodeIO

src/builtins.ts
```diff
@@ -2997,7 +2997,7 @@ function builtin_function_call(ctx: BuiltinContext): ExpressionRef {
     compiler.currentType = returnType;
     return compiler.module.unreachable();
   }
-  var indexArg = compiler.compileExpression(assert(ctx.thisOperand), ftype, Constraints.CONV_IMPLICIT);
+  var functionInstance = compiler.compileExpression(assert(ctx.thisOperand), ftype, Constraints.CONV_IMPLICIT);
```
dcodeIO, line 3000: We are using `functionInstance` at other places to indicate a `Function`, so perhaps `functionPointer`?

src/compiler.ts
```diff
@@ -8148,6 +8148,7 @@ export class Compiler extends DiagnosticEmitter {
     contextualSignature: Signature | null,
     constraints: Constraints
   ): ExpressionRef {
+    console.log("compiling function expression");
```
dcodeIO, line 8151:
```suggestion
```

dcodeIO (commented):
Makes sense to rename it in context of `compileCallExpression` etc., yeah, except in context of the `call_indirect` builtin, where it's still an `indexArg`. Perhaps, a better name for `indexArg` in `call_indirect` might be `tableIndex`.

src/builtins.ts
dcodeIO: Or maybe, to stick to other `expr` names, `functionPointerExpr` and `tableIndexExpr` respectively.

src/builtins.ts
```diff
@@ -2997,7 +2997,7 @@ function builtin_function_call(ctx: BuiltinContext): ExpressionRef {
     compiler.currentType = returnType;
     return compiler.module.unreachable();
   }
-  var indexArg = compiler.compileExpression(assert(ctx.thisOperand), ftype, Constraints.CONV_IMPLICIT);
+  var tableIndexExpr = compiler.compileExpression(assert(ctx.thisOperand), ftype, Constraints.CONV_IMPLICIT);
```
dcodeIO, line 3000: Even though a builtin as well, this one is the implementation of `Function#call`, essentially doing an idiomatic call by function pointer. Only `builtin_call_indirect` is special.

src/compiler.ts
```diff
@@ -8004,13 +8004,13 @@ export class Compiler extends DiagnosticEmitter {
       );
     }
     assert(index == numArgumentsInclThis);
-    return this.makeCallIndirect(signature, indexArg, reportNode, operands, immediatelyDropped);
+    return this.makeCallIndirect(signature, functionPointerExpr, reportNode, operands, immediatelyDropped);
   }
 
-  /** Creates an indirect call to the function at `indexArg` in the function table. */
+  /** Creates an indirect call to the function at `functionInstance` in the function table. */
```
dcodeIO, line 8010: The comment here needs to be updated now as well. Perhaps

```suggestion
  /** Creates an indirect call to the first-class function pointed to by `functionPointerExpr`. */
```

DuncanUszkay1:
@dcodeIO  does this change still make sense?

PR #1460, opened 2020-09-06
Runtime created first class functions

DuncanUszkay1:
I noticed that we were using one statically allocated memory region per function.

This won't work for Closures, since there will be many instances of a single function which all might have different values for `_env`.

In this PR, I allocate a copy of the RuntimeInstance version of the function into memory at Runtime. This will allow us to set the `_env` for each function instance independently once it exists.


src/compiler.ts
```diff
@@ -2001,6 +2001,54 @@ export class Compiler extends DiagnosticEmitter {
...
+    var module = this.module;
+    var options = this.options;
+
+    var rtInstanceOffset = this.ensureRuntimeFunction(instance);
+
+    // Create a new instance of the function object in memory
+    var rtInstance = assert(this.resolver.resolveClass(program.functionPrototype, [ instance.type ]));
+    var currentType = this.currentType;
+    var functionInstanceExpr = this.makeAllocation(rtInstance)
+    this.currentType = currentType;
```
DuncanUszkay1, line 2019: Because `makeAllocation` changes the current type into `Function` without any signature, we need to discard that type.

I'm not entirely sure what the desired behaviour is here. We discard the type in `ensureRuntimeFunction` by just returning an offset,  so I'm making the assumption here that we'll want to do the same.

src/compiler.ts
```diff
@@ -2001,6 +2001,57 @@ export class Compiler extends DiagnosticEmitter {
...
+    var program = this.program;
+    var module = this.module;
+    var options = this.options;
+
+    var rtInstanceOffset = this.ensureRuntimeFunction(instance);
+
+    // Create a new instance of the function object in memory
+    var rtInstance = assert(this.resolver.resolveClass(program.functionPrototype, [ instance.type ]));
+    var currentType = this.currentType;
+    var functionInstanceExpr = this.makeAllocation(rtInstance);
```
DuncanUszkay1, line 2018: It seems like this should leak, but the tests are green 🤔

DuncanUszkay1:
cc @torch2424 @dcodeIO Small PR that takes a step towards closures which I think might be mergeable

dcodeIO (commented):
While such a mechanism will be needed ultimately, it looks like more preliminary work is necessary before merging code assuming the presence of the necessary concepts.

* Detect functions for which the mechanism is applicable (closures over locals in the Wasm sense)
* Figure out a concrete representation of `env`s, including for closures in closures with shared parent `env`s

These two tasks somewhat require thinking about a good design and an efficient implementation/layout before implementing something, but are necessary no matter if we are building on top of the runtime or on top of the GC spec.

DuncanUszkay1:
>Detect functions for which the mechanism is applicable (closures over locals in the Wasm sense)

Do we really need this to merge or do we just need to assume it's possible (which it certainly is)? Are we concerned with the memory overhead of having non-closure functions having their own copy of the runtime instance? Seems to me like an optimization we could make later but if it's required then fine let's back burner this and we can build that first.

>Figure out a concrete representation of envs, including for closures in closures with shared parent envs

Unless we think that the env won't be in linear memory, then why would we need to know the concrete representation to merge this? All this PR asserts is that it will be represented by a pointer. (Which is already asserted by the first class function class on the master branch).

dcodeIO:
> Do we really need this to merge

Would say so, because otherwise we are slowing down and increasing the memory footprint of all existing code doing indirect calls but not utilizing closures. Seems important to avoid to me.

> why would we need to know the concrete representation to merge this?

We don't strictly need this change without `env`s, in fact even though it looks like an easy merge it's also a surface for regression as seen in the previous comment. As such I'd favor to think about the critical parts first. I'd be ok with stubbing parts of this out as a non-functional change, if it can be done and there are other ways to test the additions separately.

torch2424:
Ah! Thanks for opening this @DuncanUszkay1 ! :) My apologies for not touching closures over the past couple of weeks. All my "in-between PRs" time has been caught in writing / practicing AssemblyScript talks I'll be giving at some conferences haha! 

I'll follow this PR :smile: :+1:

DuncanUszkay1:
Closing out old PRs

PR #1464, opened 2020-09-13
Closures with First Class Functions 1: Gathering Context during Compilation

DuncanUszkay1:
**Ignore the changes to the github workflows**

**Not to be merged, just reviewed**

As the first step of the plan defined [here](https://github.com/AssemblyScript/assemblyscript/issues/798#issuecomment-688542612), I'm updating the `Function` class's `lookup` function to return a new type of `Element`. This type is called `ClosedLocal` and it represents a reference to a closed over `Local`. Each reference to a closed over `Local` has it's own `ClosedLocal` instance with the function depth attached. Here's an example:
```typescript
function generator(): (a: i32) => (b: i32) => i32 {
  COUNTER += 1;
  var c = COUNTER * 2;
  return (b: i32): (a: i32) => i32 => {
    var d = COUNTER;
    return (a: i32) => { return a + b + c + d; }
  }
}
```
The inner references to `b`, `c` and `d` each have their own `ClosedLocal`. `b` and `d` have a functionDepth of 1, and `c` has a `functionDepth` of 2.

I've added logging statements that vaguely describe what will be done with this information in future PRs. Here's the output resulting from the above block:
```
marking local b as closed in function anonymous|0
Accessing a closed local b at depth 1 and offset 0
marking local c as closed in function anonymous|0
Accessing a closed local c at depth 2 and offset 0
marking local d as closed in function anonymous|0
Accessing a closed local d at depth 1 and offset 4
writing b into the env of anonymous|0 at offset 0
writing d into the env of anonymous|0 at offset 4
writing c into the env of generator at offset 0
```

When we say `writing b into the env of anonymous|0 at offset` we mean changing all references to the local b into references to the environment. 

More specifics about the env struct will come in a future PR, for now I just wanted to ensure that we're OK with this method for deriving the necessary contextual information for closed over locals.


src/compiler.ts
```diff
@@ -1483,6 +1484,12 @@ export class Compiler extends DiagnosticEmitter {
       instance.set(CommonFlags.ERRORED);
     }
 
+    instance.localsByIndex
+      .filter((local: Local) => local.envOffset >= 0)
+      .forEach((local: Local) => {
+      console.log("writing " + local.name + " into the env of " + instance.name + " at offset " + local.envOffset);
+    })
```
DuncanUszkay1, line 1491: This is performed post-compilation- I figured it would be simplest to gather all the closure information by compiling *then* go back and update all the binaryen statements to load from the environment instead where necessary.

src/compiler.ts
```diff
@@ -8439,17 +8446,23 @@ export class Compiler extends DiagnosticEmitter {
...
-        }
+        //if (target.parent != flow.parentFunction) {
+          //// TODO: closures
+          //this.error(
+            //DiagnosticCode.Not_implemented_0,
+            //expression.range,
+            //"Closures"
+          //);
+          //return module.unreachable();
+        //}
```
DuncanUszkay1, line 8457: This should be unreachable now, since in this case we would return a `CLOSED_LOCAL` instead

src/compiler.ts
```diff
@@ -8439,17 +8446,23 @@ export class Compiler extends DiagnosticEmitter {
...
         return module.local_get(localIndex, localType.toNativeType());
       }
+      case ElementKind.CLOSED_LOCAL: {
+        var closedLocal = <ClosedLocal>target;
+        console.log("Accessing a closed local " + closedLocal.local.name + " at depth " + closedLocal.functionDepth + " and offset " + closedLocal.local.envOffset);
+
+        return module.i32(1);
```
DuncanUszkay1, line 8464: Whenever we try to access a local from an env, we'll just return a hardcoded 1 for now. Allowing the compilation to complete will help us validate that the outputted binary isn't invalid before we can replace this with an actual access

src/program.ts
```diff
@@ -2996,8 +2998,8 @@ export class File extends Element {
 
   /** Creates an imported namespace from this file. */
   asAliasNamespace(
-    name: string, 
-    parent: Element, 
+    name: string,
+    parent: Element,
```
DuncanUszkay1, line 3002: These had extra spaces 🤔  is that intentional?

src/program.ts
```diff
@@ -3291,12 +3293,35 @@ export class Parameter {
...
+      local.parent,
+      <VariableLikeDeclarationStatement>local.declaration
+    )
+    this.local = local;
+  }
+
+  lookup(name: string) {
+    return this.local.lookup(name);
+  }
+}
```
DuncanUszkay1, line 3313: This is meant to act as a wrapper over a Local to add the additional contextual information of the `functionDepth` which tells us how many environments up we'll need to go before we can find this local

src/program.ts
```diff
@@ -3588,11 +3616,37 @@ export class Function extends TypedElement {
     return local;
   }
 
+  markClosedLocal(local: Local) {
+    console.log("marking local " + local.name + " as closed in function " + this.name);
+    local.envOffset = this.nextEnvOffset;
+    // TODO: the alignment logic is slightly more complicated then this iirc
```
DuncanUszkay1, line 3622: I'm pretty sure there's somewhere in the compiler that already does this I copied off last time, but it's not terribly important for this PR so I'm leaving it.

src/program.ts
```diff
@@ -3588,11 +3616,37 @@ export class Function extends TypedElement {
...
+    var parentResult = this.parent.lookup(name);
+    if (parentResult !== null) {
+     /* We found a local in a parent of a function, mark that it must be enclosed
+     * We can infer that since this is a Local and not a Closed Local that the function
+     * depth must be 1 */
+      if (parentResult.kind == ElementKind.LOCAL) {
+        assert(this.parent.kind == ElementKind.FUNCTION);
+        /* Notify the parent that this local must be included in its env. This will also
+         * populate the Local with information about its storage within the Env */
+        (<Function>this.parent).markClosedLocal(<Local>parentResult);
```
DuncanUszkay1, line 3641: If I'm wrong about this assertion and it's possible to have a middle element between two functions, then we'll need to figure out a way to port the `makeClosedLocal` call through another `Element`. Not the end of the world but it's a bunch of added boilerplate.

However, if I'm wrong and there could be a local found in an ancestor of a function which does not belong to a function then this would become more complicated.

src/program.ts
```diff
@@ -3588,11 +3617,47 @@ export class Function extends TypedElement {
...
+        assert(this.parent.kind == ElementKind.FUNCTION);
+        /* Return a ClosedLocal with the inferred depth of 1 */
+        return new ClosedLocal(<Local>parentResult)
+      } else if (parentResult.kind == ElementKind.CLOSED_LOCAL) {
+        /* We add 1 to the depth to indicate another level of nesting */
+        var parentClosedLocal = <ClosedLocal>parentResult;
+        return new ClosedLocal(parentClosedLocal.local, parentClosedLocal.functionDepth + 1);
+      }
+    }
+    return parentResult;
```
DuncanUszkay1, line 3660: I've moved the state-changing logic out of `lookup`, since as @MaxGraey pointed out it's certainly counterintuitive to have a function called `lookup` changing state. Instead, the caller is expected to call markClosedLocal after receiving a `ClosedLocal` response, as seen above.

torch2424 (commented):
This all looks good to me @DuncanUszkay1 . I think what you're doing makes sense :smile: 

Thank you very much for all of this! 

I'll defer to @dcodeIO for the final approval :+1:

src/program.ts
dcodeIO: I like the underlying idea, but I'm also skeptical that `lookup` is the right place for it. For instance, we'd want to deduplicate the same closed local when returned for the second time inside of the same function, both to reduce allocations and to be able to use them as keys in collections, but then we're back to square one because remembering it changes state. As an alternative, an API like `Function#ensureClosedLocal(local: Local): ClosedLocal` that is called after `lookup` upon detecting a local outside of the current function's scope might work. This function would either return the remembered closed local for the given local, or bubble up its parents until it finds the matching local, and only then create a new closed local to return.

src/program.ts
dcodeIO: Not intentional, would guess this somehow slipped through.

src/compiler.ts
dcodeIO: This would be a better place than inside of `lookup` for obtaining closed locals, for example by means of `Function#ensureClosedLocal` as mentioned below. I am also not sure if the `lookup` above is able to return scoped locals, or if it just returns top-level locals of a function, so that might have been tackled as well. Then also leads back to the challenge that scoped locals are shared between different `let`s. Scenarios: Local has been used for a different `let` before: Unable to figure out which local accesses to replace when replacing with env accesses. Local will be used by a different `let` after: Not the same as in the env.

src/compiler.ts
dcodeIO: This is somewhat connected to the problem of scoped locals being shared, since if we replace too late we can encounter other accesses on the same shared local that are not actually part of the env. My intuition here is to remove sharing locals during code generation, but last time I checked this had surprising implications on other logic in that we cannot recompile blocks of code anymore without changing local order.

src/program.ts
dcodeIO: Introducing an on-demand wrapper like this makes sense to me, ideally also deduplicating these wrappers per function as mentioned further down.

src/compiler.ts
DuncanUszkay1: Hm- could you give a code example of this scoped local issue?

src/compiler.ts
DuncanUszkay1: When you say that the 'Local' is shared- are you referring to the compiler struct `Local` or the local index referred to in the binaryen IR?

dcodeIO: Currently it's both. Scoped locals can be free'd when their lifetime ends, and subsequent code can reuse one from the `tempXY` arrays,, including its index.

src/compiler.ts
dcodeIO: A simple example for the reuse problem is

```ts
{
  let a = 1;
}
{
  let b = 2;
}
```

where both `a` and `b` are of the same type. At the end of the first block, `a` is free'd, and upon `let b` the underlying local incl. its index is reused.

The lookup problem is that

```ts
var a;
someFunc() {
  a
}
```

works, but

```ts
{
  let a;
  someFunc() {
    a
  }
}
```

may not. Didn't check in detail, but iirc the latter one doesn't even trigger `Not implemented: Closures` currently.

DuncanUszkay1:
Next Steps here from @dcodeIO's review:
- Refactor the lookup sequence to avoid excess allocations.
- Come up with a solution for the scoped local problem. I'm going to take this problem back into the issue, since I think it's a non-trivial problem.

torch2424:
@dcodeIO @DuncanUszkay1 Hello! :smile: 

So I am trying to catch back up on the code side of things for closures, as I am starting to get more time for it again :smile: :tada: 

So, I know we discussed about the scoped locals, and locals re-use at this meeting: https://github.com/AssemblyScript/working-group/issues/46

If I'mt no mistaken, it kind of came down to, we just need a better GC correct? Which @dcodeIO is working on. Thus, this is blocked in the meantime? :thinking: 

My bad if this was already made clear, but I think it'd also be nice for anyone following closures PRs haha! :smile:

DuncanUszkay1:
@torch2424 That's right, this is currently blocked on local reuse issues and GC refactoring. 

There's likely some work which could be done to push this forward despite those two issues, but I'm expecting that the refactoring away from arc references will have a very large diff, and that it'll be worth waiting for that to land to avoid getting into massive conflicts.

torch2424:
@DuncanUszkay1 Awesome! Thanks for the update! I have a meeting with Saul next week, so I'm sure I'll be able to find other places where we can collaborate on AS stuff, while closures mellows back out into a good jumping in point :smile: :tada:

DuncanUszkay1:
Closing out old PRs

trusktr:
New GC is in place. Really looking forward to when this will be re-visited! :)

PR #1522, opened 2020-10-26
add cli exports

jtenner:
⯈
⯈
⯈

Hey! Just thought these might be useful for as-pect. Do you think we should add a few more?

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,6 +46,9 @@
   "types": "index.d.ts",
   "exports": {
     ".": "./index.js",
+    "./cli/asc": "./cli/asc.js",
+    "./dist/asc": "./dist/asc.js",
+    "./cli/util/options": "./cli/util/options.js",
     "./lib/loader": {
       "import": "./lib/loader/index.js",
       "require": "./lib/loader/umd/index.js"


dcodeIO:
Did you try what happens when `exports` is removed? Given that there may be more exports we'd have to add, that seems like a viable alternative for the time being :)

jtenner:
I don't know? After some testing, I discovered that adding those to the exports works just fine. Are you sure we need to remove `"exports"` altogether?

dcodeIO:
Playing around with this currently, and apparently this already works in node 15:

```json
  "exports": {
    ".": "./index.js",
    "./cli/*": "./cli/*.js",
    "./lib/*": {
      "import": "./lib/*/index.js",
      "require": "./lib/*/umd/index.js"
    }
  },
```

dcodeIO:
Also tried what happens when removing `exports` now, and that works with `require` as before it seems, but with `import` one has to append the `.js` file extension then.

dcodeIO:
Another one: Appending `"./*": "./*"` as the last exports entry allows using anything within the package by means of requiring/importing its full file name.

jtenner:
Hey. Sorry I couldn't swing around to this. I assume we are closing in favor of #1527 ?

PR #1527, opened 2020-10-28
Add package exports fallback for the CLI, dist files, etc.

dcodeIO:
With the fallback, legacy cases like

```ts
import asc from "assemblyscript/cli/asc";
import options from "assemblyscript/cli/util/options";
import distAsc from "assemblyscript/dist/asc";
```

including others we haven't though of yet are covered by always resolving `./* -> ./*.js`. Fixes https://github.com/AssemblyScript/assemblyscript/pull/1522

At some point we'll most likely want to break this, either to pull packages out of the main packages or make nicer entry points like just `/loader`, but this gives us sufficient backwards compatibility (on node >= 15, need to check 14) for now so is good to do :)

**Edit:** According to the node 14 docs, the fallback should work there as well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,12 @@
       "import": "./lib/rtrace/index.js",
       "require": "./lib/rtrace/umd/index.js"
     },
-    "./*": "./*.js"
+    "./*": "./*.js",
+    "./cli/asc": "./cli/asc.js",
+    "./cli/transform": "./cli/transform.js",
+    "./cli/util/options": "./cli/util/options.js",
+    "./dist/assemblyscript": "./dist/assemblyscript.js",
+    "./dist/asc": "./dist/asc.js"
   },
   "bin": {
     "asc": "bin/asc",
--- a/package.json
+++ b/package.json
@@ -53,7 +53,8 @@
     "./lib/rtrace": {
       "import": "./lib/rtrace/index.js",
       "require": "./lib/rtrace/umd/index.js"
-    }
+    },
+    "./*": "./*.js"
   },
   "bin": {
     "asc": "bin/asc",


SebastianSpeitel:
Maybe a test suite which tries to import/resolve the subpackages/named exports under different conditions (node versions, type=commonjs, type=module, using TS, using different bundlers, ...) could be useful.

dcodeIO:
I agree that having actual tests would be best, yeah. Needs to somehow install the package before it's published then, via a `file:` dependency or the like I guess.

dcodeIO:
Also note that 0.17.0 breaks with node <=12 anyhow (14 entered LTS just recently) due to making the ES module the default, with `/umd/index.js` being what to use on older node versions now. Hence, for exports syntax, node >=14 is of interest. Going to test with a node 12 manually for now to see how it behaves, so I can give proper recommendations.

dcodeIO:
Turns out that node 12 has some sort of experimental support with a warning:

```
(node:4732) ExperimentalWarning: The ESM module loader is experimental.
internal/process/esm_loader.js:90
    internalBinding('errors').triggerUncaughtException(
                              ^

Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './cli/asc' is not defined by "exports" in path\to\node_modules\assemblyscript\package.json imported from path\to\test\esm\index.js
```

and does not understand the `*` syntax yet, so to support node 12 this change needs to be

```
    "./*": "./*.js",
    "./cli/asc": "./cli/asc.js",
    "./cli/util/options": "./cli/util/options.js",
    "./dist/asc": "./dist/asc.js"
```

with the fallback triggering for node >=14, and being otherwise ignored.

dcodeIO:
node 12 docs: https://nodejs.org/dist/latest-v12.x/docs/api/esm.html#esm_package_entry_points
node 14 docs: https://nodejs.org/dist/latest-v14.x/docs/api/packages.html#packages_package_entry_points
node 15 docs. https://nodejs.org/dist/latest-v15.x/docs/api/packages.html#packages_package_entry_points (== 14)

dcodeIO:
For completeness, node 10 breaks on `TextDecoder` used as a global in the loader btw, and does not understand `exports` syntax at all, requiring `lib/loader/umd` otherwise. Probably not worth to support, but then maybe fallbacks for 12, simply because we can, are fan.

dcodeIO:
Merging this for now since I expect most problems to be solved by it, but I also have an external test package now that I can first verify with and perhaps make a test of eventually.

dcodeIO:
Test package: https://github.com/dcodeIO/assemblyscript-esm-test/runs/1326736751

| Import type | ^15 | 15.0.0 | ^14 | 14.0.0 | ^13 | 13.0.0 | ^12 | 12.0.0 | ^11 | 11.0.0 | ^10 | 10.0.0 |
| -------------|--------------|--------|-----|--------|-----|--------|------|--------|------|-------|------|--------|
| `import "pkg"` | ✔ | ✔ | ✔ | ✔ | ✔ | ❌<sub>2</sub> | ✔ | ❌<sub>3</sub> | ❌<sub>3</sub> | ❌<sub>3</sub> | ❌<sub>3</sub> | ❌<sub>3</sub> |
| `require("pkg")` | ✔ | ✔ | ✔ | ✔ | ✔ | ❌<sub>1</sub> | ✔ | ❌<sub>4</sub> | ❌<sub>4</sub> | ❌<sub>4</sub> | ❌<sub>4</sub> | ❌<sub>4</sub> |
| `require("pkg/umd")` | ✔ | ✔ | ✔ | ❌<sub>1</sub> | ❌<sub>1</sub> | ❌<sub>1</sub> | ❌<sub>1</sub> | ✔ | ✔ | ✔ | ❌<sub>5</sub> | ❌<sub>5</sub> |

1. Package subpath 'pkg' is not defined by "exports" in package.json
2. SyntaxError: Cannot use import statement outside a module
3. SyntaxError: Unexpected identifier
4. SyntaxError: Unexpected token export
5. ReferenceError: TextDecoder is not defined

jtenner:
So it works with node 14 and 15 just fine? I'm okay with that.

PR #1558, opened 2020-11-23
RTrace TypeScript Types Update

jtenner:
Hello! I was trying to extend the RTrace class and found out that some of the types were incomplete and missing. The pull request currently contains everything as-pect would need to properly extend the class. Is there anything that can be added which would be helpful to others?

I am also very open to suggestions on wording the comments.


jtenner:
I wasn't sure if this was enough? Just wanted to make sure it was reasonable to add. Thanks for the quick pull!

PR #1559, opened 2020-11-24
Rewrite runtime, switch to tracing GC and bootstrap

dcodeIO:
Removes the error-prone automatic reference counting implementation that has been haunting us for a long time and replaces it with a much simpler mark & sweep garbage collector. By doing so this PR also sets us up for initial experiments with Wasm GC due to introducing a very similar interface already.

* Garbage collection is not automatic anymore, but one has to manually call `__collect()` when the execution stack is unwound (typically externally) to perform a full garbage collection cycle. Has a few drawbacks, but many more advantages, from making the external interface to AS modules much easier to comprehend, to simplifying the compiler to a point where the core team can make meaningful progress again.
* Speaking of making matters easier, the `__retain` / `__release` mechanism has been removed and replaced with a new `__pin` / `__unpin` mechanism. Unlike the former, the new mechanism is optional most of the time, unless an external object (that is not being kept alive in Wasm) needs to survive the next call to `__collect()`.
* Supported runtime variants now are: `stub`, `default` and `incremental`
* Adds an `--exportRuntime` flag to achieve the effects of `--runtime stub/full` with any runtime (i.e. exporting `__new`, `__pin`, `__unpin` and `__collect`).
* Bootstraps the compiler 🎉 


lib/loader/umd/index.js
```diff
@@ -46,6 +46,7 @@ var loader = (function(exports) {
   const THIS = Symbol();
   const STRING_DECODE_THRESHOLD = 32;
   const decoder = new TextDecoder("utf-16le");
+  const rootsSym = Symbol();
```
ZoeDreams, line 49: awesome use of the newish symbol type. this works very fast in memory compared to leverage a map as a dict.

ZoeDreams (commented):
i like your implementation on the tri-color trace sweep algorithm.

dcodeIO:
Turns out that adding a shadow stack has one unfortunate consequence: It affects the `--lowMemoryLimit` mechanism, which is intended for IOT environments with [2..16kb](https://github.com/AssemblyScript/assemblyscript/issues/1089), where even conservative maximum stack sizes may easily overrun the limit by just reserving stack space. To somewhat keep supporting low memory use cases, we'll likely want a `--stackSize` compiler flag so one can experiment with very small stack sizes, with any reasonable default never fitting.

dcodeIO:
Another: Putting incomplete objects onto the stack can break things if incremental GC just so happens to run in between allocating the object and properly initializing it. Looks like allocations need to zero memory to fix this, something we didn't do before to reduce overhead from zeroing memory we are going to initialize anyway.

dcodeIO:
Meanwhile, the shadow stack thingy is getting together. Will prepare something for tomorrow, but for the time being, here's the

<details><summary>bootstrap with incremental GC</summary>

```
> ts-node tests/bootstrap

trace: GC (incremental) at mem/objs 23472, 100
trace:  down to mem/objs (sweep complete) 14176, 88

Parsing standard library ...
  ~lib/array.ts
trace: GC (incremental) at mem/objs 57968, 176
trace:  down to mem/objs (sweep complete) 57328, 157
trace: GC (incremental) at mem/objs 64544, 314
trace:  down to mem/objs (sweep complete) 62560, 261
trace: GC (incremental) at mem/objs 73936, 522
trace:  down to mem/objs (sweep complete) 71952, 466
trace: GC (incremental) at mem/objs 91824, 932
trace:  down to mem/objs (sweep complete) 87120, 800
trace: GC (incremental) at mem/objs 121264, 1600
trace:  down to mem/objs 114112, 1400
trace: GC (incremental) at mem/objs 118304, 1500
trace:  down to mem/objs (sweep complete) 117344, 1472
trace: GC (incremental) at mem/objs 179456, 2944
trace:  down to mem/objs 172496, 2744
trace: GC (incremental) at mem/objs 176688, 2844
trace:  down to mem/objs 169744, 2644
trace: GC (incremental) at mem/objs 173904, 2744
trace:  down to mem/objs (sweep complete) 173808, 2741
trace: GC (incremental) at mem/objs 288848, 5482
trace:  down to mem/objs 281984, 5282
trace: GC (incremental) at mem/objs 286112, 5382
trace:  down to mem/objs 279104, 5182
trace: GC (incremental) at mem/objs 283392, 5282
trace:  down to mem/objs 276608, 5082
trace: GC (incremental) at mem/objs 280720, 5182
trace:  down to mem/objs 273872, 4982
trace: GC (incremental) at mem/objs 278016, 5082
trace:  down to mem/objs (sweep complete) 277056, 5056
trace: GC (incremental) at mem/objs 488736, 10112
trace:  down to mem/objs 481904, 9912
trace: GC (incremental) at mem/objs 486032, 10012
trace:  down to mem/objs 479408, 9812
trace: GC (incremental) at mem/objs 483568, 9912
trace:  down to mem/objs 476672, 9712
trace: GC (incremental) at mem/objs 480832, 9812
trace:  down to mem/objs 474000, 9612
trace: GC (incremental) at mem/objs 478176, 9712
trace:  down to mem/objs 471408, 9512
trace: GC (incremental) at mem/objs 475744, 9612
trace:  down to mem/objs 468752, 9412
trace: GC (incremental) at mem/objs 473040, 9512
trace:  down to mem/objs 466304, 9312
trace: GC (incremental) at mem/objs 470480, 9412
trace:  down to mem/objs (sweep complete) 467872, 9337
  ~lib/arraybuffer.ts
  ~lib/atomics.ts
trace: GC (incremental) at mem/objs 877408, 18674
trace:  down to mem/objs 870624, 18474
  ~lib/builtins.ts
trace: GC (incremental) at mem/objs 987328, 18574
trace:  down to mem/objs 980368, 18374
trace: GC (incremental) at mem/objs 984720, 18474
trace:  down to mem/objs 977968, 18274
trace: GC (incremental) at mem/objs 982336, 18374
trace:  down to mem/objs 975264, 18174
trace: GC (incremental) at mem/objs 979680, 18274
trace:  down to mem/objs 972496, 18074
trace: GC (incremental) at mem/objs 976896, 18174
trace:  down to mem/objs 969840, 17974
trace: GC (incremental) at mem/objs 974192, 18074
trace:  down to mem/objs 966848, 17874
trace: GC (incremental) at mem/objs 971264, 17974
trace:  down to mem/objs 963968, 17774
trace: GC (incremental) at mem/objs 968432, 17874
trace:  down to mem/objs 961632, 17674
trace: GC (incremental) at mem/objs 966048, 17774
trace:  down to mem/objs 958880, 17574
trace: GC (incremental) at mem/objs 963312, 17674
trace:  down to mem/objs 956144, 17474
trace: GC (incremental) at mem/objs 960512, 17574
trace:  down to mem/objs 953408, 17374
trace: GC (incremental) at mem/objs 957808, 17474
trace:  down to mem/objs 950672, 17274
trace: GC (incremental) at mem/objs 955008, 17374
trace:  down to mem/objs 947600, 17174
trace: GC (incremental) at mem/objs 951936, 17274
trace:  down to mem/objs (sweep complete) 950848, 17241
trace: GC (incremental) at mem/objs 1706048, 34482
trace:  down to mem/objs 1698240, 34282
trace: GC (incremental) at mem/objs 1702560, 34382
trace:  down to mem/objs 1694528, 34182
trace: GC (incremental) at mem/objs 1698896, 34282
trace:  down to mem/objs 1690960, 34082
trace: GC (incremental) at mem/objs 1695360, 34182
trace:  down to mem/objs 1687328, 33982
trace: GC (incremental) at mem/objs 1691776, 34082
trace:  down to mem/objs 1683792, 33882
trace: GC (incremental) at mem/objs 1688224, 33982
trace:  down to mem/objs 1680176, 33782
trace: GC (incremental) at mem/objs 1684624, 33882
trace:  down to mem/objs 1677152, 33682
trace: GC (incremental) at mem/objs 1681600, 33782
trace:  down to mem/objs 1673776, 33582
trace: GC (incremental) at mem/objs 1678176, 33682
trace:  down to mem/objs 1670288, 33482
trace: GC (incremental) at mem/objs 1674720, 33582
trace:  down to mem/objs 1666816, 33382
trace: GC (incremental) at mem/objs 1671152, 33482
trace:  down to mem/objs 1663200, 33282
trace: GC (incremental) at mem/objs 1667536, 33382
trace:  down to mem/objs 1659632, 33182
trace: GC (incremental) at mem/objs 1664032, 33282
trace:  down to mem/objs 1656224, 33082
trace: GC (incremental) at mem/objs 1660576, 33182
trace:  down to mem/objs 1652624, 32982
trace: GC (incremental) at mem/objs 1656976, 33082
trace:  down to mem/objs 1649200, 32882
trace: GC (incremental) at mem/objs 1653600, 32982
trace:  down to mem/objs 1645760, 32782
trace: GC (incremental) at mem/objs 1650080, 32882
trace:  down to mem/objs 1642112, 32682
trace: GC (incremental) at mem/objs 1646464, 32782
trace:  down to mem/objs 1638448, 32582
trace: GC (incremental) at mem/objs 1642848, 32682
trace:  down to mem/objs 1634928, 32482
trace: GC (incremental) at mem/objs 1639344, 32582
trace:  down to mem/objs 1631488, 32382
trace: GC (incremental) at mem/objs 1635936, 32482
trace:  down to mem/objs 1628688, 32282
trace: GC (incremental) at mem/objs 1633056, 32382
trace:  down to mem/objs 1625888, 32182
trace: GC (incremental) at mem/objs 1630288, 32282
trace:  down to mem/objs 1622624, 32082
trace: GC (incremental) at mem/objs 1626912, 32182
trace:  down to mem/objs 1619280, 31982
trace: GC (incremental) at mem/objs 1623632, 32082
trace:  down to mem/objs 1616160, 31882
trace: GC (incremental) at mem/objs 1620576, 31982
trace:  down to mem/objs 1612752, 31782
trace: GC (incremental) at mem/objs 1617120, 31882
trace:  down to mem/objs (sweep complete) 1614144, 31805
  ~lib/compat.ts
  ~lib/dataview.ts
  ~lib/date.ts
  ~lib/diagnostics.ts
  ~lib/error.ts
  ~lib/function.ts
  ~lib/iterator.ts
  ~lib/map.ts
  ~lib/math.ts
trace: GC (incremental) at mem/objs 3248624, 63610
trace:  down to mem/objs 3240576, 63410
trace: GC (incremental) at mem/objs 3244784, 63510
trace:  down to mem/objs 3236896, 63310
trace: GC (incremental) at mem/objs 3241024, 63410
trace:  down to mem/objs 3233136, 63210
trace: GC (incremental) at mem/objs 3237440, 63310
trace:  down to mem/objs 3229568, 63110
trace: GC (incremental) at mem/objs 3233744, 63210
trace:  down to mem/objs 3225824, 63010
trace: GC (incremental) at mem/objs 3230080, 63110
trace:  down to mem/objs 3222160, 62910
trace: GC (incremental) at mem/objs 3226432, 63010
trace:  down to mem/objs 3218624, 62810
trace: GC (incremental) at mem/objs 3222992, 62910
trace:  down to mem/objs 3215152, 62710
trace: GC (incremental) at mem/objs 3219504, 62810
trace:  down to mem/objs 3211696, 62610
trace: GC (incremental) at mem/objs 3216048, 62710
trace:  down to mem/objs 3208224, 62510
trace: GC (incremental) at mem/objs 3212464, 62610
trace:  down to mem/objs 3204640, 62410
trace: GC (incremental) at mem/objs 3208656, 62510
trace:  down to mem/objs 3200832, 62310
trace: GC (incremental) at mem/objs 3204864, 62410
trace:  down to mem/objs 3197024, 62210
trace: GC (incremental) at mem/objs 3200960, 62310
trace:  down to mem/objs 3193040, 62110
trace: GC (incremental) at mem/objs 3197136, 62210
trace:  down to mem/objs 3189328, 62010
trace: GC (incremental) at mem/objs 3193344, 62110
trace:  down to mem/objs 3185520, 61910
trace: GC (incremental) at mem/objs 3189712, 62010
trace:  down to mem/objs 3181872, 61810
trace: GC (incremental) at mem/objs 3186096, 61910
trace:  down to mem/objs 3178176, 61710
trace: GC (incremental) at mem/objs 3182080, 61810
trace:  down to mem/objs 3174224, 61610
trace: GC (incremental) at mem/objs 3178384, 61710
trace:  down to mem/objs 3170528, 61510
trace: GC (incremental) at mem/objs 3174512, 61610
trace:  down to mem/objs 3166736, 61410
trace: GC (incremental) at mem/objs 3170768, 61510
trace:  down to mem/objs 3162992, 61310
trace: GC (incremental) at mem/objs 3167088, 61410
trace:  down to mem/objs 3159264, 61210
trace: GC (incremental) at mem/objs 3163488, 61310
trace:  down to mem/objs 3155680, 61110
trace: GC (incremental) at mem/objs 3160032, 61210
trace:  down to mem/objs 3152224, 61010
trace: GC (incremental) at mem/objs 3156560, 61110
trace:  down to mem/objs 3148144, 60910
trace: GC (incremental) at mem/objs 3152320, 61010
trace:  down to mem/objs 3144480, 60810
trace: GC (incremental) at mem/objs 3148592, 60910
trace:  down to mem/objs 3141712, 60710
trace: GC (incremental) at mem/objs 3145584, 60810
trace:  down to mem/objs 3138768, 60610
trace: GC (incremental) at mem/objs 3143056, 60710
trace:  down to mem/objs 3136080, 60510
trace: GC (incremental) at mem/objs 3140432, 60610
trace:  down to mem/objs 3133552, 60410
trace: GC (incremental) at mem/objs 3137888, 60510
trace:  down to mem/objs 3130752, 60310
trace: GC (incremental) at mem/objs 3134864, 60410
trace:  down to mem/objs 3127120, 60210
trace: GC (incremental) at mem/objs 3131232, 60310
trace:  down to mem/objs 3123152, 60110
trace: GC (incremental) at mem/objs 3127152, 60210
trace:  down to mem/objs 3119952, 60010
trace: GC (incremental) at mem/objs 3124336, 60110
trace:  down to mem/objs 3117328, 59910
trace: GC (incremental) at mem/objs 3121648, 60010
trace:  down to mem/objs 3114144, 59810
trace: GC (incremental) at mem/objs 3118496, 59910
trace:  down to mem/objs 3111856, 59710
trace: GC (incremental) at mem/objs 3116208, 59810
trace:  down to mem/objs 3109312, 59610
trace: GC (incremental) at mem/objs 3113680, 59710
trace:  down to mem/objs 3106784, 59510
trace: GC (incremental) at mem/objs 3111104, 59610
trace:  down to mem/objs 3104304, 59410
trace: GC (incremental) at mem/objs 3108640, 59510
trace:  down to mem/objs 3101952, 59310
trace: GC (incremental) at mem/objs 3106256, 59410
trace:  down to mem/objs 3099360, 59210
trace: GC (incremental) at mem/objs 3103328, 59310
trace:  down to mem/objs 3096224, 59110
trace: GC (incremental) at mem/objs 3100160, 59210
trace:  down to mem/objs 3092992, 59010
trace: GC (incremental) at mem/objs 3096896, 59110
trace:  down to mem/objs 3089840, 58910
trace: GC (incremental) at mem/objs 3093840, 59010
trace:  down to mem/objs (sweep complete) 3091328, 58934
trace: GC (incremental) at mem/objs 5537184, 117868
trace:  down to mem/objs 5529984, 117668
trace: GC (incremental) at mem/objs 5534000, 117768
trace:  down to mem/objs 5527312, 117568
trace: GC (incremental) at mem/objs 5531520, 117668
trace:  down to mem/objs 5524832, 117468
trace: GC (incremental) at mem/objs 5528976, 117568
trace:  down to mem/objs 5522080, 117368
trace: GC (incremental) at mem/objs 5526128, 117468
trace:  down to mem/objs 5519136, 117268
trace: GC (incremental) at mem/objs 5523152, 117368
trace:  down to mem/objs 5515984, 117168
trace: GC (incremental) at mem/objs 5520048, 117268
trace:  down to mem/objs 5513648, 117068
trace: GC (incremental) at mem/objs 5517824, 117168
trace:  down to mem/objs 5511056, 116968
trace: GC (incremental) at mem/objs 5515328, 117068
trace:  down to mem/objs 5508544, 116868
trace: GC (incremental) at mem/objs 5512880, 116968
trace:  down to mem/objs 5505760, 116768
trace: GC (incremental) at mem/objs 5509744, 116868
trace:  down to mem/objs 5503152, 116668
trace: GC (incremental) at mem/objs 5507264, 116768
trace:  down to mem/objs 5500368, 116568
trace: GC (incremental) at mem/objs 5504592, 116668
trace:  down to mem/objs 5498032, 116468
trace: GC (incremental) at mem/objs 5502336, 116568
trace:  down to mem/objs 5494928, 116368
trace: GC (incremental) at mem/objs 5499072, 116468
trace:  down to mem/objs 5492592, 116268
trace: GC (incremental) at mem/objs 5496704, 116368
trace:  down to mem/objs 5490192, 116168
trace: GC (incremental) at mem/objs 5494432, 116268
trace:  down to mem/objs 5487552, 116068
trace: GC (incremental) at mem/objs 5491680, 116168
trace:  down to mem/objs 5485184, 115968
trace: GC (incremental) at mem/objs 5489216, 116068
trace:  down to mem/objs 5482368, 115868
trace: GC (incremental) at mem/objs 5486448, 115968
trace:  down to mem/objs 5479744, 115768
trace: GC (incremental) at mem/objs 5483792, 115868
trace:  down to mem/objs 5477008, 115668
trace: GC (incremental) at mem/objs 5481056, 115768
trace:  down to mem/objs 5474048, 115568
trace: GC (incremental) at mem/objs 5477952, 115668
trace:  down to mem/objs 5471264, 115468
trace: GC (incremental) at mem/objs 5475360, 115568
trace:  down to mem/objs 5468320, 115368
trace: GC (incremental) at mem/objs 5472320, 115468
trace:  down to mem/objs 5465808, 115268
trace: GC (incremental) at mem/objs 5469968, 115368
trace:  down to mem/objs 5463456, 115168
trace: GC (incremental) at mem/objs 5467696, 115268
trace:  down to mem/objs 5460672, 115068
trace: GC (incremental) at mem/objs 5464864, 115168
trace:  down to mem/objs 5458336, 114968
trace: GC (incremental) at mem/objs 5462480, 115068
trace:  down to mem/objs 5455472, 114868
trace: GC (incremental) at mem/objs 5459472, 114968
trace:  down to mem/objs 5452896, 114768
trace: GC (incremental) at mem/objs 5456992, 114868
trace:  down to mem/objs 5450032, 114668
trace: GC (incremental) at mem/objs 5454144, 114768
trace:  down to mem/objs 5447536, 114568
trace: GC (incremental) at mem/objs 5451552, 114668
trace:  down to mem/objs 5444608, 114468
trace: GC (incremental) at mem/objs 5448416, 114568
trace:  down to mem/objs 5441888, 114368
trace: GC (incremental) at mem/objs 5445808, 114468
trace:  down to mem/objs 5438848, 114268
trace: GC (incremental) at mem/objs 5442896, 114368
trace:  down to mem/objs 5436368, 114168
trace: GC (incremental) at mem/objs 5440448, 114268
trace:  down to mem/objs 5433552, 114068
trace: GC (incremental) at mem/objs 5437648, 114168
trace:  down to mem/objs 5430336, 113968
trace: GC (incremental) at mem/objs 5434432, 114068
trace:  down to mem/objs 5427824, 113868
trace: GC (incremental) at mem/objs 5431856, 113968
trace:  down to mem/objs 5425408, 113768
trace: GC (incremental) at mem/objs 5429744, 113868
trace:  down to mem/objs 5423136, 113668
trace: GC (incremental) at mem/objs 5427408, 113768
trace:  down to mem/objs 5420736, 113568
trace: GC (incremental) at mem/objs 5424816, 113668
trace:  down to mem/objs 5418256, 113468
trace: GC (incremental) at mem/objs 5422288, 113568
trace:  down to mem/objs 5415680, 113368
trace: GC (incremental) at mem/objs 5419840, 113468
trace:  down to mem/objs 5413376, 113268
trace: GC (incremental) at mem/objs 5417520, 113368
trace:  down to mem/objs 5410896, 113168
trace: GC (incremental) at mem/objs 5414912, 113268
trace:  down to mem/objs 5407856, 113068
trace: GC (incremental) at mem/objs 5412064, 113168
trace:  down to mem/objs 5405376, 112968
trace: GC (incremental) at mem/objs 5409488, 113068
trace:  down to mem/objs 5402752, 112868
trace: GC (incremental) at mem/objs 5406800, 112968
trace:  down to mem/objs 5400048, 112768
trace: GC (incremental) at mem/objs 5404144, 112868
trace:  down to mem/objs 5397408, 112668
trace: GC (incremental) at mem/objs 5401424, 112768
trace:  down to mem/objs 5394944, 112568
trace: GC (incremental) at mem/objs 5398976, 112668
trace:  down to mem/objs 5392272, 112468
trace: GC (incremental) at mem/objs 5396464, 112568
trace:  down to mem/objs 5389920, 112368
trace: GC (incremental) at mem/objs 5394096, 112468
trace:  down to mem/objs 5387488, 112268
trace: GC (incremental) at mem/objs 5391488, 112368
trace:  down to mem/objs 5384832, 112168
trace: GC (incremental) at mem/objs 5388720, 112268
trace:  down to mem/objs 5381568, 112068
trace: GC (incremental) at mem/objs 5385504, 112168
trace:  down to mem/objs 5378464, 111968
trace: GC (incremental) at mem/objs 5382368, 112068
trace:  down to mem/objs 5375296, 111868
trace: GC (incremental) at mem/objs 5379280, 111968
trace:  down to mem/objs 5372112, 111768
trace: GC (incremental) at mem/objs 5376112, 111868
trace:  down to mem/objs 5369136, 111668
trace: GC (incremental) at mem/objs 5373328, 111768
trace:  down to mem/objs 5366384, 111568
trace: GC (incremental) at mem/objs 5370544, 111668
trace:  down to mem/objs 5363936, 111468
trace: GC (incremental) at mem/objs 5367968, 111568
trace:  down to mem/objs 5361136, 111368
trace: GC (incremental) at mem/objs 5365040, 111468
trace:  down to mem/objs 5358400, 111268
trace: GC (incremental) at mem/objs 5362320, 111368
trace:  down to mem/objs 5355280, 111168
trace: GC (incremental) at mem/objs 5359216, 111268
trace:  down to mem/objs 5352752, 111068
trace: GC (incremental) at mem/objs 5356656, 111168
trace:  down to mem/objs 5349792, 110968
trace: GC (incremental) at mem/objs 5353744, 111068
trace:  down to mem/objs 5347232, 110868
  ~lib/memory.ts
trace: GC (incremental) at mem/objs 5357280, 110968
trace:  down to mem/objs 5350576, 110768
trace: GC (incremental) at mem/objs 5354816, 110868
trace:  down to mem/objs 5347936, 110668
trace: GC (incremental) at mem/objs 5352272, 110768
trace:  down to mem/objs 5345776, 110568
trace: GC (incremental) at mem/objs 5349904, 110668
trace:  down to mem/objs 5342912, 110468
trace: GC (incremental) at mem/objs 5347168, 110568
trace:  down to mem/objs 5340688, 110368
trace: GC (incremental) at mem/objs 5345136, 110468
trace:  down to mem/objs 5338192, 110268
trace: GC (incremental) at mem/objs 5342432, 110368
trace:  down to mem/objs 5335936, 110168
trace: GC (incremental) at mem/objs 5340304, 110268
trace:  down to mem/objs 5333616, 110068
trace: GC (incremental) at mem/objs 5337648, 110168
trace:  down to mem/objs 5330944, 109968
trace: GC (incremental) at mem/objs 5335152, 110068
trace:  down to mem/objs 5328240, 109868
trace: GC (incremental) at mem/objs 5332416, 109968
trace:  down to mem/objs 5325472, 109768
trace: GC (incremental) at mem/objs 5329792, 109868
trace:  down to mem/objs 5323232, 109668
trace: GC (incremental) at mem/objs 5327408, 109768
trace:  down to mem/objs 5320512, 109568
trace: GC (incremental) at mem/objs 5324688, 109668
trace:  down to mem/objs 5318176, 109468
  ~lib/number.ts
trace: GC (incremental) at mem/objs 5339200, 109568
trace:  down to mem/objs 5332352, 109368
trace: GC (incremental) at mem/objs 5336608, 109468
trace:  down to mem/objs 5329984, 109268
trace: GC (incremental) at mem/objs 5334400, 109368
trace:  down to mem/objs 5327584, 109168
trace: GC (incremental) at mem/objs 5331840, 109268
trace:  down to mem/objs 5325168, 109068
trace: GC (incremental) at mem/objs 5329376, 109168
trace:  down to mem/objs 5322320, 108968
trace: GC (incremental) at mem/objs 5326560, 109068
trace:  down to mem/objs 5319968, 108868
trace: GC (incremental) at mem/objs 5324192, 108968
trace:  down to mem/objs 5317392, 108768
trace: GC (incremental) at mem/objs 5321600, 108868
trace:  down to mem/objs (sweep complete) 5315536, 108685
  ~lib/object.ts
  ~lib/polyfills.ts
  ~lib/reference.ts
  ~lib/regexp.ts
  ~lib/rt.ts
  ~lib/set.ts
  ~lib/staticarray.ts
  ~lib/string.ts
  ~lib/symbol.ts
  ~lib/table.ts
  ~lib/typedarray.ts
  ~lib/vector.ts

Parsing runtime ...
  ~lib/rt/index-noexport.ts

Parsing backlog ...
  ~lib/rt/common.ts
  ~lib/util/sort.ts
  ~lib/util/string.ts
trace: GC (incremental) at mem/objs 10258496, 217370
trace:  down to mem/objs 10251392, 217170
trace: GC (incremental) at mem/objs 10256384, 217270
trace:  down to mem/objs 10249536, 217070
trace: GC (incremental) at mem/objs 10254544, 217170
trace:  down to mem/objs 10247952, 216970
trace: GC (incremental) at mem/objs 10252944, 217070
trace:  down to mem/objs 10246304, 216870
trace: GC (incremental) at mem/objs 10251312, 216970
trace:  down to mem/objs 10244720, 216770
trace: GC (incremental) at mem/objs 10249712, 216870
trace:  down to mem/objs 10243264, 216670
trace: GC (incremental) at mem/objs 10248272, 216770
trace:  down to mem/objs 10241600, 216570
trace: GC (incremental) at mem/objs 10246592, 216670
trace:  down to mem/objs 10240080, 216470
trace: GC (incremental) at mem/objs 10245088, 216570
trace:  down to mem/objs 10238528, 216370
trace: GC (incremental) at mem/objs 10243520, 216470
trace:  down to mem/objs 10236960, 216270
trace: GC (incremental) at mem/objs 10241968, 216370
trace:  down to mem/objs 10235456, 216170
trace: GC (incremental) at mem/objs 10240448, 216270
trace:  down to mem/objs 10233008, 216070
trace: GC (incremental) at mem/objs 10238016, 216170
trace:  down to mem/objs 10230720, 215970
trace: GC (incremental) at mem/objs 10235712, 216070
trace:  down to mem/objs 10228352, 215870
trace: GC (incremental) at mem/objs 10233360, 215970
trace:  down to mem/objs 10226416, 215770
trace: GC (incremental) at mem/objs 10231408, 215870
trace:  down to mem/objs 10224464, 215670
trace: GC (incremental) at mem/objs 10229472, 215770
trace:  down to mem/objs 10222528, 215570
trace: GC (incremental) at mem/objs 10227520, 215670
trace:  down to mem/objs 10220544, 215470
trace: GC (incremental) at mem/objs 10225552, 215570
trace:  down to mem/objs 10218640, 215370
trace: GC (incremental) at mem/objs 10223632, 215470
trace:  down to mem/objs 10216880, 215270
trace: GC (incremental) at mem/objs 10221888, 215370
trace:  down to mem/objs 10215120, 215170
trace: GC (incremental) at mem/objs 10220112, 215270
trace:  down to mem/objs 10213056, 215070
trace: GC (incremental) at mem/objs 10218064, 215170
trace:  down to mem/objs 10211488, 214970
trace: GC (incremental) at mem/objs 10216480, 215070
trace:  down to mem/objs 10208656, 214870
trace: GC (incremental) at mem/objs 10213664, 214970
trace:  down to mem/objs 10206512, 214770
trace: GC (incremental) at mem/objs 10211504, 214870
trace:  down to mem/objs 10204544, 214670
trace: GC (incremental) at mem/objs 10209552, 214770
trace:  down to mem/objs 10202432, 214570
trace: GC (incremental) at mem/objs 10207424, 214670
trace:  down to mem/objs 10200336, 214470
trace: GC (incremental) at mem/objs 10205344, 214570
trace:  down to mem/objs 10197840, 214370
trace: GC (incremental) at mem/objs 10202832, 214470
trace:  down to mem/objs 10196128, 214270
trace: GC (incremental) at mem/objs 10201136, 214370
trace:  down to mem/objs 10194320, 214170
trace: GC (incremental) at mem/objs 10199312, 214270
trace:  down to mem/objs 10192400, 214070
trace: GC (incremental) at mem/objs 10197408, 214170
trace:  down to mem/objs 10190512, 213970
trace: GC (incremental) at mem/objs 10195504, 214070
trace:  down to mem/objs 10188576, 213870
trace: GC (incremental) at mem/objs 10193584, 213970
trace:  down to mem/objs 10186672, 213770
trace: GC (incremental) at mem/objs 10191664, 213870
trace:  down to mem/objs 10184688, 213670
trace: GC (incremental) at mem/objs 10189696, 213770
trace:  down to mem/objs 10182960, 213570
trace: GC (incremental) at mem/objs 10187952, 213670
trace:  down to mem/objs 10180896, 213470
trace: GC (incremental) at mem/objs 10185904, 213570
trace:  down to mem/objs 10178880, 213370
trace: GC (incremental) at mem/objs 10183872, 213470
trace:  down to mem/objs 10176800, 213270
trace: GC (incremental) at mem/objs 10181808, 213370
trace:  down to mem/objs 10175056, 213170
trace: GC (incremental) at mem/objs 10180048, 213270
trace:  down to mem/objs 10173072, 213070
trace: GC (incremental) at mem/objs 10178080, 213170
trace:  down to mem/objs 10171376, 212970
trace: GC (incremental) at mem/objs 10176368, 213070
trace:  down to mem/objs 10169472, 212870
trace: GC (incremental) at mem/objs 10174480, 212970
trace:  down to mem/objs 10167312, 212770
trace: GC (incremental) at mem/objs 10172304, 212870
trace:  down to mem/objs 10165376, 212670
trace: GC (incremental) at mem/objs 10170384, 212770
trace:  down to mem/objs 10163584, 212570
trace: GC (incremental) at mem/objs 10168576, 212670
trace:  down to mem/objs 10161808, 212470
trace: GC (incremental) at mem/objs 10166816, 212570
trace:  down to mem/objs 10159904, 212370
trace: GC (incremental) at mem/objs 10164896, 212470
trace:  down to mem/objs 10158096, 212270
trace: GC (incremental) at mem/objs 10163104, 212370
trace:  down to mem/objs 10156256, 212170
trace: GC (incremental) at mem/objs 10161248, 212270
trace:  down to mem/objs 10154224, 212070
trace: GC (incremental) at mem/objs 10159232, 212170
trace:  down to mem/objs 10152608, 211970
trace: GC (incremental) at mem/objs 10157600, 212070
trace:  down to mem/objs 10150592, 211870
trace: GC (incremental) at mem/objs 10155600, 211970
trace:  down to mem/objs 10148784, 211770
trace: GC (incremental) at mem/objs 10153776, 211870
trace:  down to mem/objs 10146976, 211670
trace: GC (incremental) at mem/objs 10151984, 211770
trace:  down to mem/objs 10145088, 211570
trace: GC (incremental) at mem/objs 10150080, 211670
trace:  down to mem/objs 10143088, 211470
trace: GC (incremental) at mem/objs 10148096, 211570
trace:  down to mem/objs 10141296, 211370
trace: GC (incremental) at mem/objs 10146288, 211470
trace:  down to mem/objs 10139168, 211270
trace: GC (incremental) at mem/objs 10144176, 211370
trace:  down to mem/objs 10137232, 211170
trace: GC (incremental) at mem/objs 10142224, 211270
trace:  down to mem/objs 10135360, 211070
trace: GC (incremental) at mem/objs 10140368, 211170
trace:  down to mem/objs 10133632, 210970
trace: GC (incremental) at mem/objs 10138624, 211070
trace:  down to mem/objs 10131856, 210870
trace: GC (incremental) at mem/objs 10136864, 210970
trace:  down to mem/objs 10130208, 210770
trace: GC (incremental) at mem/objs 10135200, 210870
trace:  down to mem/objs 10128240, 210670
trace: GC (incremental) at mem/objs 10133248, 210770
trace:  down to mem/objs 10126352, 210570
trace: GC (incremental) at mem/objs 10131344, 210670
trace:  down to mem/objs 10124704, 210470
trace: GC (incremental) at mem/objs 10129712, 210570
trace:  down to mem/objs 10122864, 210370
trace: GC (incremental) at mem/objs 10127856, 210470
trace:  down to mem/objs 10121184, 210270
trace: GC (incremental) at mem/objs 10126192, 210370
trace:  down to mem/objs 10119376, 210170
trace: GC (incremental) at mem/objs 10124368, 210270
trace:  down to mem/objs 10117632, 210070
trace: GC (incremental) at mem/objs 10122640, 210170
trace:  down to mem/objs 10115888, 209970
trace: GC (incremental) at mem/objs 10120880, 210070
trace:  down to mem/objs 10113824, 209870
trace: GC (incremental) at mem/objs 10118832, 209970
trace:  down to mem/objs 10111808, 209770
trace: GC (incremental) at mem/objs 10116800, 209870
trace:  down to mem/objs 10109904, 209670
trace: GC (incremental) at mem/objs 10114736, 209770
trace:  down to mem/objs 10108096, 209570
trace: GC (incremental) at mem/objs 10112752, 209670
trace:  down to mem/objs 10105632, 209470
trace: GC (incremental) at mem/objs 10110624, 209570
trace:  down to mem/objs 10103472, 209370
trace: GC (incremental) at mem/objs 10108480, 209470
trace:  down to mem/objs 10101264, 209270
trace: GC (incremental) at mem/objs 10106256, 209370
trace:  down to mem/objs 10099504, 209170
trace: GC (incremental) at mem/objs 10104512, 209270
trace:  down to mem/objs 10097696, 209070
trace: GC (incremental) at mem/objs 10102688, 209170
trace:  down to mem/objs 10095200, 208970
trace: GC (incremental) at mem/objs 10100208, 209070
trace:  down to mem/objs 10093008, 208870
trace: GC (incremental) at mem/objs 10098000, 208970
trace:  down to mem/objs 10090800, 208770
trace: GC (incremental) at mem/objs 10095808, 208870
trace:  down to mem/objs 10088928, 208670
trace: GC (incremental) at mem/objs 10093920, 208770
trace:  down to mem/objs 10087136, 208570
trace: GC (incremental) at mem/objs 10092144, 208670
trace:  down to mem/objs 10085024, 208470
trace: GC (incremental) at mem/objs 10090016, 208570
trace:  down to mem/objs 10082800, 208370
trace: GC (incremental) at mem/objs 10087808, 208470
trace:  down to mem/objs 10080720, 208270
trace: GC (incremental) at mem/objs 10085712, 208370
trace:  down to mem/objs 10078896, 208170
trace: GC (incremental) at mem/objs 10083904, 208270
trace:  down to mem/objs 10076944, 208070
trace: GC (incremental) at mem/objs 10081936, 208170
trace:  down to mem/objs 10074864, 207970
trace: GC (incremental) at mem/objs 10079872, 208070
trace:  down to mem/objs 10072736, 207870
trace: GC (incremental) at mem/objs 10077728, 207970
trace:  down to mem/objs 10070672, 207770
trace: GC (incremental) at mem/objs 10075680, 207870
trace:  down to mem/objs 10068992, 207670
trace: GC (incremental) at mem/objs 10073984, 207770
trace:  down to mem/objs 10066960, 207570
trace: GC (incremental) at mem/objs 10071968, 207670
trace:  down to mem/objs 10064832, 207470
trace: GC (incremental) at mem/objs 10069824, 207570
trace:  down to mem/objs 10062608, 207370
trace: GC (incremental) at mem/objs 10067616, 207470
trace:  down to mem/objs 10060432, 207270
trace: GC (incremental) at mem/objs 10065424, 207370
trace:  down to mem/objs 10058864, 207170
trace: GC (incremental) at mem/objs 10063872, 207270
trace:  down to mem/objs 10056784, 207070
trace: GC (incremental) at mem/objs 10061776, 207170
trace:  down to mem/objs 10054592, 206970
trace: GC (incremental) at mem/objs 10059600, 207070
trace:  down to mem/objs 10052464, 206870
trace: GC (incremental) at mem/objs 10057456, 206970
trace:  down to mem/objs 10050304, 206770
trace: GC (incremental) at mem/objs 10055312, 206870
trace:  down to mem/objs 10048544, 206670
trace: GC (incremental) at mem/objs 10053536, 206770
trace:  down to mem/objs 10046448, 206570
trace: GC (incremental) at mem/objs 10051456, 206670
trace:  down to mem/objs 10044192, 206470
trace: GC (incremental) at mem/objs 10048976, 206570
trace:  down to mem/objs 10041856, 206370
trace: GC (incremental) at mem/objs 10046592, 206470
trace:  down to mem/objs 10039712, 206270
trace: GC (incremental) at mem/objs 10044720, 206370
trace:  down to mem/objs 10037856, 206170
trace: GC (incremental) at mem/objs 10042848, 206270
trace:  down to mem/objs 10035760, 206070
trace: GC (incremental) at mem/objs 10040768, 206170
trace:  down to mem/objs 10033504, 205970
trace: GC (incremental) at mem/objs 10038496, 206070
trace:  down to mem/objs 10031408, 205870
trace: GC (incremental) at mem/objs 10036416, 205970
trace:  down to mem/objs 10029600, 205770
trace: GC (incremental) at mem/objs 10034592, 205870
trace:  down to mem/objs 10027632, 205670
trace: GC (incremental) at mem/objs 10032640, 205770
trace:  down to mem/objs 10025568, 205570
trace: GC (incremental) at mem/objs 10030560, 205670
trace:  down to mem/objs 10023312, 205470
trace: GC (incremental) at mem/objs 10028320, 205570
trace:  down to mem/objs 10021168, 205370
trace: GC (incremental) at mem/objs 10026160, 205470
trace:  down to mem/objs 10019472, 205270
trace: GC (incremental) at mem/objs 10024480, 205370
trace:  down to mem/objs 10017488, 205170
trace: GC (incremental) at mem/objs 10022480, 205270
trace:  down to mem/objs 10015312, 205070
trace: GC (incremental) at mem/objs 10020320, 205170
trace:  down to mem/objs 10013104, 204970
trace: GC (incremental) at mem/objs 10018096, 205070
trace:  down to mem/objs 10010912, 204870
trace: GC (incremental) at mem/objs 10015920, 204970
trace:  down to mem/objs 10009360, 204770
trace: GC (incremental) at mem/objs 10014352, 204870
trace:  down to mem/objs 10007280, 204670
trace: GC (incremental) at mem/objs 10012288, 204770
trace:  down to mem/objs 10005088, 204570
trace: GC (incremental) at mem/objs 10010080, 204670
trace:  down to mem/objs 10002944, 204470
trace: GC (incremental) at mem/objs 10007952, 204570
trace:  down to mem/objs 10000800, 204370
trace: GC (incremental) at mem/objs 10005792, 204470
trace:  down to mem/objs 9999024, 204270
trace: GC (incremental) at mem/objs 10004032, 204370
trace:  down to mem/objs 9996944, 204170
trace: GC (incremental) at mem/objs 10001936, 204270
trace:  down to mem/objs 9994672, 204070
trace: GC (incremental) at mem/objs 9999680, 204170
trace:  down to mem/objs 9992560, 203970
trace: GC (incremental) at mem/objs 9997552, 204070
trace:  down to mem/objs 9990672, 203870
trace: GC (incremental) at mem/objs 9995680, 203970
trace:  down to mem/objs 9988816, 203770
trace: GC (incremental) at mem/objs 9993808, 203870
trace:  down to mem/objs 9986816, 203670
trace: GC (incremental) at mem/objs 9991824, 203770
trace:  down to mem/objs 9984832, 203570
trace: GC (incremental) at mem/objs 9989824, 203670
trace:  down to mem/objs 9982784, 203470
trace: GC (incremental) at mem/objs 9987792, 203570
trace:  down to mem/objs 9980784, 203370
trace: GC (incremental) at mem/objs 9985776, 203470
trace:  down to mem/objs 9978784, 203270
trace: GC (incremental) at mem/objs 9983792, 203370
trace:  down to mem/objs 9976576, 203170
trace: GC (incremental) at mem/objs 9981568, 203270
trace:  down to mem/objs 9974400, 203070
trace: GC (incremental) at mem/objs 9979408, 203170
trace:  down to mem/objs 9972416, 202970
trace: GC (incremental) at mem/objs 9977408, 203070
trace:  down to mem/objs 9970128, 202870
trace: GC (incremental) at mem/objs 9975136, 202970
trace:  down to mem/objs 9968240, 202770
trace: GC (incremental) at mem/objs 9973232, 202870
trace:  down to mem/objs 9966160, 202670
trace: GC (incremental) at mem/objs 9971168, 202770
trace:  down to mem/objs 9963984, 202570
trace: GC (incremental) at mem/objs 9968976, 202670
trace:  down to mem/objs 9961936, 202470
trace: GC (incremental) at mem/objs 9966944, 202570
trace:  down to mem/objs 9959952, 202370
trace: GC (incremental) at mem/objs 9964944, 202470
trace:  down to mem/objs 9957872, 202270
trace: GC (incremental) at mem/objs 9962880, 202370
trace:  down to mem/objs 9956144, 202170
trace: GC (incremental) at mem/objs 9961136, 202270
trace:  down to mem/objs 9954048, 202070
trace: GC (incremental) at mem/objs 9959056, 202170
trace:  down to mem/objs 9952000, 201970
trace: GC (incremental) at mem/objs 9956992, 202070
trace:  down to mem/objs 9950080, 201870
trace: GC (incremental) at mem/objs 9955088, 201970
trace:  down to mem/objs 9948096, 201770
trace: GC (incremental) at mem/objs 9953088, 201870
trace:  down to mem/objs 9945824, 201670
trace: GC (incremental) at mem/objs 9950832, 201770
trace:  down to mem/objs 9944208, 201570
trace: GC (incremental) at mem/objs 9949200, 201670
trace:  down to mem/objs 9942640, 201470
trace: GC (incremental) at mem/objs 9947648, 201570
trace:  down to mem/objs (sweep complete) 9944752, 201484
  ~lib/builtins.ts
  ~lib/util/error.ts
  ~lib/rt/common.ts
  ~lib/builtins.ts
  ~lib/util/error.ts
  ~lib/util/error.ts
  ~lib/rt/common.ts
  ~lib/util/error.ts
  ~lib/bindings/Date.ts
  ~lib/util/hash.ts
  ~lib/util/error.ts
  ~lib/bindings/Math.ts
  ~lib/util/math.ts
  ~lib/util/memory.ts
  ~lib/util/error.ts
  ~lib/util/number.ts
  ~lib/util/string.ts
  ~lib/shared/typeinfo.ts
  ~lib/util/error.ts
  ~lib/rt/common.ts
  ~lib/util/hash.ts
  ~lib/rt/common.ts
  ~lib/util/error.ts
  ~lib/util/string.ts
  ~lib/rt/common.ts
  ~lib/util/string.ts
  ~lib/util/casemap.ts
  ~lib/util/error.ts
  ~lib/util/error.ts
  ~lib/util/sort.ts
  ~lib/util/error.ts
  ~lib/util/string.ts
  ~lib/rt/tlsf.ts
  ~lib/rt/tcms.ts
  ~lib/util/string.ts
  ~lib/util/number.ts
  ~lib/rt/rtrace.ts
  ~lib/rt/rtrace.ts

Parsing entry file ...
  index.ts

Parsing backlog ...

Initializing program ...

Compiling program ...
(module
 (type $none_=>_none (func))
 (type $i32_i32_=>_i32 (func (param i32 i32) (result i32)))
 (memory $0 0)
 (table $0 1 funcref)
 (export "memory" (memory $0))
 (export "add" (func $index/add))
 (start $~start)
 (func $start:~lib/rt/rtrace
  (drop
   (i32.const 0)
  )
 )
 (func $start:~lib/rt/tlsf
  (call $start:~lib/rt/rtrace)
 )
 (func $start:~lib/rt/index-noexport
  (call $start:~lib/rt/tlsf)
 )
 (func $index/add (param $0 i32) (param $1 i32) (result i32)
  (i32.add
   (local.get $0)
   (local.get $1)
  )
 )
 (func $~start
  (call $start:~lib/rt/index-noexport)
 )
)


Checking diagnostics ...
Errors: 0
trace: GC (full) at mem/objs 16928448, 340916
trace: GC (full) done at mem/objs 11840, 43

Collect took 26 ms
```
</details>

dcodeIO:
To not inflict too much hope here: Adjusting the GC parameters can still lead to what seems to be memory corruption or bugs in the incremental mechanism at some places, and I have concerns with a few side products like that users again have to retain/release everything externally because GC may kick in at any time in an incremental GC. Feels a bit like we'll get the cons of both reference counting and tracing as soon as this gets incremental. Will have more details to report at today's meeting.

dcodeIO:
Last commit makes it so that the GC can be invoked either to perform a full collection with `__collect()`, or to perform incremental steps with `__collect(true)`. The bootstrap test passes in both modes, but there is no automation on exports yet. Experimented with automation on the boundary, but wasn't quite happy with it, so figured I'd need some more time to think about that.

MaxGraey:
Could you sync this PR with master?

dcodeIO:
So, here's something cool: `asc --wasm` loads the AS compiler compiled to Wasm. Must be built with `npm run asbuild` first. `npm run test:compiler X -- --wasm` runs an existing test case with it (doesn't behave well with `--parallel`, so avoid). Some tests work, others don't. Fix one of them to attain eternal glory.

dcodeIO:
At this point it passes all the tests, generating the exact same code the compiler compiled to JS produces, except that if one runs multiple tests in sequence there's still something surprising going on. I'm surprised.

dcodeIO:
One error is that in `std/math` the fixture reads

```diff
-  f32.const -7.888609052210118e-31
   f32.const -0
+  f32.const -0
```

```diff
   f64.const -8.900295434028806e-308
-  f64.const -8.900295434028806e-308
+  f64.const -0
```

```diff
   f64.const 2.225073858507201e-308
-  f64.const 2.225073858507201e-308
   f64.const 0
+  f64.const 0
```

in certain test run situations. Wondering if that has something to do with float parsing, and not quite sure where these constants are in the `std/math` test. cc @MaxGraey - ideas?

MaxGraey:
Yeah, It may be problem with float parse / stringify. It's not fully precise in some situations and may have an error in 2-3 ULPs in worst case

dcodeIO:
Btw, using an optimized assemblyscript.wasm, I am seeing faster compile times compared to JS as expected.

Testcase | JS | Wasm
----------|----|-------
std/array | 1360.582 ms | 921.210 ms
std/math | 1549.867 ms | 1374.688 ms
std/typedarray | 551.416 ms | 412.052 ms

Times for Wasm include a final full GC that's technically unnecessary when terminating right away.

dcodeIO:
Last commit replaces the old makeshift bootstrap test with the compiler compiled to Wasm running the *entire* test suite. It passes all tests individually now, but runs into issues eventually when reusing the same compiler instance for all the tests in succession with GCs in between. Made it instantiate a new compiler instance on each invocation for now, which is already great to have in CI and much more than I expected :)

dcodeIO:
Affected test cases are

* function-expression
* infer-generic

both doing something with function expressions, apparently corrupting state. All other tests in succession, with these two disabled, work without reusing a fresh compiler instance.

dcodeIO:
Not quite sure why it's still failing on CI but working for me locally. Might have something to do with different test order or so, and indicates that there's still a runtime bug somewhere. Most likely some sort of memory corruption or use-after-free, as the error happens when `__visit_members`ing an object with an invalid class id in what looks like a propagating effect from an earlier run.

dcodeIO:
Interesting: Running the test suite with an untouched build worked in the last commit, but failed with an optimized build.

dcodeIO:
Last commit fixes a related set of memory management issues:
1. Compiler-generated static memory segments had invalid `mmInfo`, not accounting for the runtime header and not properly aligning segment sizes so TLSF flag bits are never set
2. TCMS `__renew` is a kinda unwieldy beast due to having to account for various edge cases (that were previously missing) when reallocating (from, and possibly to) static memory, which `__realloc` already handles, but doesn't know about GC so requires extra branches in `__renew` as well. Makes me question whether the ability to realloc GC objects is actually worth it, but I also don't see a good alternative yet.

trusktr:
This is neat:

Screenshot from 2020-12-17 19-35-32

Do you generate the graphic, or use a tool? Or by hand?

dcodeIO:
By hand, but once proposed [a solution](https://twitter.com/dcodeIO/status/1301205891035934720). If not hardware, perhaps a VSCode extension adding a toolbar to insert these into the document :)

trusktr:
Vim might have everything you need. Check out this video

https://youtu.be/GDa7hrbcCB8

At 4:12, there is a tool called toilet that generates a box around some text, and he pipes it into the text buffer within vim.

I can't find a video, but the Vim `DrawIt` plugin is nifty! Here is an article with good enough description (better than DrawIt's own description). But it doesn't cover everything. You just have to try it!

http://www.drchip.org/astronaut/vim/index.html#DRAWIT

dcodeIO:
Did a bit more testing out of curiosity to get a feel for what kind of performance we are talking when comparing the AS compiler compiled to JS with the AS compiler compiled to Wasm, compiling itself. I used `--measure` as point of reference, comparing the JS compiler with the Wasm compiler using -O3:

* I/O is a bit slower, which is expected due to having to copy all sources into Wasm memory. Adds about 20ms overall.
* Parsing the standard library and the compiler sources goes down from 452.095 ms to 178.869 ms (~2.5x).
* Initializing the program after parsing goes down from 36.376 ms to 9.105 ms (~3.9x)
* Compiling the initialized program goes down from 1302.123 ms to 778.243 ms (~1.6x)
* Final TSMS GC takes 46.153 ms, collecting ~104mb of garbage

dcodeIO:
Over the course of the entire test suite, which is doing lots of other things like diffing, running rtrace, optimizing with Binaryen and whatnot, the picture is

* 84756 ms JS
* 95519 ms Wasm debug build without optimizations, expected
* 79042 ms Wasm release build with -O3

The ~6.7 % improvement isn't exciting but seems reasonable given the sheer scope of tasks the test suite performs.

dcodeIO:
Speaking of things that are embarrassingly simple, recent commits strip down the GC to what's absolutely necessary, eliminating the need for write barriers, the incremental state machine and branching initialization. Classic two-color mark & sweep with space/color flipping ¯\\_(ツ)_/¯. Won't win a price for sophistication obviously, but it doesn't has to as long as it is the leanest and meanest of solutions I guess.

dcodeIO:
Quick sanity check after getting things pass again:

```
// JS
I/O Read   :    17.285 ms  n=95
I/O Write  :    56.713 ms  n=3
Parse      :   446.875 ms  n=150
Initialize :    32.999 ms  n=1
Compile    :  1367.154 ms  n=1
Emit       :  1864.109 ms  n=3
Validate   :   253.162 ms  n=1
Optimize   : 41177.931 ms  n=1
Transform  :          n/a  n=0

// Wasm
I/O Read   :    24.875 ms  n=95   | I/O
I/O Write  :    64.239 ms  n=3    |
Parse      :   162.742 ms  n=150  <---
Initialize :     9.681 ms  n=1    <--- actually the compiler
Compile    :   817.929 ms  n=1    <---
Emit       :  1930.280 ms  n=3    |
Validate   :   260.536 ms  n=1    |
Optimize   : 46204.186 ms  n=1    | Binaryen
Transform  :          n/a  n=0
```

dcodeIO:
Last commit removes retain/release for good and replaces it with a slightly simpler pin/unpin mechanism that doesn't refcount (can pin exactly once, unpin only if pinned). This way around the module can provide all the helpers necessary for all kinds of embedders (JS, Rust, ...) due to not having to maintain a potentially largish (in code size) reference count map. It's more general in a way, since retain/release can be implemented on top of it if so desired.

MaxGraey:
I guess after [this line](https://github.com/AssemblyScript/assemblyscript/blob/eb3aaabc355cd7a6b3c1fbf0f1878cb64d14ebb3/std/assembly/rt/tlsf.ts#L444) make sense insert `__collect` call. wdyt?

dcodeIO:
Memory may grow when already arbitrarily deep into the stack, so putting a `__collect` there is likely to prematurely collect objects that are only being kept alive by the stack. That's about why we'd need a shadow stack, and currently only call `__collect` externally when the stack is known to be unwound.

std/assembly/console.ts
```diff
@@ -0,0 +1,104 @@
...
+    stdout.write("\n");
+  }
+
+  export function time(label: string): void {
+    var stdout = process.stdout;
+    if (timers.has(label)) {
+      stdout.write("Duplicate label '");
+      stdout.write(label);
+      stdout.write("' for console.time()");
+      stdout.write("\n");
```
MaxGraey, line 59:
```suggestion
      stdout.write("' for console.time()\n");
```

std/assembly/console.ts
```diff
@@ -0,0 +1,104 @@
...
+    timers.set(label, process.hrtime());
+  }
+
+  export function timeLog(label: string): void {
+    var stdout = process.stdout;
+    if (!timers.has(label)) {
+      stdout.write("No such label '");
+      stdout.write(label);
+      stdout.write("' for console.timeLog()");
+      stdout.write("\n");
```
MaxGraey, line 71:
```suggestion
      stdout.write("' for console.timeLog()\n");
```

std/assembly/console.ts
```diff
@@ -0,0 +1,104 @@
...
+    timeLogImpl(label);
+  }
+
+  export function timeEnd(label: string): void {
+    var stdout = process.stdout;
+    if (!timers.has(label)) {
+      stdout.write("No such label '");
+      stdout.write(label);
+      stdout.write("' for console.timeEnd()");
+      stdout.write("\n");
```
MaxGraey, line 83:
```suggestion
      stdout.write("' for console.timeEnd()\n");
```

std/assembly/console.ts
```diff
@@ -0,0 +1,104 @@
...
+  var end = process.hrtime();
+  var nanos = end - start;
+  var millis = nanos / 1000000;
+  var millisStr = millis.toString();
+  var stdout = process.stdout;
+  stdout.write(label);
+  stdout.write(": ");
+  stdout.write(millisStr);
+  stdout.write("ms");
+  stdout.write("\n");
```
MaxGraey, line 102:
```suggestion
  stdout.write("ms\n");
```

dcodeIO: Yeah, still wondering how much it's worth to make use of `writeStringFast`. In this particular call, due to the strings being two static latin1 chars each, we get two times `writeStringFast` instead of a redundant encode with garbage. May be worth it here, but not so much in the other cases.

MaxGraey: ok

dcodeIO:
Finally got an incremental GC (with a mini stack) working behind an `--incrementalgc` compiler flag. Means: Performs incremental GC steps at the boundary, as described in my last meeting presentation, without a full-blown shadow stack. 

Here's how it compares (see Parse, Initialize, Compile):

```
// JS
I/O Read   :    17.707 ms  n=105
I/O Write  :    59.846 ms  n=3
Parse      :   478.449 ms  n=161
Initialize :    34.294 ms  n=1
Compile    :  1546.980 ms  n=1
Emit       :  1758.977 ms  n=3
Validate   :   295.908 ms  n=1
Optimize   : 39370.252 ms  n=1
Transform  :          n/a  n=0

// Wasm
I/O Read   :    24.029 ms  n=105
I/O Write  :    67.430 ms  n=3
Parse      :   183.717 ms  n=161
Initialize :    10.902 ms  n=1
Compile    :   768.639 ms  n=1
Emit       :  2032.510 ms  n=3
Validate   :   270.763 ms  n=1
Optimize   : 45017.865 ms  n=1
Transform  :          n/a  n=0

// Wasm --incrementalgc
I/O Read   :    23.222 ms  n=105
I/O Write  :    59.438 ms  n=3
Parse      :   205.150 ms  n=161
Initialize :    17.507 ms  n=1
Compile    :   924.010 ms  n=1
Emit       :  1883.196 ms  n=3
Validate   :   289.005 ms  n=1
Optimize   : 43256.879 ms  n=1
Transform  :          n/a  n=0
```

So far it bootstraps but fails on the test suite. Consider this an experimental starting point. Any help in debugging it is much appreciated :)

dcodeIO:
Current state: The compiler with incremental GC (at the boundary) now bootstraps _and_ passes the entire test suite. The heuristics used is very simple, and I'm still contemplating whether making it more sophisticated would even have merits due to operating on sparse information only obtained when invoking the GC at the boundary anyway. Here's how it works:

* The GC has a constant `STEPSIZE` indicating how often it wants to run in general. Currently that's every 200 new objects.
* When invoked at the boundary, the GC tries to sweep `STEPMUL * STEPSIZE` objects, with `STEPMUL` configuring eagerness. Currently set to 200% of allocations so meaningful progress can be made when previously live objects become garbage.
* To account for very different amounts of debt, i.e. if an export yields much more garbage than `STEPMUL * STEPSIZE`, a linear overshoot is applied on top to always try to reclaim at least half of the debt. Idea is that the GC effectively degrades into stop-the-world territory if exports are heavy, but since it runs right after a heavy export this shouldn't be much different from multiple smaller invocations during the heavy export. How well this hypothesis holds in practice remains to be seen.
* Whenever a collection cycle is completed through stepping, the GC pauses until the amount of allocated objects doubles again. Eventually there should be a break-even point where object counts constantly double and half again, with memory not growing anymore.
* Whenever a collection cycle is not completed through stepping, the GC reschedules itself for the next export, unless debt was low anyhow, then it delays by `STEPSIZE` to potentially aid a sequence of light exports to run without interruption.

As mentioned, this is all quite experimental and I marked it as such. Suggesting to leave the feature opt-in for a while so those interested in it can fiddle with it. There are multiple possible outcomes:
* It's okish
* It's not quite okish, but the heuristics can be improved to make it viable
* Stepping at the boundary is a bad idea in general and the only proper solution is a shadow stack

Hard to tell right now and remains to be seen.

Devs now have the following options:

* By default, the runtime is *not* exported and the simple manual GC suitable in most cases will be used
  * Will include TLSF if there is a dynamic allocation with `new` or `heap.alloc` (the old `--runtime none/stub` is gone)
  * Will eliminate most of the simple GC since `__collect` is typically not called
* One may opt into `--exportRuntime`, gaining external access to `__new` and `__collect`, with MM/GC included
* One may opt into `--incrementalgc`, driving the GC automatically at the boundary, with MM/GC included

The part I'm not 100% sure about is whether we should keep supporting a bump allocator, and if we would, how we'd do that. Settling for TLSF makes matters simpler for us, since there's only one thing to support (in both code and issues), but if there is significant interest in a minimal allocator we may either
* make it a separate runtime package one has to opt into with `--runtime @assemblyscript/minimalRuntime` or
* make it a flag `--minimalRuntime` or
* try our luck with automatic detection of whether memory will ever be `free`d, biting the bullet of always having to ask which one people see in their binaries if someone reports an issue.

Thoughts?

src/asconfig.json
```diff
@@ -7,7 +7,7 @@
     "explicitStart": true,
     "exportRuntime": true,
     "initialMemory": 256,
-    "incrementalgc": true,
+    "runtime": "incremental",
```
MaxGraey, line 10: How about:
```js
   "gc": "none",
   "gc": "default",
   "gc": "incremental",
```

std/assembly/rt/itcms.ts
```diff
@@ -239,15 +225,13 @@ function free(obj: Object): void {
 @global @unsafe
 export function __new(size: usize, id: i32): usize {
   if (size >= OBJECT_MAXSIZE) throw new Error("allocation too large");
-  if (state == STATE_INIT) init();
   var obj = changetype<Object>(__alloc(OBJECT_OVERHEAD + size) - BLOCK_OVERHEAD);
-  obj.linkTo(fromSpace, white); // inits next/prev
   obj.rtId = id;
   obj.rtSize = <u32>size;
-  var ptr = changetype<usize>(obj) + TOTAL_OVERHEAD;
+  obj.linkTo(fromSpace, white); // inits next/prev
```
MaxGraey, line 231: Hmm, `toSpace / fromSpace`. Does it mean we have copying GC (semi-space collector)?

dcodeIO: Thesse are just two linked list (via gcInfo and gcInfo2 in the header). Pointers to objects do not change. `fromSpace` is where new objects go, and `toSpace` is a separate list where reachable objects are moved to, dividing into live objects to keep and garbage objects to sweep, essentially.

src/asconfig.json
dcodeIO: Not sure, since "none" in this case would not only ship no GC, but also a bump allocator instead of TLSF, affecting `heap.alloc` and `heap.free`, which are unrelated to GC. Also doesn't play so well with a custom path to a runtime (here: GC), which must implement both MM and GC.

dcodeIO:
Relevant updates to the documentation are in https://github.com/AssemblyScript/website/pull/39. The diff there should help to get an impression on what'll change.

aminya:
I am glad to see this PR, and sad because I am late to the party. However, I have a suggestion. Now, that you are changing the API, I would recommend using a better notation for the assembly script functions.

There is no need to prepend `__` before the end of the functions. Usually, `__` is used before something that should not be used, not something that every developer needs to call manually.

I would recommend renaming `__collect` to `collect`. If someone wants these functions to have different names (rare), they can use `
```
import { collect as as_collect } from '...'
```

Using `__` and similar notations are used in languages like C that do not have any notion of namespace or module and the names conflict with each other. This is not certainly the case for AssemblyScript.

dcodeIO:
The helpers are named this way so they don't conflict with user functions exported from the entry file. For instance, a module may already export a `collect` or `_collect`, function, hence naming the helper `__collect` avoids a collision in most common cases.

MaxGraey:
What about create inner namespace for collect etc during loader's demangling which rename `__collect` to `collect` and insert into `rt` object? So it could looks like this:
```ts
loader(...)
.than({ exports }) => {
   let ptr = exports.rt.pin(...);
   ...
   exports.rt.collect();
});
```

aminya:
> The helpers are named this way so they don't conflict with user functions exported from the entry file. For instance, a module may already export a collect or _collect, function, hence naming the helper __collect avoids a collision in most common cases.

But TypeScript provides many ways to prevent name conflicting that are better than prefixing underscores.

> What about create inner namespace for collect etc during loader's demangling which rename `__collect` to `collect` and insert into `rt` object / namespace? So it could looks like this:

This sounds great.

dcodeIO:
After that, what if the module exports something named `rt`?

aminya:
FYI, Julia, the only language I know that allows triggering garbage collection manually, uses `GC.gc()` function for triggering garbage collection. 
https://docs.julialang.org/en/v1/base/base/#Base.GC.gc

GC is a module (namespace) and gc is the function name.

MaxGraey:
> After that, what if the module exports something named rt?
```js
function rt(...args) {
   this.collect = function() {
       return mod.exports.__rt_collect();
   };
   this.pin = function () {...}
   // etc
   
   if (exports.rt) {
     return mod.exports.rt(...args);
   } else {
     return this;
   }
}
```

usage:
```js
exports.rt.collect() // call exports.__rt_collect
exports.rt(...) // call exports.rt
```

MaxGraey:
> After that, what if the module exports something named rt?

```js
const _exports = {
  rt: function () {
    console.log("rt call");
  }
}

_exports.rt.collect = function () {
  console.log("__rt_collect call");
}
```
usage:
```js
_exports.rt();
_exports.rt.collect();
```

dcodeIO:
Last commit revives the shadow stack pass and makes the `incremental` runtime fully automatic, removing the mini stack that would probably remain controversial forever.

* Ministack was ~2.2x JS when parsing, shadow stack is ~1.5x JS when parsing.
* Optimized compiler is 1032kb with the shadow stack vs 649kb without a shadow stack.

That's quite a price to pay for fully automatic GC, which was generally expected and has been extensively discussed in meetings, yet the significant code size hit is a bit surprising. From what I can tell so far:

* Shadow stack offsets must be maintained, memory zeroed and slots populated, costing CPU time native would not have
* Instrumentation with stores etc. increases code size by quite a bit
* Even more importantly it seems that instrumentation inhibits Binaryen optimizations where it hurts the most, at locals

Not much to do about that, but it's automatic. May be fine as long as it is not the default and users have options to do manual collect instead. Thoughts?

dcodeIO:
Here's an unfortunate side effect. If one does for example

```ts
assemblyscript.setGlobalAlias(compilerOptions,
  __newString(alias),
  __newString(name)
);
```

one could expect that this just works with `setGlobalAlias` instrumented on the Wasm side to push its managed arguments to the stack, but it doesn't because if one is unlucky the GC steps at `__newString(name)`, prematurely freeing the object returned by `__newString(alias)`. To make that correct. one would have to do

```ts
let aliasPtr = __pin(__newString(alias));
assemblyscript.setGlobalAlias(compilerOptions,
  aliasPtr ,
  __newString(name)
);
__unpin(aliasPtr);
```

preventing premature collection through pinning, which is unfortunate. Currently evaluating how this situation can be avoided, as not having to pin arguments would be so much nicer.

std/assembly/rt/itcms.ts
```diff
@@ -194,7 +199,7 @@ function step(): usize {
         iter = obj.next;
         if (DEBUG) assert(obj.color == i32(!white)); // old white
         free(obj);
-        return 1;
+        return 10;
```
MaxGraey, line 202: why `10`? What does this value means?

std/assembly/rt/itcms.ts
```diff
@@ -176,6 +180,7 @@ function step(): usize {
         __visit_members(changetype<usize>(obj) + TOTAL_OVERHEAD, VISIT_SCAN);
       } else {
         visitRoots(VISIT_SCAN);
+        visitStack(VISIT_SCAN);
```
MaxGraey: According previous `STATE_IDLE` is "fall through" and already called `visitRoots(VISIT_SCAN);`. Could this be simplified to only `visitStack(VISIT_SCAN);` without `visitRoots(VISIT_SCAN)`?

std/assembly/rt/itcms.ts
dcodeIO: This is a magic number introduced by means of guessing, representing the overall "cost" attached to sweeping one object. Affects how many steps the GC performs in succession when it runs incrementally. Per run it has a budget of `GRANULARITY * EAGERNESS / 100` (=400 by default).

std/assembly/rt/itcms.ts
dcodeIO: It does not typically fall through to here. The common case is that `obj != toSpace`, taking the `then` branch visiting one object and yielding. Only when all the initial roots have been marked it will take the `else` branch, where it re-visits all roots and the stack to not miss a new root added since starting the run, and it repeats that (marking one new root/object and yielding, checking again) until it is guaranteed that no root has been missed. Only then it flips and enters the sweep phase. Not visiting the stack initially is an optimization in that objects on the stack are likely to change significantly, hence only visiting the stack before flipping.

std/assembly/rt/itcms.ts
MaxGraey: Could it be hoist to some named enum of constant?)

dcodeIO:
Last commit switches the default runtime to `incremental` (and renames `default` to `minimal`), favoring to provide a carefree package over raw Wasm (stub) or a compromise (minimal), as discussed at the last meeting. Increases test fixture size quite dramatically (now a half a million LOC PR), yet it should be fine to test everything against `incremental`, no matter how simple, just to be sure. Typical optimized runtime-only binary size is ~4KB.

PR #1561, opened 2020-11-29
ASUtil methods are not been exported on loader, such as __newString

breath103:

Hi, since it seems like very straightforward bug (also serious one honestly). 
i'm just making straight up PR. 

# How to reproduce bug
1. some AS file
```
export function concat(a: string, b: string): string {
  return a.concat(b);
}
```
2. compile and load 
```
import * as fs from "fs";
import * as loader from "@assemblyscript/loader";

// this file is from "asc assembly/index.ts -d as.d.ts"
import ASModule from "../as"; 

const wasmModule =
  loader.instantiateSync<typeof ASModule>(fs.readFileSync(__dirname + "/build/optimized.wasm"), {});
console.log(wasmModule.exports);

export default wasmModule.exports;
```

this gives you output of
```
{
  memory: Memory [WebAssembly.Memory] {},
  __new: [Function (anonymous)] { original: [Function: 10] },
  __renew: [Function (anonymous)] { original: [Function: 15] },
  __retain: [Function (anonymous)] { original: [Function: 16] },
  __release: [Function (anonymous)] { original: [Function: 17] },
  __rtti_base: Global [WebAssembly.Global] {},
  concat: [Function (anonymous)] { original: [Function: 19] },
}
```
So all the ASUtil methods (such as **__newString**, which i was looking for in order to use concat method) are missing
this turns out to be problem of Object.create. first argument have to be prototype not properties

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -29,6 +29,7 @@ under the licensing terms detailed in LICENSE:
 * Valeria Viana Gusmao <valeria.viana.gusmao@gmail.com>
 * Gabor Greif <ggreif@gmail.com>
 * Martin Fredriksson <martin.fredriksson@vikinganalytics.se>
+* Kurt Lee <breath103@gmail.com>
 
 Portions of this software are derived from third-party works licensed under
 the following terms:
--- a/lib/loader/umd/index.js
+++ b/lib/loader/umd/index.js
@@ -371,7 +371,7 @@ var loader = (function(exports) {
   
   
   function demangle(exports, extendedExports = {}) {
-    extendedExports = Object.create(extendedExports);
+    extendedExports = { ...extendedExports };
     const setArgumentsLength = exports["__argumentsLength"] ? length => {
       exports["__argumentsLength"].value = length;
     } : exports["__setArgumentsLength"] || exports["__setargc"] || (() => {


breath103:
@dcodeIO So this part of code (https://github.com/AssemblyScript/assemblyscript/blame/8c97612d50a6e5fc3a14d5753c7a577ee22e044d/lib/loader/umd/index.js#L374) seems like originally written by you. i wasn't really sure about intention, can you check on this? 
let me know, thanks

dcodeIO:
The loader-added functions are on the prototype of the returned exports object. Even though these do not show up in a `console.log`, these are still callable. The idea is that none of the loader functions interferes with a Wasm export of the same name, for example if one manually exports a function named `__getString`. Code internal to the loader can still access the original this way, even though the returned demangled object shadows them.

breath103:
that's strange - i actually got "NoMethod" error for this. let me check again

breath103:
@dcodeIO my bad. on actual testing code i wrote __createString(); 🤦 
i get design intention also. that makes sense
i honestly think it's cleaner to just throw error on loader if the given module has conflicting method with util 
- since most likely, overriding __newString kind of methods is not what developer would intended
or just exposing those methods under "Util" class? (with disallowing overriding this class)

let me know about that, but in the meantime, i'll just close this PR since it's not a bug

PR #1568, opened 2020-12-08
fix: Backport heroic fixes made in #1559

dcodeIO:
This PR backports the series of herculean fixes made in the tracing GC branch to master, in the hope that these also benefit ARC.

### DIFF
--- a/src/builtins.ts
+++ b/src/builtins.ts
@@ -68,8 +68,6 @@ import {
   getConstValueI32,
   getConstValueF32,
   getConstValueF64,
-  Relooper,
-  RelooperBlockRef,
   SIMDLoadOp,
   getLocalGetIndex,
   createType,
@@ -82,7 +80,6 @@ import {
   Field,
   Global,
   DecoratorFlags,
-  Element,
   ClassPrototype,
   Class
 } from "./program";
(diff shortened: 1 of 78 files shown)


src/builtins.ts
```diff
@@ -8691,134 +8688,184 @@ export function compileVisitGlobals(compiler: Compiler): void {
   );
 }
 
-/** Compiles the `visit_members` function. */
-export function compileVisitMembers(compiler: Compiler): void {
+/** Ensures that the visitor function of the specified class is compiled. */
+function ensureVisitMembersOf(compiler: Compiler, instance: Class): void {
```
dcodeIO, line 8692: A new runtime visitor implementation that generates per-class functions instead of a huge unwieldy switch. Increases binary size somewhat, but is much easier to reason about.

src/compiler.ts
```diff
@@ -10508,8 +10508,12 @@ export class Compiler extends DiagnosticEmitter {
       }
     }
     var parameterTypes = signature.parameterTypes;
+    var parameterNodes = reportNode.parameters;
     for (let i = 0, k = parameterTypes.length; i < k; ++i) {
-      if (!this.checkTypeSupported(parameterTypes[i], reportNode.parameters[i])) {
+      let parameterReportNode: Node;
+      if (parameterNodes.length > i) parameterReportNode = parameterNodes[i];
```
dcodeIO, line 10514: Here, the function may come from a function expression with an omitted argument being compiled in context of a function type with fewer argument expressions, implicitly dropping excess arguments due to being irrelevant. Doesn't error when compiling to JS due to undefined, but produced a fine runtime index-out-of-bounds in Wasm.

src/flow.ts
```diff
@@ -612,7 +612,7 @@ export class Flow {
   /** Tests if the specified `this` field has the specified flag or flags. */
   isThisFieldFlag(field: Field, flag: FieldFlags): bool {
     var fieldFlags = this.thisFieldFlags;
-    if (fieldFlags) {
+    if (fieldFlags != null && fieldFlags.has(field)) {
```
dcodeIO, line 615: This is old code apparently, written before the requirement for a `Map#has` before a `Map#get` was conceived. Produced a wholesome key-does-not-exist error in Wasm.

src/glue/wasm/i64.ts
```diff
@@ -23,7 +23,7 @@ function i64_is<T>(value: T): bool {
 // @ts-ignore: decorator
 @global @inline
 function i64_new(lo: i32, hi: i32 = 0): i64 {
-  return lo | (hi << 32);
+  return <i64><u32>lo | (<i64>hi << 32);
```
dcodeIO, line 26: A spectacular piece of code that was completely wrong before, and looks absolutely bogus now. Apart from the missing cast to `i64` to ensure proper bit size, the `<i64><u32>` is necessary so the compiler doesn't attempt to sign-extend due to the cast value being signed. Somewhat strange, and made me wonder whether we can do better, but without any deeper changes that's the awesome fix fixing things.

src/program.ts
```diff
@@ -4273,9 +4275,8 @@ export class Class extends TypedElement {
     var instance: Class | null = this;
     do {
       let overloads = instance.overloads;
-      if (overloads) {
-        let overload = overloads.get(kind);
-        if (overload) return overload;
+      if (overloads != null && overloads.has(kind)) {
```
dcodeIO, line 4278: Another fix within the `Map#has` before `Map#get` wasnt-even-conceived-by-someone category. No error in JS, entirely justified fireworks in Wasm.

std/assembly/array.ts
```diff
@@ -563,7 +563,7 @@ export class Array<T> {
 
   // RT integration
 
-  @unsafe private __visit_impl(cookie: u32): void {
+  @unsafe private __visit(cookie: u32): void {
```
dcodeIO, line 566: Perfection is achieved not when there is nothing more to add, but when there is nothing left to take away.

dcodeIO (commented):
Here's why

dcodeIO:
Btw, in case you are disappointed, a more impressive fix was in https://github.com/AssemblyScript/assemblyscript/pull/1567 already. Finding that without source maps support running Wasm in node was quite the achievement. Thanks to Max for helping me to figure this one out :)

dcodeIO:
Also, I opened a bottle of sparkling wine that has been occupying my fridge for far too long with no proper valuation for the sheer amount of work and energy invested into producing and shipping it. I forgot who gave it to me. I'm either sorry or I'm not.

PR #1572, opened 2020-12-10
Export block and object layout constants from Rtrace

dcodeIO:
cc @jtenner 


jtenner (approved):
I really like these changes. Thank you for your help

PR #1574, opened 2020-12-11
Support more options for providing Transforms to compiler frontend API

A-Babin:
Hi there!

This PR adds support for providing custom Transforms directly to the compiler frontend. A new optional parameter is added to `compileString` and a new prop is added to `APIOptions`, accepting an array of Transform constructors or instances. To keep it simple, the implementation avoids affecting any existing arg/option parsing. 

Configuration can be passed to the Transform, which can't be done trivially with dynamically imported modules.  

```javascript
// ...
}

const output = compileString(source, opts, [MyTransform, new MyTransform()]);

// or

main(argv, {
  //...
  transforms: [MyTransform, new MyTransfrom()]
});
```

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -29,6 +29,7 @@ under the licensing terms detailed in LICENSE:
 * Valeria Viana Gusmao <valeria.viana.gusmao@gmail.com>
 * Gabor Greif <ggreif@gmail.com>
 * Martin Fredriksson <martin.fredriksson@vikinganalytics.se>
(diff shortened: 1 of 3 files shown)


dcodeIO:
Makes me wonder whether we may just accept an instantiated `Transform` as a `--transform` argument, which the API can handle while irrelevant for the CLI, avoiding appending somewhat special arguments. For instance, the next similar feature would append an argument again and so on, which is probably good to avoid.

PR #1590, opened 2020-12-18
Unify ctor generation with and without base ctor

dcodeIO:
Fixes an oddity with automatically generated constructors, where if the class has a base class the generated ctor would not be exported along an `export class Something`.


dcodeIO:
On a second look this also fixes a bug with field initialization when calling an exported constructor externally where the derived class has additional own fields, due to export logic picking up the member named `"constructor"`, which is the inherited base constructor that does not initialize the additional fields.

PR #1598, opened 2020-12-23
asinit: Add WebAssembly text files to build/.gitignore

Pospelove:
After running `asinit` and `npm run asbuild`, `optimized.wat` and `untouched.wat` has been generated. Since they are in the build directory they probably should not be versioned.


Pospelove:
image

PR #1625, opened 2021-01-18
Expose program diagnostics in cli main callback

piotr-oles:
This PR adds `reportDiagnostic` callback to the `asc.main()` API options. This will help integrate AssemblyScript with other tools, like webpack. I could use the `assemblyscript` package directly, but there is a lot of logic inside `asc.main()` that I wouldn't like to duplicate and then synchronize :)


cli/asc.js
```diff
@@ -625,7 +625,7 @@ exports.main = function main(argv, options, callback) {
     if (numErrors) {
       const err = Error(numErrors + " parse error(s)");
       err.stack = err.message; // omit stack
-      return callback(err);
+      return callback(err, program.diagnostics);
```
dcodeIO, line 628: When compiling the compiler to Wasm, `program.diagnostics` is just a pointer into memory, making this non-portable. Perhaps, an alternative could be to add a callback to `APIOptions` that receives all the concrete data as arguments, not just a pointer, with the compiler only translating the diagnostics from memory to JS objects if the callback is present?

piotr-oles: Makes sense :+1: It could be named `reportDiagnostic` to follow [TypeScript naming](https://github.com/microsoft/TypeScript/blob/716b167f2b19e1642bd65aba33400afe8c181889/src/compiler/watchPublic.ts#L196). I will update it having WASM portability in mind :)

piotr-oles: How can I test it with the WASM version? I was able to build it with `asbuild` script from package.json - I have the `out` directory with binaries. I don't know what to do next :)

dcodeIO: The Wasm variant in master is only partially functional currently. But the [stw](https://github.com/AssemblyScript/assemblyscript/tree/stw) branch, which is scheduled to land soon-ish, has a fully functional bootstrap. You could base this PR on top of the stw branch, and rebase once it lands. Instructions to run the bootstrapped compiler in the stw branch are:

```
npm run bootstrap:untouched
node bin/asc --wasm out/assemblyscript.untouched-bootstrap.wasm ...
```

Note that some internals changed there, with the changes documented in [this PR](https://github.com/AssemblyScript/website/pull/39) scheduled to land along the branch.

piotr-oles: I've updated the PR and tested it with regular and WASM version :) I'm not sure if we want to have all this code, that copies diagnostic from WASM, in `asc.js`, or if we want to move some of the logic to the assemblyscript binary.

cli/asc.js
```diff
@@ -1165,6 +1171,64 @@ function checkDiagnostics(program, stderr) {
         EOL + EOL
       );
     }
+    if (reportDiagnostic) {
+      const diagnostic = __wrapDiagnosticMessage(diagnosticPtr);
+      __pin(diagnostic.message);
```
dcodeIO, line 1176: Pinning here can be reduced to just `__pin(diagnostic)` while it is used, as the diagnostic object itself keeps all its members alive. A single `__unpin(diagnostic)` at the end will do then as well.

cli/asc.js
```diff
@@ -140,13 +140,19 @@ function loadAssemblyScript() {
     __pin = assemblyscript.__pin;
     __unpin = assemblyscript.__unpin;
     __collect = assemblyscript.__collect;
+    __wrapDiagnosticMessage = assemblyscript.DiagnosticMessage.wrap;
+    __wrapRange = assemblyscript.Range.wrap;
+    __wrapSource = assemblyscript.Source.wrap;
```
dcodeIO, line 145: Good idea to use the loader here. Makes me wonder if it'd make sense to simplify how this is used, perhaps a general helper along the lines of:

```js
function ensureObject(ptrOrObj, ctor) {
  if (typeof ptrOrObj === "number") return ctor.wrap(ptr);
  return ptrOrObj;
}
ensureObject(range, assemblyscript.Range);
```

dcodeIO (commented):
Cool, this is getting alone nicely :) Just a few comments.

cli/asc.js
dcodeIO: Hopes are that we can get rid of the translation layer eventually, once Wasm can work with objects and pass them between the two worlds. Until then, if we can make this simpler somehow, that'd also be good to do, but typically quickly goes into the direction of generating JS glue code. Has its pros and cons.

cli/asc.js
piotr-oles: oh, good to know :)

cli/asc.js
piotr-oles: I've created `__wrap` function :)

dcodeIO:
Thanks! :)

PR #1631, opened 2021-01-24
Adding valueOf override to generated TS class definitions.

forcepusher:
This PR complements [this issue](https://github.com/AssemblyScript/website/issues/41). I've realized that part of the problem why I found that feature so late is because the TS definition doesn't come with valueOf() override that returns `usize` pointer instead of an `Object`.


dcodeIO:
Thank you! :)

PR #1639, opened 2021-01-27
Pedantic complains about bounds check in set.values()

forcepusher:
This should be pretty straightforward fix. Pedantic throws this warning whenever set.values() method is used.
```
PEDANTIC AS904: Indexed access may involve bounds checking.
         values[length++] = entry.key;
         ~~~~~~~~~~~~~~~~
 in ~lib/set.ts(188,9)
 ```

### DIFF
--- a/std/assembly/set.ts
+++ b/std/assembly/set.ts
@@ -185,7 +185,7 @@ export class Set<T> {
     for (let i = 0; i < size; ++i) {
       let entry = changetype<SetEntry<T>>(start + <usize>i * ENTRY_SIZE<T>());
       if (!(entry.taggedNext & EMPTY)) {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by forcepusher** [comment]
Not sure why it's failing.
```
FAILURE: 1 compiler tests had failures:

  std/set 
```
That's all the info I've got from it. Tried poking it with a debugger, it didn't break on exception.
Halp.

forcepusher:
Too much trouble for a preliminary code. I'm sure it'll get rewritten later anyway.

dcodeIO:
Fwiw, the test failure looks like a fixture difference after applying the changes, making it necessary to recreate the ground truth fixture via `npm run test:compiler std/set -- --create`.

PR #1662, opened 2021-02-03
Reject identifier or keyword immediately following a numeric literal

dcodeIO:
Rejects invalid code like `1u8`, just like TS would, by complaining about an identifier or keyword immediately following a numeric literal, during parsing.

Does not yet account for a BigInt literal's `n` postfix, since BigInts are not yet supported anyway. Other than that there may still be more to do here, like rejecting `1 2` or `a b` complaining about a missing `;` as TS would, but seems less urgent and is unrelated.

fixes https://github.com/AssemblyScript/assemblyscript/issues/1637


dcodeIO:
Pushed a fix for `pos + 1` to the branch, but doesn't update for some reason. Perhaps a problem with GH.

MaxGraey (approved):

Issue #1677, opened 2021-02-08
Use of super in a non-derived class results in an assertion


saulecabrera:
The following snippet should emit a diagnostic instead of an assertion

```ts
export class B {
  constructor(a: i32) {
    super(x);
  }
}

export function add(a: i32, b: i32): i32 {
  new B(a);
  return a + b;
}
```

I believe that the diagnostic should be something like:

```
'super' can only be referenced in a derived class.
```

PR #1678, opened 2021-02-08
Emit a diagnostic when 'super' is called in ctor of a non-derived class

dcodeIO:
Calling `super` in the constructor of a non-derived class was hitting an assertion. Emit a diagnostic instead.

fixes https://github.com/AssemblyScript/assemblyscript/issues/1677


MaxGraey (approved):

saulecabrera:
Thanks for the quick fix.

Issue #1675, opened 2021-02-07
WASM validation errors when using generic getters?


jriehl:
So a funny thing happened while on my way to trying to demonstrate #1635.  Running from a fresh install of  https://github.com/jriehl/sandbox/tree/master/assemblyscript/issue1635, I see the following:

```shell
% npm run asbuild

> issue1635@1.0.0 asbuild
> npm run asbuild:untouched && npm run asbuild:optimized


> issue1635@1.0.0 asbuild:untouched
> asc assembly/index.ts --target debug

[wasm-validator error in function export:assembly/tree/Tree.Node<i32>#get:payload] unexpected false: local.get index must be small enough, on 
(local.get $0)
[wasm-validator error in function export:assembly/tree/Tree.Node<i32>#get:payload] unexpected false: call target must exist, on 
(call $assembly/tree/Tree.Node<i32>#get:payload)
[wasm-validator error in function export:assembly/tree/Tree.Node<i32>#get:left] unexpected false: local.get index must be small enough, on 
(local.get $0)
[wasm-validator error in function export:assembly/tree/Tree.Node<i32>#get:left] unexpected false: call target must exist, on 
(call $assembly/tree/Tree.Node<i32>#get:left)
[wasm-validator error in function export:assembly/tree/Tree.Node<i32>#get:right] unexpected false: local.get index must be small enough, on 
(local.get $0)
[wasm-validator error in function export:assembly/tree/Tree.Node<i32>#get:right] unexpected false: call target must exist, on 
(call $assembly/tree/Tree.Node<i32>#get:right)
FAILURE validate error
    at Object.main (/Users/jon/git/sandbox/assemblyscript/issue1635/node_modules/assemblyscript/cli/asc.js:850:23)
    at /Users/jon/git/sandbox/assemblyscript/issue1635/node_modules/assemblyscript/bin/asc:19:47
```

I don't know why the getters aren't being emitted for the Node class, but that seems to be what the WASM validator is complaining about.

PR #1683, opened 2021-02-14
Ensure exported field getters and setters are compiled

dcodeIO:
Exports logic made the false assumption that a field that is tagged as `COMPILED` always has its implicit getters and setters compiled, which is not always the case.

fixes https://github.com/AssemblyScript/assemblyscript/issues/1675


dcodeIO:
Reproduction here appears to be ordering dependent, somewhere in between star exports, generics and virtual overloads, not sure. The fix is rather trivial and reliable, though, so merging without a test.

PR #1687, opened 2021-02-16
Prevent webpack from detecting calls to `require` from `dyrequire`

AurelienRibon:
# ISSUE
Webpack is used to generate the bundle in `dist/` folder, that is included in the `npm` package. However, when users install the package in their project `npm install assemblyscript`, include ASC with `const asc = require('assemblyscript/dist/asc.js')`, and try to bundle their own project with Webpack, there are multiple warnings emitted, saying that calls to `require` are made in a way that prevents static analysis of dependencies.

This is caused by this code:
```js
// Original source code
const dynrequire = typeof __webpack_require__ === "function"
  ? __non_webpack_require__
  : require;

// Generated code from `npm run build:bundle` (without minification)
const dynrequire = true
  ? require // <--- that's our issue
  : 0;
```
When users call Webpack themselves in their project, it will detect that `dynrequire === require` (I was surprised it could do that, the constant condition `true` may help). Thus, that leads to many errors, from `ts-node` module is not found (we are in a web bundle...), to "require is called in a way that prevents static analysis".

image

# SOLUTION
The variable `require` must not be present in the distributed code when requiring Node-only stuff. `__non_webpack_require__` is replaced by webpack with `require`. Thus it should only be used by leaf projects, not by libraries, else that creates issues in leaf projects. The solution consists in just replacing that `__non_webpack_require__ ` by `eval('require')`, so that

### DIFF
--- a/cli/asc.js
+++ b/cli/asc.js
@@ -47,7 +47,7 @@ const find = require("./util/find");
 const binaryen = global.binaryen || (global.binaryen = require("binaryen"));
 
 const dynrequire = typeof __webpack_require__ === "function"
-  ? __non_webpack_require__
+  ? eval('require')
   : require;
 
 const WIN = process.platform === "win32";
--- a/cli/asc.js
+++ b/cli/asc.js
@@ -47,7 +47,7 @@ const find = require("./util/find");
 const binaryen = global.binaryen || (global.binaryen = require("binaryen"));
 
 const dynrequire = typeof __webpack_require__ === "function"
-  ? eval('require')
+  ? __non_webpack_require__
   : require;
 
 const WIN = process.platform === "win32";
--- a/cli/asc.js
+++ b/cli/asc.js
@@ -47,7 +47,7 @@ const find = require("./util/find");
 const binaryen = global.binaryen || (global.binaryen = require("binaryen"));
 
 const dynrequire = typeof __webpack_require__ === "function"
-  ? __non_webpack_require__
+  ? () => module[`require`] // backticks required
   : require;
 
 const WIN = process.platform === "win32";
--- a/cli/util/options.js
+++ b/cli/util/options.js
@@ -238,7 +238,7 @@ function merge(config, currentOptions, parentOptions, parentBaseDir) {
 exports.merge = merge;
 
 const dynrequire = typeof __webpack_require__ === "function"
-  ? __non_webpack_require__
+  ? () => module[`require`] // backticks required
   : require;
 
 /** Resolves a single possibly relative path. Keeps absolute paths, otherwise prepends baseDir. */
--- a/cli/util/options.js
+++ b/cli/util/options.js
@@ -238,7 +238,7 @@ function merge(config, currentOptions, parentOptions, parentBaseDir) {
 exports.merge = merge;
 
 const dynrequire = typeof __webpack_require__ === "function"
-  ? __non_webpack_require__
+  ? eval('require')
   : require;
 
 /** Resolves a single possibly relative path. Keeps absolute paths, otherwise prepends baseDir. */
--- a/cli/util/options.js
+++ b/cli/util/options.js
@@ -238,7 +238,7 @@ function merge(config, currentOptions, parentOptions, parentBaseDir) {
 exports.merge = merge;
 
 const dynrequire = typeof __webpack_require__ === "function"
-  ? eval('require')
+  ? __non_webpack_require__
   : require;
 
 /** Resolves a single possibly relative path. Keeps absolute paths, otherwise prepends baseDir. */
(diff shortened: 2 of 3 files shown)


dcodeIO:
Interesting, the `eval` is what we had originally, but then switched. I don't recall exactly anymore why we did that, hmm. @MaxGraey ?

AurelienRibon:
I saw that in `dist/sdk.js`, you expose `asc` as `require('assemblyscript/cli/asc.js')`, for the Node version. Does this mean that `dist/asc.js` is only meant to be used by browsers? If so, it may even be better to replace `__non_webpack_require__` by a `throw`, because no code path should try to require `ts-node` in the browser.

In the meantime, this solution works as well, as `eval('require')` will return `undefined`, which will throw when called as a function. And this has the benefit of letting Node users still require `dist/asc.js`, instead of `cli/asc.js`.

dcodeIO:
Yes, `dist/asc.js` originated as a browser-only build of the compiler. I'd not be opposed to make it more universal, though, if it can be done without breakage.

AurelienRibon:
The current `dist/asc.js` should be useable from the CLI, on master and on this branch too, as it still tries to require `ts-node` and inject stuff.

That aside, not being able to use `dist/asc.js` from a Webpack project is an issue, as there are few production grade projects that import dependencies without a bundler. Ideally, `asc.js` should be decoupled between a lib agnostic from any Node requirement (and that would be exposed for browsers), and a wrapper that deals with `ts-node` and all that stuff. But one thing at a time :-)

cli/asc.js
```diff
@@ -47,7 +47,7 @@ const find = require("./util/find");
 const binaryen = global.binaryen || (global.binaryen = require("binaryen"));
 
 const dynrequire = typeof __webpack_require__ === "function"
-  ? __non_webpack_require__
+  ? eval('require')
```
dcodeIO, line 50: One reason for making the switch that comes to my mind is that using `eval` may not be allowed in certain environments, for example if `eval` is disabled by CSP. Iirc there are setups where such a JS file would not work at all, while other environments may print a warning. At the very least we need a try-catch around `eval` I think.

AurelienRibon: You are right, this won't fit like this. And when I think about it, the main reason may just be that it will not work at all, because `require` being undefined (as in "not defined at all"), this `eval` will throw instead of returning undefined.

Forget this for now, I'll rework all this by using `__non_webpack_require__` WITHOUT letting webpack detect that `dynrequire ===  __non_webpack_require__`, that would solve the problem.

AurelienRibon:
More background: if webpack is able to infer that `dynrequire === require`, it is because of the minifier:
```js
// Original source code
const dynrequire = typeof __webpack_require__ === "function"
  ? __non_webpack_require__
  : require;

// Generated code from `npm run build:bundle` (without minification)
const dynrequire = true
  ? require
  : 0

// Generated code with minification
d=require // `d` is dynrequire
```
Webpack does not resolve static conditions, but it can resolve variable reassignment.
I will find a way. It's important that `asc.js` can be integrated in  Webpack bundles.

cli/util/options.js
```diff
@@ -238,7 +238,7 @@ function merge(config, currentOptions, parentOptions, parentBaseDir) {
 exports.merge = merge;
 
 const dynrequire = typeof __webpack_require__ === "function"
-  ? __non_webpack_require__
+  ? () => module[`require`] // backticks required
   : require;
```
MaxGraey, line 242:
```js
const dynrequire = typeof __webpack_require__ === "function"
  ? typeof module !== 'undefined' && 
    typeof module.require === 'function' && 
    module.require.bind(module)
  : require;
```

dcodeIO: Would that work with ESM (no `module`)?

MaxGraey: I don't know

AurelienRibon: Unfortunately `module.require` is not `require`. The latter is a wrapper around the former, and only this latter provides the `.resolve()` method that you are relying on.

I give up. Besides using `eval('require')` (with a try/catch), I cannot find a way to make it work.
But that's a problem, because the library cannot be included in a Webpack project without dirty tricks (such as forcing webpack to ignore `ts-node`).

Do you really need `require.resolve`? I need to delve into how it is used by options.js and asc.js. Since we are sure this call will be done by a Node environment, we could do the same with `fs` maybe.

AurelienRibon:
Related to this issue, `asc.js` cannot be required on Observable.

Another solution I though of was to generate to versions of asc.js, one for the CLI, and one just for the browser.
Basically, what people want to do with a browser version of asc is just to be able to run `compileString` and get a WASM binary, nothing more. All this is possible with the assemblyscript.js file, but it's not easy to figure out in which order to call methods (build a program, inject lib strings, inject user strings, parse all that, instantiate the program, compile). A demo proof of concept of such "easy compile from browser" would be awesome :)

image

dcodeIO:
The roadmap I have in mind currently is about:
- Get an ESM build of binaryen.js going
- Get an ESM build of long.js going (trivial)
- Transition asc.js to ESM
- Make an ESM bundle of everything, and transpile to an ES5 module in addition

HerrCai0907:
AS already ports to ESM, We can close this PR.

PR #1688, opened 2021-02-16
fix: reexport Source, Range and DiagnosticMessage from compiler's index.d.ts

willemneal:
This means any tool which imports the types from both, e.g. vistor-as there will be one source of truth.

### DIFF
--- a/cli/asc.d.ts
+++ b/cli/asc.d.ts
@@ -4,6 +4,8 @@
  */
 
 import { OptionDescription } from "./util/options";
+import { Source, Range, DiagnosticMessage } from "..";
+
 export { OptionDescription };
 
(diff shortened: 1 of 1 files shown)


dcodeIO:
If I understood correctly, `asc` does indeed not expose all members of the internal classes in question, and I think that made it necessary to define a subset of only those properties actually present. Not sure if that applies to all of them, though.

piotr-oles:
Yes, these typings are a subset on purpose - we are exposing only a few fields. It's because the compiler can be compiled to WASM and if it runs in a WASM environment, then we have to copy these diagnostics. We could use the `Pick<>` type instead of duplicating them - but re-exporting them would create invalid expectations in some cases :)

PR #1694, opened 2021-02-22
Make loader more compatible with plain ESM

trusktr:
This makes it possible to also write `import {...} from "@assemblyscript/loader/index.js"`. Some people need to write this because some tools (for example TypeScript compiler) do not yet understand how to read the `exports` field from package.json. These tools that don't know `exports` will work fine with paths like `"@assemblyscript/loader/index.js"`, regardless if there's an `exports` field or not.

However, without adding this line, the same code that imports from `"@assemblyscript/loader/index.js"` will not work in a system that uses Node ESM (Node or Webpack) and they will throw an error that `index.js` exists but it is not exported therefore not importable.

TLDR: this allows us to write code that works in more places.

⯈
⯈
⯈

- [ ] I've read the contributing guidelines

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -33,6 +33,7 @@ under the licensing terms detailed in LICENSE:
 * Piotr Oleś <piotrek.oles@gmail.com>
 * Saúl Cabrera <saulecabrera@gmail.com>
 * Chance Snow <git@chancesnow.me>
+* Joe Pea <joe@trusktr.io>
 
 Portions of this software are derived from third-party works licensed under
 the following terms:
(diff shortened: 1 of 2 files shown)


trusktr:
Adding to the commit message, the error people will currently get if they import from `@assemblyscript/loader/index.js` is `ERR_PACKAGE_PATH_NOT_EXPORTED`.

trusktr:
For now, I published this fork of AS Loader at http://npmjs.com/@lume/as-loader so we can use it in lume/glas.

dcodeIO:
Iirc there is a fallback notation to allow this, for everything that isn't otherwise defined. Something involving forwards to `*.js*  or so, don't remember exactly.

trusktr:
Yeah, that format would be `"./*": "./*"` if you want to unblock everything

trusktr:
I'm still using this fork in lume/glas. I'll circle back around to clean this up...

PR #1703, opened 2021-02-28
fix misleading compile error on multi-assignment

### LINKED ISSUE #1699: misleading compiler error
See the output from my travis build using the latest AssemblyScript 0.18.11 ( upgrading from 0.17.7 )

https://travis-ci.com/github/petersalomonsen/javascriptmusic/jobs/485666265#L267

```
ERROR TS2469: The '++' operator cannot be applied to type 'assembly/midi/midisynth/MidiVoice'.
       for (let n=0; n < numActiveVoices; n++) {
                                          ~~~
```

I would expect that `n` is interpreted as type `i32` and not as `MidiVoice`.

The exact same code pass in AssemblyScript 0.17.7.

Also tested with 0.17.13 and 0.18.0, and found that the error was introduced in 0.18.0

------------------
Test case for reproducing:

```
class Foo {

}

function test(): void {
    const foos = new Array<Foo>(1);
    foos[0] = foos[0] = new Foo();

    for (let n = 0; n < foos.length; n++) {
        const v = foos[n];
    }
}

test();
```

The error is actually caused by a mistake in my code in the line `foos[0] = foos[0] = new Foo();` which should have been `foos[0] = new Foo();`. However the compile error message is misleading, and is not really a compile error.

I'll look into this myself later on, but just wanted to file the issue in case anyone else get an instant idea of what it might be.

petersalomonsen:
multiassignment causes compiler error later on code that has no errors

fix #1699

⯈
⯈
⯈

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -33,6 +33,7 @@ under the licensing terms detailed in LICENSE:
 * Piotr Oleś <piotrek.oles@gmail.com>
 * Saúl Cabrera <saulecabrera@gmail.com>
 * Chance Snow <git@chancesnow.me>
+* Peter Salomonsen <petersalomonsen@runbox.com>
 
 Portions of this software are derived from third-party works licensed under
 the following terms:
--- a/src/compiler.ts
+++ b/src/compiler.ts
@@ -5988,7 +5988,6 @@ export class Compiler extends DiagnosticEmitter {
           let tempTarget = flow.getTempLocal(thisType);
           let tempElement = flow.getTempLocal(elementType);
           let returnType = getterInstance.signature.returnType;
-          flow.freeTempLocal(tempTarget);
           let ret = module.block(null, [
             this.makeCallDirect(setterInstance, [
               module.local_tee(tempTarget.index, thisExpr, thisType.isManaged),
--- a/tests/compiler/issues/1699.optimized.wat
+++ b/tests/compiler/issues/1699.optimized.wat
@@ -1,0 +1,2442 @@
+(module
+ (type $i32_i32_=>_none (func (param i32 i32)))
+ (type $none_=>_none (func))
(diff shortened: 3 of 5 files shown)


src/flow.ts
```diff
@@ -327,7 +327,8 @@ export class Flow {
       }
       local = parentFunction.addLocal(type);
     } else {
-      if (temps !== null && temps.length > 0) {
+      if (temps !== null && temps.length > 0
+          && temps[temps.length-1].type === type) {
```
dcodeIO, line 331: What are the types here when the conflict happens? Asking because the `temps` array should already only contain matching types (see the switch above), so this is a bit unexpected.

petersalomonsen: it's because custom class types becomes usize which resolves to i32 in the switch above. and so the temps contains both i32 and other class types.

dcodeIO: Hmm, thinking about this a little more, what may be happening there is that a temporary local is not properly blocked to be not re-used during compilation. The mechanism there is to obtain a temporary local, use it, and free it again so it can be reused later on. Now, if the free happens too early, things may go wrong, which may be the root cause.

petersalomonsen: Yeah I was suspecting I’m just workin around a more underlying problem here. Will see if freeing happens too early.

dcodeIO:
Ouch :)

src/flow.ts
petersalomonsen: I think I got it now @dcodeIO , `flow.freeTempLocal(tempTarget);` was called twice in compiler / makeAssignment

petersalomonsen:
ouch what ? :) 

Ready for review now @dcodeIO

dcodeIO:
Ouch as in "what the heck did I think when I wrote that code" 😄

dcodeIO:
Perfect, thanks for the fix! :)

PR #1716, opened 2021-03-06
Transform glob entries into paths

jerrygreen:
- fix https://github.com/AssemblyScript/assemblyscript/issues/1712

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -33,6 +33,7 @@ under the licensing terms detailed in LICENSE:
 * Piotr Oleś <piotrek.oles@gmail.com>
 * Saúl Cabrera <saulecabrera@gmail.com>
 * Chance Snow <git@chancesnow.me>
+* Jerry Green <me@jerrygreen.me>
 * Peter Salomonsen <petersalomonsen@runbox.com>
 
 Portions of this software are derived from third-party works licensed under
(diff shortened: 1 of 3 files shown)


JerryGreen:
@dcodeIO can you tell me what is this `LIBRARY_PREFIX.valueOf()` and why `LIBRARY_PREFIX` might be `undefined`? I don't quite understand this...

```txt
 /home/runner/work/assemblyscript/assemblyscript/cli/asc.js:175
exports.libraryPrefix = __getString(assemblyscript.LIBRARY_PREFIX.valueOf());
                                                                  ^
TypeError: Cannot read property 'valueOf' of undefined
```

And this `asconfig.json` actually works now:

```json
{
  "entries": ["./src/**/*.as.ts"]
}
```

dcodeIO:
`LIBRARY_PREFIX` can be an exported WebAssembly global, and the proper way to get its value is to use `valueOf`, which is specified to return the primitive value of an object, so also works on strings. Not sure why it would fail.

JerryGreen:
Ok, I don't know either, but I can say compiler actually works. I temporarily use my fork in my project like this:

```jsonc
// package.json
{
  "devDependencies": {
    "assemblyscript": "https://github.com/jerrygreen/assemblyscript/tarball/jerrygreen",
  }
}
```

Which accumulates both this change and also: https://github.com/AssemblyScript/assemblyscript/pull/1705. And I'm ok with it currently. Maybe I'll figure out what's wrong with this change later. Or not. Will see :)

PR #1728, opened 2021-03-11
tests: Add tests for StaticArray

saulecabrera:
Adds the following missing tests for StaticArray:

- constructor
- fromArray
- concat (static)
- slice (static)
- concat (instance)
- includes
- indexOf
- join + toString

### DIFF
--- a/tests/compiler/std/staticarray.optimized.wat
+++ b/tests/compiler/std/staticarray.optimized.wat
@@ -1,13 +1,13 @@
 (module
+ (type $i32_i32_=>_i32 (func (param i32 i32) (result i32)))
+ (type $i32_i32_i32_=>_i32 (func (param i32 i32 i32) (result i32)))
  (type $i32_i32_=>_none (func (param i32 i32)))
  (type $none_=>_none (func))
  (type $i32_i32_i32_=>_none (func (param i32 i32 i32)))
- (type $i32_i32_=>_i32 (func (param i32 i32) (result i32)))
  (type $i32_=>_none (func (param i32)))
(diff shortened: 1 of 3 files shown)


saulecabrera:
@dcodeIO @MaxGraey one thing that I noticed while working on this is that we have `isArray` for which the documentation states the following:

> Tests if the specified type or expression can be used as an array. Compiles to a constant.

But `isArray`  returns false with `StaticArray`, which can be used as an array (I know that it technically doesn't extend it) but this seems inconsistent/counterintuitive. Should this be fixed in the docs? 

There's also `isArrayLike`, and in this case, `isArrayLike` returns `true` for a `StaticArray`, but there is no documentation for this one, is there a reason for this? Or should we document it?

saulecabrera:
@dcodeIO 👋 any thoughts on this?

dcodeIO:
It's a bit unfortunate that `StaticArray` is somewhat incompatible with normal `Array`s, even though that's about the point to introduce it. My immediate thought was that we should have named it differently, something without an `Array` in it perhaps, but now we are kinda stuck with it. So, to be honest, I don't know what to do, but documenting the difference certainly makes sense :)

saulecabrera:
> So, to be honest, I don't know what to do, but documenting the difference certainly makes sense :)

That's a valid outcome of this discussion, I think that as long as the differences are pointed out correctly then the name becomes less relevant. That said, I didn't intend to make my comment above a blocker to merge this, it was mostly observation. I'll follow up with a documentation issue/PR.

MaxGraey (approved):

dcodeIO:
Thanks! :)

PR #1742, opened 2021-03-17
fix: read stream signature

willemneal:
It is compared to a usize later and used as one later.

### DIFF
--- a/std/assembly/process.ts
+++ b/std/assembly/process.ts
@@ -133,7 +133,7 @@ abstract class WritableStream extends Stream {
 abstract class ReadableStream extends Stream {
   read(buffer: ArrayBuffer, offset: usize = 0): i32 {
     var end = <usize>buffer.byteLength;
-    if (offset < 0 || offset > end) {
+    if (offset > end) {
       throw new Error(E_INDEXOUTOFRANGE);
     }
     store<usize>(iobuf, changetype<usize>(buffer) + offset);
--- a/std/assembly/process.ts
+++ b/std/assembly/process.ts
@@ -131,7 +131,7 @@ abstract class WritableStream extends Stream {
 
 @unmanaged
 abstract class ReadableStream extends Stream {
-  read(buffer: ArrayBuffer, offset: isize = 0): i32 {
+  read(buffer: ArrayBuffer, offset: usize = 0): i32 {
     var end = <usize>buffer.byteLength;
     if (offset < 0 || offset > end) {
       throw new Error(E_INDEXOUTOFRANGE);
--- a/std/assembly/process.ts
+++ b/std/assembly/process.ts
@@ -131,9 +131,9 @@ abstract class WritableStream extends Stream {
 
 @unmanaged
 abstract class ReadableStream extends Stream {
-  read(buffer: ArrayBuffer, offset: usize = 0): i32 {
+  read(buffer: ArrayBuffer, offset: isize = 0): i32 {
     var end = <usize>buffer.byteLength;
-    if (offset > end) {
+    if (offset < 0 || <usize>offset > end) {
       throw new Error(E_INDEXOUTOFRANGE);
     }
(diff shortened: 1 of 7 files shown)


std/assembly/process.ts
```diff
@@ -131,7 +131,7 @@ abstract class WritableStream extends Stream {
 
 @unmanaged
 abstract class ReadableStream extends Stream {
-  read(buffer: ArrayBuffer, offset: isize = 0): i32 {
+  read(buffer: ArrayBuffer, offset: usize = 0): i32 {
```
dcodeIO, line 134: With this change, the range check below, in particular `offset < 0`, will never be `true`. This can also be written differently of course, with an `usize` here, and explicit casts to `isize` below, if you think that's clearer?

std/assembly/process.ts
```diff
@@ -131,7 +131,7 @@ abstract class WritableStream extends Stream {
...
     var end = <usize>buffer.byteLength;
     if (offset < 0 || offset > end) {
```
willemneal, line 136:
```suggestion
    if (offset > end) {
```

std/assembly/process.ts
willemneal: Where does it need to cast below?

dcodeIO: I'd assume you saw an error there, due to comparing an `isize` and an `usize` with `<`? If so, this can be fixed by keeping the code as-is, but changing the check below to `if (offset < 0 || <isize>offset > end)` I think. Or wasn't there a compilation error?

willemneal: I'm just confused why `offeset` is `isize` only to be used with `usize`'s.  For example, `<isize>offset > end` is what is causing the error in the original code since `end` is `usize`.  If there is motivation for this API then I can revert and just cast `offset` to `usize` -> `<usize>offset > end`.

dcodeIO: My expectation was that this API will be called with `i32`s most of the time, and upping that to `isize` was mostly an optimization to do the potential conversion exactly once when the function is called, instead of convert multiple times below. Other than in relational binary ops where there are two distinct Wasm ops to compare values (here: i32.lt_s and i32.lt_u), `isize` and `usize` are compatible.

std/assembly/process.ts
```diff
@@ -131,9 +131,9 @@ abstract class WritableStream extends Stream {
 
 @unmanaged
 abstract class ReadableStream extends Stream {
-  read(buffer: ArrayBuffer, offset: isize = 0): i32 {
+  read(buffer: ArrayBuffer, offset: usize = 0): i32 {
     var end = <usize>buffer.byteLength;
-    if (offset < 0 || offset > end) {
+    if (offset > end) {
```
willemneal, line 136:
```suggestion
    if (offset < 0 || <usize>offset > end) {
```

std/assembly/process.ts
```diff
@@ -131,9 +131,9 @@ abstract class WritableStream extends Stream {
 
 @unmanaged
 abstract class ReadableStream extends Stream {
-  read(buffer: ArrayBuffer, offset: isize = 0): i32 {
+  read(buffer: ArrayBuffer, offset: usize = 0): i32 {
```
willemneal, line 134:
```suggestion
  read(buffer: ArrayBuffer, offset: isize = 0): i32 {
```

dcodeIO:
The mechanism to enable certain features in the test suite is setting the `ASC_FEATURES` environment variable. One can set it to a comma separated list of features (see tests/features.json), or just `*` to enable all features. CI runs a subset of them in the "Features" job. If a feature is not enabled, affected tests are skipped fully, including when compiling to/updating fixtures.

willemneal:
@dcodeIO I added `std-wasi/process` and `std-wasi/crypto` to the features CI test.

dcodeIO:
Thank you! :)

PR #1792, opened 2021-04-07
add createCompoundNameType function

jtenner:
This pull request adds a simple function for creating a compound named type with type parameters using an array of `IdentifierExpression` objects

Question: How should we handle errors for this function?

### DIFF
--- a/src/ast.ts
+++ b/src/ast.ts
@@ -187,7 +187,7 @@ export abstract class Node {
     let result = node;
     for (let i = 1; i < length; i++) {
       let next = identifiers[i];
-      let nextNode = new TypeName(next, null, next.range)
+      let nextNode = new TypeName(next, null, next.range);
       node.next = nextNode;
       node = nextNode;
     }
--- a/src/ast.ts
+++ b/src/ast.ts
@@ -174,6 +174,30 @@ export abstract class Node {
     return new TypeParameterNode(name, extendsType, defaultType, range);
   }
 
+  static createCompoundNamedType(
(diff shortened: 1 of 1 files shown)


src/ast.ts
```diff
@@ -174,6 +174,30 @@ export abstract class Node {
...
   }
 
+  static createCompoundNamedType(
+    identifiers: IdentifierExpression[],
+    typeParameters: TypeNode[] | null,
+    isNullable: bool,
+    range: Range
+  ): NamedTypeNode {
+    let length = identifiers.length;
+    if (length == 0) throw new Error(); // how should errors be handled?
```
dcodeIO, line 184: If it was purely internal, I'd recommend an `assert`, but I think throwing is fine here since this is kinda a user-facing API when modifying the AST in transforms. Perhaps `Error("identifiers must not be empty")`.

src/ast.ts
```diff
@@ -174,6 +174,30 @@ export abstract class Node {
     return new TypeParameterNode(name, extendsType, defaultType, range);
   }
 
+  static createCompoundNamedType(
+    identifiers: IdentifierExpression[],
+    typeParameters: TypeNode[] | null,
+    isNullable: bool,
+    range: Range
+  ): NamedTypeNode {
+    let length = identifiers.length;
```
dcodeIO, line 183: nit: Top-level vars are `var` in compiler code, mostly historically and to aid the debugging the bootstrap, whereas `let` indicates a shared local in an inner block.

PR #1794, opened 2021-04-09
fix: pass arg to Duplicate_identifier_0 for markVirtuals

yjhmelody:
Now, the error message is not format normally.
```
ERROR TS2300: Duplicate identifier '{0}'.

     foo: number;
```


dcodeIO:
Thank you! :)

PR #1872, opened 2021-05-27
BREAKING CHANGE: Flatten generated TypeScript definitions

dcodeIO:
Generated TypeScript definitions for WebAssembly modules are now flat, that is do not `export default ASModule` anymore, but the module members directly, with the `ASModule` namespace removed.

fixes #1705

### DIFF
--- a/src/definitions.ts
+++ b/src/definitions.ts
@@ -605,33 +605,28 @@ export class TSDBuilder extends ExportsWalker {
   build(): string {
     var sb = this.sb;
     var isWasm64 = this.program.options.isWasm64;
-    sb.push("declare module ASModule {\n");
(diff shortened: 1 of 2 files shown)


JerryGreen:
It’s better to merge parent PR (https://github.com/AssemblyScript/assemblyscript/pull/1705) because in addition to fixing how `.d.ts` is generated, it also:

1. Has tests
2. Actually runs `tsc`, so in cases where definitions are not found by some reason, maybe due to changes to AssemblyScript compiler, or to TypeScript compiler, a test will fail
3. Also runs some simple code after importing in tests, checking both static and dynamic types of imports, making sure it can find related functions using the flat structure

PR #1894, opened 2021-06-09
Fixed stdlib builtins not working with typedoc

torch2424:
Hello!

This is just something I noticed when working on the [Fastly AS SDK](https://www.npmjs.com/package/@fastly/as-compute).

It seems like in version `0.18.13`, we introduced a change in which caused a type error, as we try to assign `void` to an `i32`, and `i31ref`:

Screenshot from 2021-06-08 17-19-44

I got the error above when running typedoc on version `0.19.1`, but it works again on `0.18.12`. So, I assume unreachable is the instruction that causes things to crash. So instead of returning void, I still call the function, but return a "void number" (0), which I assume is the expected behavior?

With this fix, things start working again in the type system:

Screenshot from 2021-06-08 17-26-09

Thanks to whoever picks up this review! 😄 🎉 

cc @dcodeIO @MaxGraey 😄

### DIFF
--- a/std/assembly/builtins.ts
+++ b/std/assembly/builtins.ts
@@ -2302,11 +2302,17 @@ export abstract class i31 { // FIXME: usage of 'new' requires a class :(
 
   // @ts-ignore: decorator
   @builtin
-  static new(value: i32): i31ref { return unreachable(); }
+  static new(value: i32): i31ref { 
+    unreachable();
+    return 0;
+  }
 
   // @ts-ignore: decorator
   @builtin
-  static get(i31expr: i31ref): i32 { return unreachable(); }
+  static get(i31expr: i31ref): i32 { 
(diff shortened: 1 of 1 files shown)


MaxGraey (approved):
Thanks!

dcodeIO:
Hmm, interesting. Unreachable is currently typed as

```ts
declare function unreachable(): never;
```

and in AS itself is an expression of any type. I guess the following doesn't work either?

```ts
static new(value: i32): i31ref { return <i31ref>unreachable(); }

static get(i31expr: i31ref): i32 { return <i32>unreachable(); }
```

dcodeIO (approved):

torch2424:
Thanks for the review y'all @dcodeIO @MaxGraey ! Merging! 😄

PR #1898, opened 2021-06-09
initial tests for arguments length

jtenner:
⯈ Initial attempt on implementing `arguments.length`


src/builtins.ts
```diff
@@ -705,6 +705,8 @@ export namespace BuiltinNames {
   export const wasiAbort = "~lib/wasi/index/abort";
   export const wasiTrace = "~lib/wasi/index/trace";
   export const wasiSeed = "~lib/wasi/index/seed";
+
```
willemneal, line 708:
```suggestion
```

std/assembly/builtins.ts
```diff
@@ -2336,4 +2336,11 @@ declare function trace(
 @external("env", "seed")
 declare function seed(): f64;
 
-/* eslint-enable @typescript-eslint/no-unused-vars */
+@final
+export declare class arguments {
```
willemneal, line 2340: Is it possible to declare a namespace?  If not, that should be an added feature request.

jtenner:
```lisp
(func $builtins/arglength1 (param $0 i32)
 (local $1 i32)
 (local $2 i32)
 global.get $~lib/memory/__stack_pointer
 i32.const 4
 i32.sub
 global.set $~lib/memory/__stack_pointer
 call $~stack_check
 global.get $~lib/memory/__stack_pointer
 i32.const 0
 i32.store
+  global.get $~argumentsLength
+  local.set $1
 i32.const 1024
 local.set $2
 global.get $~lib/memory/__stack_pointer
 local.get $2
 i32.store
 local.get $2
 i32.const 2
 + call $~lib/builtins/arguments.get:length
 f64.convert_i32_s
 + call $~lib/builtins/arguments.get:length
 f64.convert_i32_s
 f64.const 0
 f64.const 0
 f64.const 0
 call $~lib/builtins/trace
 global.get $~lib/memory/__stack_pointer
 i32.const 4
 i32.add
 global.set $~lib/memory/__stack_pointer
)
```

I put a plus sign before the important bits. Looks like it's calling a function instead of getting a local.

jtenner:
Of course, it adds a function like this:

```wat
+ (func $~lib/builtins/arguments.get:length (result i32)
+  (local $0 i32)
+  global.get $~argumentsLength
+  local.set $0
+  local.get $0
+ )
```

🙄

src/builtins.ts
```diff
@@ -9421,6 +9424,18 @@ function builtin_f64x2_ge(ctx: BuiltinContext): ExpressionRef {
 }
 builtins.set(BuiltinNames.f64x2_ge, builtin_f64x2_ge);
 
+// access arguments.length
+function builtin_arguments_length(ctx: BuiltinContext): ExpressionRef {
+  const compiler = ctx.compiler;
+  const mod = compiler.module;
+  compiler.currentType = Type.i32;
+  const local = compiler.currentFlow.lookupLocal(BuiltinNames.argumentsLengthLocal)!;
+  return mod.local_get(local.index, TypeRef.I32);
```
jtenner, line 9432: This becomes the function body of the builtin. Using `arguments.length` accesses the namespace, and calls the function. This is not desired behavior.

I believe we need to override `compilePropertyAccess()`

HerrCai0907:
clean outdated draft PR

PR #1903, opened 2021-06-13
[WIP] implement for..of loops and iterators

pitust:

This PR implements for..of loops as well as iterators, which are important (convenience) features when writing code in assemblyscript.

The optimal solution would be to use `Symbol.iterator`, however symbols (and the api itself) are not yet implemented by the compiler.

Instead, this PR uses a different approach: a new decorator, `@iterator`.

TODO:
 - [ ] Add tests
 - [x] Run `npm check` and `npm test` to make sure the code is fine

### DIFF
--- a/src/compiler.ts
+++ b/src/compiler.ts
@@ -2681,12 +2681,85 @@ export class Compiler extends DiagnosticEmitter {
   private compileForOfStatement(
     statement: ForOfStatement
   ): ExpressionRef {
-    this.error(
-      DiagnosticCode.Not_implemented_0,
-      statement.range,
-      "Iterators"
-    );
-    return this.module.unreachable();
+    const compiler = this
+    function earlyBailout(ty: Type) {
+      compiler.error(
+        DiagnosticCode.Type_0_is_not_iterable,
+        statement.range,
+        ty.toString()
+      );
+      return compiler.module.unreachable();
+    }
+    if (statement.variable.kind != NodeKind.VARIABLE) {
+      this.error(
+        DiagnosticCode.Not_implemented_0,
+        statement.range,
+        "For .. of where the target is not a variable"
+      );
+      return this.module.unreachable();
+    }
+    const variable = <VariableStatement>statement.variable
+    if (variable.declarations.length != 1) {
+      this.error(
+        DiagnosticCode.Only_a_single_variable_declaration_is_allowed_in_a_for_of_statement,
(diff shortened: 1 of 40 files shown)


pitust:
I'm getting some memory leaks, but they are unrelated to the added code - they exist even when writing similar code by hand somehow.

src/compiler.ts
```diff
@@ -2681,12 +2681,87 @@ export class Compiler extends DiagnosticEmitter {
...
+    const iteratorcls = iterator.getClass();
+    if (!iteratorcls) return earlyBailout(iterablety);
+    
+    const nextelem = iteratorcls.getMethod('next');
+    if (!nextelem) return earlyBailout(iterablety);
+
+    const doneelem = iteratorcls.getMethod('done');
+    if (!doneelem) return this.module.unreachable();
+    
+    const iteratorvaluetype = (<Function>nextelem).signature.returnType;
```
dcodeIO, line 2731: The cast here (and some below) can fail when `nextelem` (or `doneelem`) is not a `Function`, which would lead to a compiler crash. To mitigate, the `kind` of the element can be checked explicitly, and a diagnostic produced if it is not of the required type or shape. With shape I mean that there also needs to be a signature check that calling the methods above will actually produce valid WebAssembly code, plus making sure that the methods are indeed compiled before being called.

src/program.ts
```diff
@@ -4353,6 +4357,25 @@ export class Class extends TypedElement {
     return null;
   }
 
+  /** Looks up the iterator for this class. */
+  lookupIterator(): Function | null {
+    var instance: Class | null = this;
+    do {
+      if (instance.members) {
+        for (let member of instance.members.values()) {
```
dcodeIO, line 4365: Unlike TypeScript, which is unsound in this case, AssemblyScript cannot track that `instance.members` is non-null after the check for type safety reasons. Assigning `let members = instance.members` to a local, and then doing the check, solves this.

dcodeIO (commented):
Just a few early pointers that should help getting the PR pass CI eventually :)

Other than that, to set realistic expectations, I'd like to raise awareness that iterators can become complex when it comes to concurrent modification and data types that are not as trivial as arrays. If there is interest in working this out in detail, including Symbol properties and ECMAScript behavior, that would be pretty cool of course! Otherwise, however, the PR may not make it.

PR #1949, opened 2021-07-02
fix(portable): avoid duplicated adding wrap function

### LINKED ISSUE #1948: Always meet error `TypeError: Cannot redefine property: wrap` in CI
I writen some jest unit tests, but the first case is fine but the later cases are all failed because `Cannot redefine property: wrap`:

```
FAIL src/tests/env.spec.ts
  ● Test suite failed to run

    TypeError: Cannot redefine property: wrap
        at Function.defineProperty (<anonymous>)

      at Object.call (../../node_modules/assemblyscript/dist/webpack:/assemblyscript/std/portable/index.js:384:8)
      at i (../../node_modules/assemblyscript/dist/webpack:/assemblyscript/webpack/bootstrap:19:32)
      at Object.call (../../node_modules/assemblyscript/dist/webpack:/assemblyscript/src/glue/js/index.ts:6:1)
      at __webpack_require__ (../../node_modules/assemblyscript/dist/webpack:/assemblyscript/webpack/bootstrap:19:32)
      at ../../node_modules/assemblyscript/dist/webpack:/assemblyscript/webpack/startup:3:1
      at factory (../../node_modules/assemblyscript/dist/assemblyscript.js:7:758416)
      at ../../node_modules/assemblyscript/dist/webpack:/assemblyscript/webpack/universalModuleDefinition:3:20
      at Object.<anonymous> (../../node_modules/assemblyscript/dist/webpack:/assemblyscript/webpack/universalModuleDefinition:1:1)
```
I don't know why my local code have not met such errors. But I cannot avoid this error in github CI.
Maybe we should add `if` before `Object.defineProperty` for `portable/index.js`:

```
Object.defineProperty(Int16Array, "wrap", {
  value: function wrap(buffer, byteOffset, length) {
    return new Int16Array(buffer, byteOffset, length);
  }
});
```

yjhmelody:
Try to fix #1948

### DIFF
--- a/std/portable/index.js
+++ b/std/portable/index.js
@@ -375,56 +375,24 @@ globalScope["trace"] = function(message, n) {
   console.error("trace: " + message);
 };
 
-Object.defineProperty(Int8Array, "wrap", {
-  value: function wrap(buffer, byteOffset, length) {
-    return new Int8Array(buffer, byteOffset, length);
-  }
-});
+var wrapTypes = [
+  Int8Array,
+  Uint8Array,
+  Uint8ClampedArray,
+  Int16Array,
+  Uint16Array,
+  Int32Array,
+  Uint32Array,
+  Float32Array,
+  Float64Array
+];
 
-Object.defineProperty(Uint8Array, "wrap", {
-  value: function wrap(buffer, byteOffset, length) {
-    return new Uint8Array(buffer, byteOffset, length);
+for(const type of wrapTypes) {
+  if(!type.wrap) {
+    Object.defineProperty(type, "wrap", {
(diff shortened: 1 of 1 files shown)


yjhmelody:
Another strange point is that there is no `wrap` function provided by Uint64Array/Int64Array.

yjhmelody:
@MaxGraey Hi

dcodeIO:
This looks like it should be fixed by https://github.com/AssemblyScript/assemblyscript/pull/2049, where now all the glue code is only executed once. Can someone affected by the original issue confirm? :)

PR #1967, opened 2021-07-11
Statically eliminate logical expression branches

dcodeIO:
As brought up in Discord, we currently do not statically eliminate expressions like `isNullable<T>() && value === null`, which may be surprising since we do this for normal `if`s. This PR also applies it for logical expressions when interested in `bool` returns. Question is: Is this always safe?

### DIFF
--- a/src/compiler.ts
+++ b/src/compiler.ts
@@ -10057,17 +10057,15 @@ export class Compiler extends DiagnosticEmitter {
     return supported;
   }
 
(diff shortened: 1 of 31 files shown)


SebastianSpeitel:
Looking at the changes I already like how much unnecessary code was removed.

MaxGraey (approved):
LGTM

Issue #1975, opened 2021-07-13
Unable to integrate the code of another transform package


yjhmelody:
I need to reuse a transform for another transform. But I meet the following errors, it's about `redefine` problem.
```
FAIL  src/tests/debug.spec.ts
  ● Test suite failed to run

    TypeError: Cannot redefine property: signbit
        at Function.defineProperties (<anonymous>)

      at Object.call (../../node_modules/as-debug-transform/node_modules/assemblyscript/dist/webpack:/assemblyscript/std/portable/index.js:350:8)
      at i (../../node_modules/as-debug-transform/node_modules/assemblyscript/dist/webpack:/assemblyscript/webpack/bootstrap:19:32)
      at Object.call (../../node_modules/as-debug-transform/node_modules/assemblyscript/dist/webpack:/assemblyscript/src/glue/js/index.ts:6:1)
      at __webpack_require__ (../../node_modules/as-debug-transform/node_modules/assemblyscript/dist/webpack:/assemblyscript/webpack/bootstrap:19:32)
      at ../../node_modules/as-debug-transform/node_modules/assemblyscript/dist/webpack:/assemblyscript/webpack/startup:3:1
      at factory (../../node_modules/as-debug-transform/node_modules/assemblyscript/dist/assemblyscript.js:7:758432)
      at ../../node_modules/as-debug-transform/node_modules/assemblyscript/dist/webpack:/assemblyscript/webpack/universalModuleDefinition:3:20
      at Object.<anonymous> (../../node_modules/as-debug-transform/node_modules/assemblyscript/dist/webpack:/assemblyscript/webpack/universalModuleDefinition:1:1)
```

yjhmelody:
pin the issue for #1977

dcodeIO:
Tackled in https://github.com/AssemblyScript/assemblyscript/pull/2049, now logging a warning on mismatch. Note that this fixes the redeclare problem, but not the underlying cause which comes from building against multiple potentially incompatible compiler versions. The proper fix remains to always depend on the same compiler installation everywhere.

PR #1977, opened 2021-07-13
Execute portable glue only once?

dcodeIO:
This is a (not very good) attempt to address issues like #1975 and #1948, where two different versions of the portable glue are mistakenly being included. It doesn't really fix the base problem, because one will most likely also include multiple versions of the compiler which will explode another way. But perhaps it depends on exact use case whether it works or not?

### DIFF
--- a/src/compiler.ts
+++ b/src/compiler.ts
@@ -10532,14 +10532,31 @@ export class Compiler extends DiagnosticEmitter {
     var temp = flow.getTempLocal(type);
     var instanceofInstance = this.program.instanceofInstance;
     assert(this.compileFunction(instanceofInstance));
-    expr = module.if(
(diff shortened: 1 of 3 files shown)


willemneal:
While this will certainly be a solution the error, I still think it will promote transforms with various different versions of the complier. Though perhaps the transform author has a reason for this?

I propose that in addition to this change, the transform interface is updated so that each transform is passed a reference to `assemblyscript`.  This way there is no need for any transform to have the complier as a dependency and will ensure that there is only one version present.  However, with your change the transform author can import a particular version of the compiler and not have it blow up.

Wdyt?

dcodeIO:
Whoops, tried force-pushing a rebase but that somehow closed it. See new PR above.

PR #1979, opened 2021-07-13
feat(CLI): ignore a specific warning

willemneal:
`--ignoreWarning` will suppress a specific warning or list of warnings.

This is useful in cases where the warning is known by the author and is okay. For example, in my work on adding a generic encoder interface I add generic methods to a class and if that class isn't used I get 232, which can be ignored.

- [x]  I've read the contributing guidelines


willemneal:
Some more context for my use case, testing.  I only want the warning to go away during the test but not production builds.

PR #1998, opened 2021-07-21
fix(astBuilder): fix export statment

yjhmelody:
Now `export * from "xx"` will be built to `export {} from "xx"`.

@willemneal  visitor-as also has this problem.

It makes the behavior of the compiler plug-in difficult to understand, because there may be many files that are modified.

### DIFF
--- a/src/extra/ast.ts
+++ b/src/extra/ast.ts
@@ -1071,7 +1071,9 @@ export class ASTBuilder {
       sb.push("declare ");
     }
     var members = node.members;
-    if (members != null && members.length > 0) {
+    if (members == null) {
+      sb.push("export *");
+    } else if (members.length > 0) {
       let numMembers = members.length;
       sb.push("export {\n");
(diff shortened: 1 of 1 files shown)


yjhmelody:
@willemneal Can you take a look?

willemneal:
I'm still unsure what the issue is here.  Is it that an AST from `export * from ""` when turned back into a string became `export {} from ""` and now it will stay the same?

yjhmelody:
> I'm still unsure what the issue is here. Is it that an AST from `export * from ""` when turned back into a string became `export {} from ""` and now it will stay the same?

Yes, ASTbuilder does not handle the situation when the members of ExportStatement are null

willemneal (approved):
Okay makes sense to me.

yjhmelody:
@MaxGraey Hi, anything wrong ?

MaxGraey (approved):

PR #2042, opened 2021-08-20
Properly handle nulls in runtime upcast checks

### LINKED ISSUE #2035: Upcast check is missing a null check when casting to a nullable type
Code like

```ts
class Foo {}
class Bar extends Foo {}
var foo: Foo | null = null;
foo instanceof Bar | null // here
```

is missing a check for when the LHS value is `0`, directly calling the `__instanceof` helper which then can fail due to UB.

dcodeIO:
Fixes #2035 by bypassing `null` values in emitted runtime upcast checks (which `__instanceof` naturally cannot handle) while keeping the current behavior as a simplification where the value can be statically determined to be non-null.

### DIFF
--- a/src/compiler.ts
+++ b/src/compiler.ts
@@ -10565,14 +10565,31 @@ export class Compiler extends DiagnosticEmitter {
     var temp = flow.getTempLocal(type);
     var instanceofInstance = this.program.instanceofInstance;
     assert(this.compileFunction(instanceofInstance));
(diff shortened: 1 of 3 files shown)


dcodeIO:
cc @trusktr Can you confirm that this fixes the issue you are seeing?

trusktr:
I need to try this. I have TODO comments in place.

PR #2084, opened 2021-09-26
fix: rewrite __visit of class only effect in ~/lib

HerrCai0907:
⯈ When use @unsafe __visit as class member function, the way of visiting class during GC will be changed
⯈ Consider inherit class. If inherit class override __visit, both will be used. 


src/program.ts
```diff
@@ -4533,6 +4533,25 @@ export class Class extends TypedElement {
...
+    if (visitPrototype == null) {
+      return null;
+    }
+    if (this.isDeclaredInLibrary) {
+      return visitPrototype;
+    }
+    const decorator = visitPrototype.declaration.decorators;
+    if (!decorator || (decorator && decorator.every((v) => v.decoratorKind != DecoratorKind.UNSAFE))) {
+      return null;
+    }
```
MaxGraey, line 4548: As I understand currently if `@unsafe` doesn't exists than rewriting of `__visit` silently doesn't happening. Will be great add new diagnostic error message that would describe that this operation is unsafe and requires the explicit addition of the "@unsafe" decorator. And the user has to decide whether to add it or go some other way

HerrCai0907: In old logic, we can only rewrite `__visit` in `~/lib`. 
What I want to do is using `@unsafe` to enable rewriting `__visit` if this class is in the outside. For example, some class in assembly/index.ts

MaxGraey: Yes, I understand this. I'm suggesting also fix current behaviour when you try to do the same but without `@unsafe` decorator (in main branch it just silently ignored)

src/diagnosticMessages.generated.ts
```diff
@@ -366,6 +367,7 @@ export function diagnosticCodeToString(code: DiagnosticCode): string {
     case 6054: return "File '{0}' not found.";
     case 6188: return "Numeric separators are not allowed here.";
     case 6189: return "Multiple consecutive numeric separators are not permitted.";
+    case 6190: return "override '__visit' need @unsafe decorator";
```
romdotdog, line 370:
```suggestion
    case 6190: return "Overriding '__visit' needs an @unsafe decorator.";
```

src/diagnosticMessages.json
```diff
@@ -181,6 +181,7 @@
   "File '{0}' not found.": 6054,
   "Numeric separators are not allowed here.": 6188,
   "Multiple consecutive numeric separators are not permitted.": 6189,
+  "override '__visit' need @unsafe decorator": 6190,
```
romdotdog, line 184:
```suggestion
  "Overriding '__visit' needs an @unsafe decorator.": 6190,
```

src/builtins.ts
```diff
@@ -9538,34 +9538,33 @@ function ensureVisitMembersOf(compiler: Compiler, instance: Class): void {
 
   // Some standard library components provide a custom visitor implementation,
   // for example to visit all members of a collection, e.g. arrays and maps.
+  // if __visit with unsafe decorder, it alse works.
```
romdotdog, line 9541:
```suggestion
  // __visit can be overridden with the @unsafe decorator
```

src/diagnosticMessages.generated.ts
```diff
@@ -182,6 +182,7 @@ export enum DiagnosticCode {
   File_0_not_found = 6054,
   Numeric_separators_are_not_allowed_here = 6188,
   Multiple_consecutive_numeric_separators_are_not_permitted = 6189,
+  override___visit_need_unsafe_decorator = 6190,
```
romdotdog, line 185:
```suggestion
  Overriding___visit_needs_an_unsafe_decorator = 6190,
```

src/program.ts
```diff
@@ -4533,6 +4534,26 @@ export class Class extends TypedElement {
...
+    const visitPrototype = this.getMember("__visit");
+    if (visitPrototype == null) {
+      return null;
+    }
+    if (this.isDeclaredInLibrary) {
+      return visitPrototype;
+    }
+    const decorator = visitPrototype.declaration.decorators;
+    if (!decorator || (decorator && decorator.every((v) => v.decoratorKind != DecoratorKind.UNSAFE))) {
+      compiler.warning(DiagnosticCode.override___visit_need_unsafe_decorator,visitPrototype.declaration.name.range);
```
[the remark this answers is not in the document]
MaxGraey: And I think it should be an error, not warning. @dcodeIO wdyt?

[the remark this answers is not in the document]
dcodeIO: If the compiler ignores it when the decorator is not present, I think a warning is sufficient so one could theoretically use the name `__visit` otherwise. Seems unlikely, but probably does not harm to be a little more conservative about it.

dcodeIO:
I'm wondering if the mechanism could utilize a new decorator, say `@gcvisit`, with the name of the method becoming irrelevant. This would be a breaking change, but I think changing the behavior as proposed is one already anyhow, at least in principle.

src/program.ts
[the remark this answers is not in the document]
MaxGraey: I don't think the user will want to use `__visit` specifically. It's more likely that it's the visiting GC roots that he wants to owerride. But he may not notice the warning, especially if there are a lot of such warnings, but concerning some other unimportant things. That's why I think it's better to make it an error

MaxGraey:
Yeah. Also good idea. Just what bout `@gc.visitor`? If we have namespaced decorator we could group more things and with autocomplete / suggestions.

dcodeIO:
I expect this to go away with Wasm GC again anyway, and checking for nested expressions in decorators is a little awkward atm. Not sure if worth it. The same argument can of course be made about not introducing a decorator for this at all, hmm. Perhaps just keeping `__visit` isn't so bad, feels legacy, which is kinda appropriate.

MaxGraey:
yeah, makes sense keep backward comp in this case

HerrCai0907:
I think namespaced decorator may be not necessary because editor such as vscode can autocomplete decoder pretty good now. Using a common prefix is more easy than add namespace for parser, don't need to add new grammar and only a new enumerate.

PR #2096, opened 2021-10-14
fix: transform cannot find module

mooooooi:
Environment:
assemblyscript: 0.19.17
platform: macos
cmd: yarn asbuild
asconfig: 
```js
{
    // ...someting not important
    options: {
        "transform": "./transform.ts"
    }
}
```

The segment of options.js
```javascript
function resolvePath(p, baseDir, useNodeResolution = false) {
  if (path.isAbsolute(p)) return p;
  if (useNodeResolution && !p.startsWith(".")) {
    return dynrequire.resolve(p, { paths: [ baseDir ] });
  }
  return path.join(baseDir, p);
}
```
The default value of `baseDir` is `.`. When p is relative path, then ...
```typescript
path.join(".", "./transform.ts"); // The result is 'transform.ts'. It lost the information about relative
```

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -43,6 +43,7 @@ under the licensing terms detailed in LICENSE:
 * Joe Pea <trusktr@gmail.com>
 * Felipe Gasper <FGasper@users.noreply.github.com>
 * Congcong Cai <77575210+HerrCai0907@users.noreply.github.com>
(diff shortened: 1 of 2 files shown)


dcodeIO (approved):

dcodeIO:
Thank you! :)

Issue #2101, opened 2021-10-22
do-while + switch + continue cause compiler crash


HerrCai0907:
It cause compiler crash
```ts
export function foo(): void {
    let a = 9;
    do {
        switch (a) {
            case 0:
                continue;
        }
    } while (a < 3);
}
```

[wasm-validator error in function assembly/index/foo] unexpected false: all break targets must be valid, on 
(br $do-continue|0)
FAILURE validate error

PR #2102, opened 2021-10-22
fix: compileSwitchStatement will hide inner conditional flags

HerrCai0907:
⯈fix issue: do-while + switch + continue cause compiler crash #2101
⯈when `compileSwitchStatement`, `commonConditional` don't consider the situations that `CONTINUES` should change to `CONDITIONALLY_CONTINUES` in `FlowFlags`.

### DIFF
--- a/src/compiler.ts
+++ b/src/compiler.ts
@@ -2970,7 +2970,7 @@ export class Compiler extends DiagnosticEmitter {
         commonCategorical &= innerFlow.flags;
       }
 
-      commonConditional |= innerFlow.getConditionFlags();
+      commonConditional |= innerFlow.deriveConditionalFlags();
 
       // Switch back to the parent flow
       innerFlow.unset(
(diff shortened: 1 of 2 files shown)


src/flow.ts
```diff
@@ -266,6 +266,26 @@ export class Flow {
   /** Unsets the specified flag or flags. */
   unset(flag: FlowFlags): void { this.flags &= ~flag; }
 
+  getConditionFlags(): FlowFlags {
```
dcodeIO, line 269:
```suggestion
  deriveConditionalFlags(): FlowFlags {
```

src/compiler.ts
```diff
@@ -2969,7 +2969,8 @@ export class Compiler extends DiagnosticEmitter {
       if (terminates || isLast || innerFlow.isAny(FlowFlags.BREAKS | FlowFlags.CONDITIONALLY_BREAKS)) {
         commonCategorical &= innerFlow.flags;
       }
-      commonConditional |= innerFlow.flags & FlowFlags.ANY_CONDITIONAL;
+
+      commonConditional |= innerFlow.getConditionFlags();
```
dcodeIO, line 2973:
```suggestion
      commonConditional |= innerFlow.deriveConditionalFlags();
```

dcodeIO (approved):
It seems that there'll be more work necessary to get this perfect, but since this fixes the reported problem I think it's fine to merge for now :)

Just a nit with the naming, which I can apply myself.

dcodeIO:
Thanks! :)

PR #2129, opened 2021-11-09
Fix Binaryen glue string encoding

dcodeIO:
Fixes some old code inherited from Emscripten around the initial commit, that is [not handling surrogates correctly](https://github.com/emscripten-core/emscripten/issues/15324). Did not surface so far because it is only used to call Binaryen APIs with constant names respectively identifiers coming from the parser. For now retains the assumption that WTF-8 is fine with Binaryen/C++ (`const char*`, `std::string`, ...).


MaxGraey (approved):

dcodeIO:
Merging this for now until more is known.

PR #2158, opened 2021-11-28
Allow duplicate fields

romdotdog:

## What's this?
This PR replaces the "duplicate identifier" diagnostic used in the `markVirtuals` function (which also doesn't catch private duplicate fields) with a few checks in the `finishResolveClass` function for visibility and assignability, conforming to TS. ~~I also implemented ambient fields, or fields with the `declare` keyword.~~

## What does this allow?
This allows, in general, duplicate fields, but also allows duplicate fields that are assignable to the respective field in the base class
```ts
class A {}
class B extends A {}

class A1 {
  constructor(public a: A) {}
}

class A2 extends A1 {
  constructor(public a: B) {
    super(a)
  }
}
```

In the end, this feature was implemented in order to allow the `VariableDeclaration` class to upgrade its `name` field from an `IdentifierExpression` to an `Expression`; this allows array destructuring to be implemented cleanly.


src/resolver.ts
```diff
@@ -3092,17 +3092,18 @@ export class Resolver extends DiagnosticEmitter {
             let fieldPrototype = <FieldPrototype>member;
             let fieldTypeNode = fieldPrototype.typeNode;
             let fieldType: Type | null = null;
-            // TODO: handle duplicate non-private fields specifically?
+            let existingField: Field | null = null;
+            if (base) {
+              let baseMembers = base.members;
+              if (baseMembers !== null && baseMembers.has(fieldPrototype.name)) {
+                let baseField = assert(baseMembers.get(fieldPrototype.name));
+                assert(baseField.kind == ElementKind.FIELD);
```
dcodeIO, line 3100: Doesn't seem to be guaranteed, could as well be a method for example and then crash the compiler.

src/resolver.ts
```diff
@@ -3130,6 +3131,46 @@ export class Resolver extends DiagnosticEmitter {
...
+                  fieldPrototype.name, baseClass.internalName
+                );
+              } else if (!fieldType.isStrictlyAssignableTo(existingField.type)) {
+                this.errorRelated(
+                  DiagnosticCode.Property_0_in_type_1_is_not_assignable_to_the_same_property_in_base_type_2,
+                  fieldPrototype.identifierNode.range, existingField.identifierNode.range,
+                  fieldPrototype.name, instance.internalName, baseClass.internalName
+                );
+              }
+            }
```
dcodeIO, line 3173: Appears that below here, a new field with the same name and a different memory offset is created, but I think the memory offset should be reused if it is compatible?

src/parser.ts
```diff
@@ -1870,6 +1871,39 @@ export class Parser extends DiagnosticEmitter {
...
+          }
+        }
+        if (!startPos) startPos = tn.tokenPos;
+      } else { // identifier
+        tn.reset(state);
+      }
+    } else if (contextIsAmbient) {
+      flags |= CommonFlags.AMBIENT;
+    }
+
```
dcodeIO, line 1906: I'd prefer to make the `declare` changes a separate PR, since `declare`d class members would yield imports and it is not clear to me how these should work, in particular for fields.

dcodeIO (commented):
A few observations :)

src/resolver.ts
romdotdog: that's what I thought, but in the tests it appears as if the class has the proper size.. maybe I'm using the wrong functions in the test?

src/parser.ts
romdotdog: should I change the respective error message? (the one which mentions the declare keyword)

src/resolver.ts
dcodeIO: The way to test this would be
```ts
class A {
  public foo: i32;
}
class B extends A {
  public foo: i32;
}
assert(offsetof<A>("foo") == 0);
assert(offsetof<B>("foo") == 0); // this one should fail if my assumption above is correct
```

src/parser.ts
dcodeIO: There is not really a mechanism currently to make a `declare`d class member work in the first place, I think. What I imagine here is that this can still become a separate PR, so the parser understands the syntax, but would produce an error that ambient class members are not supported yet?

src/resolver.ts
romdotdog: Huh, I thought I could keep that assert in there because it was in the old code, and from my testing I wasn't able to crash the compiler pre-PR. I'll change that right away

dcodeIO: Makes me wonder if this is checked earlier already, when the class is initialized? Could that be?

src/resolver.ts
romdotdog: > if it is compatible

this just means if it's assignable right? or is there another check I should put in?

dcodeIO: Right, the sub type must be strictly assignable to the super type, that is something like `i32` that can technically assign to `i64` must be disallowed because it's not the right width in memory. Iirc you used `isStrictlyAssignable` there, which should be correct.

src/resolver.ts
```diff
@@ -3130,12 +3138,56 @@ export class Resolver extends DiagnosticEmitter {
               }
             }
             if (!fieldType) break; // did report above
+            if (existingField !== null) {
+              // visibility checks
+              let thisFieldIsPublic = fieldPrototype.isPublic;
+              let existingFieldIsPublic = existingField.isPublic;
```
dcodeIO, line 3144: There is also the case of:

```ts
class A {
  protected foo: i32;
}
class B extends A {
  [private | public | protected] foo: i32;
}
```

Since this is inheritance here, I guess `protected` can be treated like `public`, except that a `protected` field cannot shadow a `public` field?

src/resolver.ts
```diff
@@ -3130,12 +3138,62 @@ export class Resolver extends DiagnosticEmitter {
               }
             }
             if (!fieldType) break; // did report above
+            if (existingField !== null) {
+              // visibility checks
+              let baseClass = <Class>base;
+              let thisFieldIsPubOrProt = !fieldPrototype.is(CommonFlags.PRIVATE);
+              let existingFieldIsPublic = existingField.isPublic;
+
+              if (thisFieldIsPubOrProt && !existingFieldIsPublic) {
```
dcodeIO, line 3147: This seems to also catch `protected < protected`, which I guess would be OK?

src/resolver.ts
```diff
@@ -3130,12 +3138,62 @@ export class Resolver extends DiagnosticEmitter {
...
+                  fieldPrototype.identifierNode.range, existingField.identifierNode.range,
+                  fieldPrototype.name, baseClass.internalName, instance.internalName
+                );
+              } else if (fieldPrototype.is(CommonFlags.PROTECTED) && existingFieldIsPublic) {
+                this.errorRelated(
+                  DiagnosticCode.Property_0_is_protected_in_type_1_but_public_in_type_2,
+                  fieldPrototype.identifierNode.range, existingField.identifierNode.range,
+                  fieldPrototype.name, instance.internalName, baseClass.internalName
+                );
+              } else if (!thisFieldIsPubOrProt && existingFieldIsPublic) {
```
dcodeIO, line 3159: There seems to be a missing case where `private < protected`. I might be missing something, though, but in general I'd prefer if these checks were a little easier to grasp. Seems the rules here are:

* If existing field is private, it cannot be redeclared at all
* If existing field is protected, it can be redeclared as either protected or public
* If existing field is public, it can only be redeclared as public
* Plus: If existing field has an initializer but the redeclared one has not, that's an error

(that correct?)

src/resolver.ts
romdotdog: Argh, I forgot to add a test for that! 😅

src/resolver.ts
romdotdog: You're completely right, I'll make the checks a bit easier to grasp

> Plus: If existing field has an initializer but the redeclared one has not, that's an error

This doesn't throw an error:
```ts
class A {
  public a: number = 1;
  constructor(a: number) { this.a = 2; }
}

class B extends A {
  constructor(public a: number) { super(a) }
}
```
perhaps you're referring to if the redeclared field is redundant but has an initializer, no error is thrown?

dcodeIO: In particular I was wondering about the diagnostic

> Property_0_will_overwrite_the_base_property_in_1_If_this_is_intentional_add_an_initializer_Otherwise_remove_the_redundant_declaration

which seems to refer to the case

```ts
class A {
  public foo = 1;
}
class B extends A {
  public foo;
}
```

while if it is a ctor parameter as well then an initializer is kinda implied. (but I have actually no idea)

dcodeIO: The two cases can be told apart btw by looking at `Field.parameterIndex`. If it is `-1` it is just a field (as in my snippet), otherwise it is a field that is also a constructor parameter at the specified index.

romdotdog: agreed to scrap the `Property_0_will_overwrite_the_base_property_in_1_If_this_is_intentional_add_an_initializer_Otherwise_remove_the_redundant_declaration` error

src/resolver.ts
```diff
@@ -3130,12 +3138,83 @@ export class Resolver extends DiagnosticEmitter {
...
+              } else {
+                // fieldPrototype is public here
+                if (existingField.is(CommonFlags.PRIVATE)) {
+                  this.errorRelated(
+                    DiagnosticCode.Property_0_is_private_in_type_1_but_not_in_type_2,
+                    fieldPrototype.identifierNode.range, existingField.identifierNode.range,
+                    fieldPrototype.name, baseClass.internalName, instance.internalName
+                  );
+                }
+              }
```
dcodeIO, line 3197: Wondering: Would this become shorter when checking for existingField first, since when it is private, that's a common case for all essentially?

src/resolver.ts
```diff
@@ -3130,12 +3138,83 @@ export class Resolver extends DiagnosticEmitter {
...
+            }
             let fieldInstance = new Field(fieldPrototype, instance, fieldType);
             assert(isPowerOf2(fieldType.byteSize));
-            let mask = fieldType.byteSize - 1;
-            if (memoryOffset & mask) memoryOffset = (memoryOffset | mask) + 1;
-            fieldInstance.memoryOffset = memoryOffset;
-            memoryOffset += fieldType.byteSize;
+            if (existingField !== null) {
+              fieldInstance.memoryOffset = existingField.memoryOffset;
```
dcodeIO, line 3211: I have a vague suspicion that this can lead to issues, since typically inherited fields share the same flow flags for example, which is not the case anymore when duplicating the field. I don't see a case right now where this would trigger for sure (perhaps field initialization checks in ctors, hmm), though, but leaving a note here in case we ever encounter something along those lines.

src/resolver.ts
romdotdog: The errors are different if both fields are private. I think that would make the checks a bit harder to read.

src/resolver.ts
romdotdog: Would you like me to insert some sort of comment there?

dcodeIO: I guess it's fine without, just wanted to leave the note so you and me are aware in case :)

dcodeIO (approved):

dcodeIO:
Thank you!

dcodeIO:
Turns out that covariance on class fields as introduced here in unsound and is rejected by Wasm GC. Means we'll likely have to revert this. See PR linked above for details.

PR #2160, opened 2021-12-01
Add `declare` to parser

romdotdog:

https://github.com/AssemblyScript/assemblyscript/pull/2158#discussion_r757962361 suggests to allow the parser to understand the `declare` modifier, but mark it as unsupported when parsed.

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -1874,30 +1874,37 @@ export class Parser extends DiagnosticEmitter {
     var declareStart = 0;
     var declareEnd = 0;
     var contextIsAmbient = parent.is(CommonFlags.AMBIENT);
-    if (tn.skip(Token.DECLARE)) {
-      if (isInterface) {
-        this.error(
-          DiagnosticCode._0_modifier_cannot_be_used_here,
-          tn.range(), "declare"
-        );
-      } else {
-        if (contextIsAmbient) {
+    if (tn.peek() == Token.DECLARE) {
+      let state = tn.mark();
+      tn.next();
+      if (tn.peek() != Token.COLON) { // modifier
+        tn.discard(state);
+        if (isInterface) {
           this.error(
-            DiagnosticCode.A_declare_modifier_cannot_be_used_in_an_already_ambient_context,
-            tn.range()
-          ); // recoverable
+            DiagnosticCode._0_modifier_cannot_be_used_here,
+            tn.range(), "declare"
+          );
         } else {
-          this.error(
-            DiagnosticCode.Not_implemented_0,
-            tn.range(), "Ambient fields"
-          ); // recoverable
-          
-          flags |= CommonFlags.DECLARE | CommonFlags.AMBIENT;
-          declareStart = tn.tokenPos;
-          declareEnd = tn.pos;
+          if (contextIsAmbient) {
+            this.error(
+              DiagnosticCode.A_declare_modifier_cannot_be_used_in_an_already_ambient_context,
+              tn.range()
+            ); // recoverable
+          } else {
+            this.error(
+              DiagnosticCode.Not_implemented_0,
+              tn.range(), "Ambient fields"
+            ); // recoverable
+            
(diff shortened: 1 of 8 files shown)


src/parser.ts
```diff
@@ -1870,6 +1871,44 @@ export class Parser extends DiagnosticEmitter {
...
+        } else {
+          if (contextIsAmbient) {
+            this.error(
+              DiagnosticCode.A_declare_modifier_cannot_be_used_in_an_already_ambient_context,
+              tn.range()
+            ); // recoverable
+          } else {
+            this.error(
+              DiagnosticCode.Not_implemented_0,
+              tn.range(), "Ambient fields"
```
dcodeIO, line 1896: Appears to also catch methods and properties, so perhaps:
```suggestion
              tn.range(), "Ambient class members"
```

dcodeIO: Just saw that there is another error below handling the case of non-fields, but seems one then gets both diagnosticss?

romdotdog: Non-fields cannot be ambient, so ambient "class members" doesn't make sense. The logic is that one diagnostic will tell the user that ambient fields aren't supported, and one will tell the user that declaring non-fields is illegal :)

dcodeIO: Mostly thinking in terms of having a `declare someMethod();` that seems to produce both errors, with the first one being "Not implemented: Ambient fields" which seems odd. I would be fine with dropping the "Not implemented" one and just generally saying "'declare' modifier cannot be used here". Then the one below can be removed. Or am I missing something?

romdotdog: What you're saying makes sense, but it is ultimately subjective. I'm relying on the fact that the user will see the other error and find that what they are doing is forbidden. "Ambient class members" just doesn't make sense to me as a concept because non-fields can't be ambient.

dcodeIO: The ideal solution would be of course to have exactly one diagnostic. For non-fields:

> 'declare' modifier cannot appear on class elements of this kind.(1031)

For fields:

> Not implemented: Ambient fields.

But two in the former case (the first referring to fields even though not declarring a field) seems odd :)

src/parser.ts
```diff
@@ -1870,6 +1871,39 @@ export class Parser extends DiagnosticEmitter {
     // implemented methods are virtual
     if (isInterface) flags |= CommonFlags.VIRTUAL;
 
+    var declareStart = 0;
+    var declareEnd = 0;
+    var contextIsAmbient = parent.is(CommonFlags.AMBIENT);
+    if (tn.peek() == Token.DECLARE) {
+      let state = tn.mark();
+      tn.next();
+      if (tn.peek() != Token.COLON) { // modifier
```
dcodeIO, line 1880: Seems the same applies to `public`, `protected` and `private` as well? At least the following is allowed in JS: `class B { public = 1; protected = 2; private = 3; declare = 4; }`. Perhaps there could be a more general mechanism for this?

dcodeIO: Also appears that not only `:` terminates here, but `(` as well, say `class B { declare() { } }`?

dcodeIO: Suggestion: `parseClassMemberModifier` returning the `Token` upon success or otherwise a sentinel, setting `this.modifierStart` and `this.modifierEnd` that then can be obtained?

romdotdog: Is that really in this PR's scope though?

dcodeIO: I think the other alternative would be to not perform the check to be in line with public etc. here, and do it in a separate PR after. Would you prefer that? :) Otherwise we have inconsistency I think.

romdotdog: If you're fine with it I'll do it, but it seems more fitting to do this in a separate PR

romdotdog:
After this PR is merged I'll do another that refactors field modifier parsing, how does that sound?

dcodeIO (approved):

dcodeIO:
Thanks :)

Issue #2166, opened 2021-12-07
generic infer error in special case


HerrCai0907:
This code cannot compile. It seems like generic infer error. 
In` subscribe<T>`, T is `AAAAAA`; In `fieldparam.reduce`, T is `string`. But compiler cannot distinguish them.
```
function subscribe<T>(): void {
    let a = new TestClass2(new Array<string>(10));
}

class TestClass2 {
    constructor(fieldparam: Array<string>) {
        function reducefunc(total: i32, curr: string, index: i32, array: Array<string>): i32 {
            return 0;
        }
        fieldparam.reduce(reducefunc, 0);
    }
}

class AAAAAA {}

subscribe<AAAAAA>();
```

```
ERROR TS2322: Type '(i32, ~lib/string/String, i32, ~lib/array/Array<~lib/string/String>) => i32' is not assignable to type '(i32, assembly/index/AAAAAA, i32, ~lib/array/Array<assembly/index/AAAAAA>) => i32'.

         fieldparam.reduce(reducefunc, 0);
                           ~~~~~~~~~~
 in assembly/index.ts(10,27)

ERROR TS2322: Type '~lib/array/Array<~lib/string/String>' is not assignable to type '~lib/array/Array<assembly/index/AAAAAA>'.

       acc = fn(acc, load<T>(this.dataStart + (<usize>i << alignof<T>())), i, this);
                                                                              ~~~~
 in ~lib/array.ts(325,78)
```

dcodeIO:
fixed in #2167

PR #2167, opened 2021-12-07
fix: instance method cannot inherit class generic (#2166)

HerrCai0907:
⯈For class instance method, ctxtypes should use class generic type instead of context. So override is more reasonable.

### DIFF
--- a/src/resolver.ts
+++ b/src/resolver.ts
@@ -2609,12 +2609,8 @@ export class Resolver extends DiagnosticEmitter {
         assert(numClassTypeArguments == classTypeParameters.length);
         for (let i = 0; i < numClassTypeArguments; ++i) {
           let classTypeParameterName = classTypeParameters[i].name.text;
-          if (!ctxTypes.has(classTypeParameterName)) {
-            ctxTypes.set(
-              classTypeParameterName,
-              classTypeArguments[i]
-            );
-          }
+          // override contextual
+          ctxTypes.set(classTypeParameterName, classTypeArguments[i]);
         }
       }
     } else {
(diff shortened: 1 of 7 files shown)


dcodeIO (approved):

dcodeIO:
Iirc the check was there for situations like

```ts
class Foo<T> {
  bar<T>() { baz<T>(); } // here
}
function baz<U>() { ... }
```

where contextual types of the method take precedence over the class's. Before I merge, can you confirm that this case isn't regressing? :)

HerrCai0907:
> Iirc the check was there for situations like
> 
> ```ts
> class Foo<T> {
>   bar<T>() { baz<T>(); } // here
> }
> function baz<U>() { ... }
> ```
> 
> where contextual types of the method take precedence over the class's. Before I merge, can you confirm that this case isn't regressing? :)

The result is OK, I add them into test.
After overridden by class generic, `ctxtype` will be overridden by function generic type again. see below code.
```
    // override whatever is contextual with actual function type arguments
    var signatureNode = prototype.functionTypeNode;
    var typeParameterNodes = prototype.typeParameterNodes;
    var numFunctionTypeArguments: i32;
    if (typeArguments !== null && (numFunctionTypeArguments = typeArguments.length) > 0) {
      assert(typeParameterNodes !== null && numFunctionTypeArguments == typeParameterNodes.length);
      for (let i = 0; i < numFunctionTypeArguments; ++i) {
        ctxTypes.set(
          (<TypeParameterNode[]>typeParameterNodes)[i].name.text,
          typeArguments[i]
        );
      }
    } else {
      assert(!typeParameterNodes || typeParameterNodes.length == 0);
    }
```


They ctxtype process flow is like that: 
- `maybeInferCall` will infer generic of `fn` by outside function type (T-1) . In this step, class generic should not be used.
- `resolveFunction` will override context type with class generic of instance method (T-2) and then override it again with `fn` generic types which are inferred in step1.

So it seems like matching the expectations.
```
function testfunc2166<T-1>(): void {
  let a = new Test2166Ref1<string>();
  // class generic should override non-related function generic(testfunc2166): T in fn is string
  a.fn("11", 1);
}
class Test2166Ref1<T-2> {
  fn<U>(a1: T, a2: U): void {
  }
}
testfunc2166<i64>();
```

dcodeIO:
Makes sense, thanks. Good to now have a test case as well :)

PR #2173, opened 2021-12-11
fix OOB on String.UTF8.encode using nullTerminated = true

### LINKED ISSUE #2021: Maybe there is a bug for String.UTF8.encode and String.UTF8.decode
my test function is :
function putData():void{
    let buf1 = String.UTF8.encode("bench_key_**10000**",true)
    let str1 = String.UTF8.decode(buf1)
    let buf2 = String.UTF8.encode(" the arg is " + str1, true);
}
when executing this function ,then it will crash.
if I change this function to be:
function putData():void{
    let buf1 = String.UTF8.encode("bench_key_**9999**",true)
    let str1 = String.UTF8.decode(buf1)
    let buf2 = String.UTF8.encode(" the arg is " + str1, true);
}
or change it to be:
function putData():void{
    let buf1 = String.UTF8.encode("bench_key_**10000**",true)
    let str1 = String.UTF8.decode(buf1, **true**)
    let buf2 = String.UTF8.encode(" the arg is " + str1, true);
}
it will not crash.

But I do not know why.

Can anyone explain this phenomenon?  Thank you very much.

romdotdog:
fixes #2021

## Examples
before it worked like this
```ts
String.UTF8.byteLength("a\0", true); // -> 2
String.UTF8.encode("a\0", true); // writes 0x61, 0x00, and then another 0x00
```
now it works like this
```ts
String.UTF8.byteLength("a\0", true); // -> 2
String.UTF8.encode("a\0", true); // writes 0x61, 0x00 and breaks because of the null
```


tests/compiler/std/string-encoding.ts
```diff
@@ -89,6 +89,7 @@ function testUTF8EncodeNullTerminated(): void {
   assert(load<u8>(ptr, 8) == 0xAD);
   assert(load<u8>(ptr, 9) == 0xA2);
   assert(load<u8>(ptr, 10) == 0x00);
+  assert(String.UTF8.encode("aaaaaaaaaaaaaaaaaaaaaaaaaaa\0", true));
```
dcodeIO, line 92: I guess this one could check the result of `UTF8.byteLength` against the buffer's `.byteLength` to be sure.

romdotdog: sure

dcodeIO:
Thanks :)

Issue #2178, opened 2021-12-22
Template literals cause GC crash


HerrCai0907:
code like this, after serval times running, wasm will crash
```typescript
@global const ASC_GC_SWEEPCOST: i32 = 1000;  // make sweep bigger so the bug will easier to be reproduced

function randomNew(): void {
  trace("test");
  const a = Math.random();
  `bbbbbbb${a.toString()}bb`;
}

for (;;) {
  randomNew();
}

```

Error: abort:  at ~lib/rt/itcms.ts:228:20
    at abort (/home/ccc/Desktop/test/as/node_modules/@assemblyscript/loader/umd/index.js:105:13)
    at ~lib/rt/itcms/step (wasm-function[42]:0x21cd)
    at ~lib/rt/itcms/interrupt (wasm-function[43]:0x220f)
    at ~lib/rt/itcms/__new (wasm-function[54]:0x279e)
    at ~lib/util/number/dtoa (wasm-function[71]:0x2b7d)
    at ~lib/number/F64#toString (wasm-function[55]:0x27e0)
    at assembly/index/randomNew (wasm-function[70]:0x2a5c)
    at start:assembly/index (wasm-function[62]:0x28ea)
    at ~start (wasm-function[68]:0x29b3)
    at Object.curr.<computed> [as _start] (/home/ccc/Desktop/test/as/node_modules/@assemblyscript/loader/umd/index.js:545:20)

HerrCai0907:
This is my analysis but I'm not sure whether it is correct.

```typescript
function step(): usize {
  // Magic constants responsible for pause times. Obtained experimentally
  // using the compiler compiling itself. 2048 budget pro run by default.
  const MARKCOST = isDefined(ASC_GC_MARKCOST) ? ASC_GC_MARKCOST : 1;
  const SWEEPCOST = isDefined(ASC_GC_SWEEPCOST) ? ASC_GC_SWEEPCOST : 10;
  var obj: Object;
  switch (state) {
    case STATE_IDLE: {
      state = STATE_MARK;
      visitCount = 0;
      visitRoots(VISIT_SCAN);
      iter = toSpace;
      return visitCount * MARKCOST;
    }
    case STATE_MARK: {
      let black = i32(!white);
      obj = iter.next;
      while (obj != toSpace) {
        iter = obj;
        if (obj.color != black) { // skip already-blacks (pointerfree)
          obj.color = black;
          visitCount = 0;
[...]

PR #2179, opened 2021-12-22
Fix: Template literals cause GC crash #2178

HerrCai0907:
⯈color of StaticArray in data section will set to 3 (transparent) to avoid reference type make it gray and cause GC crash when call `__link`


dcodeIO:
One aspect that comes to me mind is that if static data is always transparent (=3), then it won't be trivially possible to pass these down to JS, since on the JS side one would `__pin` the object, but it is already pinned (=3). We'll need a different fix there, sadly.

PR #2197, opened 2022-02-07
Fix empty interface compile crash

romdotdog:
A while ago I reverse engineered the stack trace of #2085 to find [this bug](https://www.assemblyscript.org/editor.html#IyFvcHRpbWl6ZT1zcGVlZCZydW50aW1lPXN0dWIKaW50ZXJmYWNlIEZvbyB7fQpjb25zdCBmb286IEZvbyA9IHt9OwojIWh0bWwKPHRleHRhcmVhIGlkPSJvdXRwdXQiIHN0eWxlPSJoZWlnaHQ6IDEwMCU7IHdpZHRoOiAxMDAlIiByZWFkb25seT48L3RleHRhcmVhPgo8c2NyaXB0Pgpsb2FkZXIuaW5zdGFudGlhdGUobW9kdWxlX3dhc20sIHsgLyogaW1wb3J0cyAqLyB9KQogIC50aGVuKCh7IGV4cG9ydHMgfSkgPT4gewogICAgY29uc3Qgb3V0cHV0ID0gZG9jdW1lbnQuZ2V0RWxlbWVudEJ5SWQoJ291dHB1dCcpCiAgICBmb3IgKGxldCBpID0gMDsgaSA8PSAxMDsgKytpKSB7CiAgICAgIG91dHB1dC52YWx1ZSArPSBgZmliKCR7aX0pID0gJHtleHBvcnRzLmZpYihpKX1cbmAKICAgIH0KICB9KQo8L3NjcmlwdD4=)

Since interfaces are extended from classes, the asserts can be modified to check for interfaces.


dcodeIO:
Right now, interfaces only describe what's present in classes that implement them, but don't have own storage. There are more like types. As such, constructing an interface from an object literal, as in the test, would be limited to this exact test case where there is no actual data. Is this the intent, to just unblock this specific case?

romdotdog:
I see what you mean. Implementing hidden classes that implement interfaces would solve this problem more elegantly than just adjusting asserts, though this solution would work. I guess I will just have to wait for interface hidden classes to be implemented. Would a better temporary solution be to error when using interfaces in this fashion?

dcodeIO:
Producing a proper error seems preferable for now, yeah, until we know more. Hitting the asserts there is definitely a bug.

src/compiler.ts
```diff
@@ -8629,6 +8629,13 @@ export class Compiler extends DiagnosticEmitter {
       );
       return module.unreachable();
     }
+    if (classReference.kind == ElementKind.INTERFACE) {
+      this.error(
+        DiagnosticCode.Not_implemented_0,
+        expression.range, "Interface hidden classes"
+      );
+      return module.unreachable();
```
dcodeIO, line 8637: Let's also set `this.currentType` here even though emitting the error. Otherwise its value is undefined and can lead to unwanted diagnostics down the road. Could just move the next two lines below up.

dcodeIO:
Thanks! :)

PR #2203, opened 2022-02-12
Fix compiler crash when certain operators used on types

romdotdog:
## The issue

When the prefix compound unary operators (`++x` and `--x`) are used on types, the compiler crashes
[Example](https://www.assemblyscript.org/editor.html#IyFvcHRpbWl6ZT1zcGVlZCZydW50aW1lPXN0dWIKKytTdHJpbmc7CiMhaHRtbAo8dGV4dGFyZWEgaWQ9Im91dHB1dCIgc3R5bGU9ImhlaWdodDogMTAwJTsgd2lkdGg6IDEwMCUiIHJlYWRvbmx5PjwvdGV4dGFyZWE+CjxzY3JpcHQ+CmxvYWRlci5pbnN0YW50aWF0ZShtb2R1bGVfd2FzbSwgeyAvKiBpbXBvcnRzICovIH0pCiAgLnRoZW4oKHsgZXhwb3J0cyB9KSA9PiB7CiAgICBjb25zdCBvdXRwdXQgPSBkb2N1bWVudC5nZXRFbGVtZW50QnlJZCgnb3V0cHV0JykKICAgIGZvciAobGV0IGkgPSAwOyBpIDw9IDEwOyArK2kpIHsKICAgICAgb3V0cHV0LnZhbHVlICs9IGBmaWIoJHtpfSkgPSAke2V4cG9ydHMuZmliKGkpfVxuYAogICAgfQogIH0pCjwvc2NyaXB0Pg==)
```
▌ 
▌ 
▌     at assert (D:\assemblyscript\std\portable\index.js:200:11)
▌     at Compiler.makeAssignment (D:\assemblyscript\src\compiler.ts:6150:5)
▌     at Compiler.compileUnaryPrefixExpression (D:\assemblyscript\src\compiler.ts:9921:17)
▌     at Compiler.compileExpression (D:\assemblyscript\src\compiler.ts:3539:21)
▌     at Compiler.compileExpressionStatement (D:\assemblyscript\src\compiler.ts:2521:17)
▌     at Compiler.compileStatement (D:\assemblyscript\src\compiler.ts:2213:21)
▌     at Compiler.compileTopLevelStatement (D:\assemblyscript\src\compiler.ts:2175:25)
▌     at Compiler.compileFile (D:\assemblyscript\src\compiler.ts:1111:12)
▌     at Compiler.compile (D:\assemblyscript\src\compiler.ts:486:14)
▌     at Object.compile (D:\assemblyscript\src\index.ts:280:32)
▌     at D:\assemblyscript\cli\asc.js:902:31
▌     at measure (D:\assemblyscript\cli\asc.js:1409:3)
▌     at Object.main (D:\assemblyscript\cli\asc.js:900:24)
▌     at runTest (D:\assemblyscript\tests\compiler.js:178:7)
▌     at Array.forEach (<anonymous>)
▌ 
▌ If it refers to the dist files, try to 'npm install source-map-support' and
▌ run again, which should then show the actual code location in the sources.
▌ 
▌ 
▌ Thank you!
```

The postfix compound unary operators (`x++` and `x--`) actually don't crash the compiler, and emit this error
```
;; ERROR AS234: Expression does not compile to a value at runtime.
;; 
;;  String++;
;;  ~~~~~~
;;  in module.ts(1,1)
```

This is due to [the compiler checking](https://github.com/AssemblyScript/assemblyscript/blob/fa14b3b03bd4607efa52aaff3132bea0c03a7989/src/compiler.ts#L9312) that the operand doesn't compile to an unreachable instruction in the `compileUnaryPostfixExpression` function, in contrast to the `compileUnaryPrefixExpression` function where this check never happens. This PR adds the appropriate checks in that function. 


dcodeIO:
The problem here appears to come from `currentType` remaining `i32` when an expression does not resolve to a type. Perhaps, another option to fix this could be to set `this.currentType = Type.void` whenever `DiagnosticCode.Expression_does_not_compile_to_a_value_at_runtime` is diagnosed, which then still triggers `DiagnosticCode.The_0_operator_cannot_be_applied_to_type_1` alongside?

romdotdog:
That seems to be congruent to the behavior of 
```ts
class Foo {}
Foo += 1;
```
which I agree with. I'll update this PR to include updating postfix compound unary operators to match intended behavior.

romdotdog:
I've already asked this in the Discord but I'll ask again here.
Can we remove `.exceptVoid` here
https://github.com/AssemblyScript/assemblyscript/blob/89035f576a5f33ab6026abe8e1d9f14f84797264/src/compiler.ts#L9583
and use pre-existing checks to match intended behavior?
https://github.com/AssemblyScript/assemblyscript/blob/89035f576a5f33ab6026abe8e1d9f14f84797264/src/compiler.ts#L9597

dcodeIO:
Iirc it is or was a workaround for not having a better type than `void` to indicate (resolve) failure, such as `never` in TS. Adding a proper `never` would probably make it trivial to remove.

romdotdog:
Upon closer inspection,
```ts
class Foo {}
Foo += 1;
```
actually emits
```
;; ERROR TS2469: The '+' operator cannot be applied to type 'void'.
;; 
;;  Foo += 1;
;;  ~~~~~~~~
;;  in module.ts(2,1)
```
rather than the desired `DiagnosticCode.The_0_operator_cannot_be_applied_to_type_1`. If you wish to unite these inconsistent behaviors, I can close this PR (since this is beyond its scope) and open a new one for that purpose.

PR #2213, opened 2022-02-16
trusktr/esm

trusktr:


cli/index.js
```diff
@@ -402,24 +402,15 @@ export async function main(argv, options) {
       let filename = transformArgs[i].trim();
       let resolved;
       let transform;
-      if (require.resolve) {
```
trusktr, line 405: `require.resolve` always exists according to Node types, since Node 0.3. So I don't think this branching is needed here.

dcodeIO: The underlying idea here was that `require` itself is polyfilled when compiling for the browser (throws an error), but the `.resolve` property is not, so this can be used to detect the environment. [See](https://github.com/AssemblyScript/assemblyscript/blob/esm/util/browser/module.js). Perhaps could need a comment :)

trusktr: Yeah, comments definitely helpful. :) Also, why polyfill require for browser?

trusktr: To be "moving forward", I think good if we totally eliminate all `require`s, especially for browsers. :)

trusktr: Is there a browser test for this? If not, I can try to add it, but I need some guidance. I've no idea how this works in a browser yet.

dcodeIO: The use of `require` is meant to support existing CommonJS transforms, which may again `require`. For this, `createRequire` is used under Node with ESM, since `require` isn't defined directly, which is the mechanism being polyfilled, returning a mock that produces a "module not found" kind of error in browsers. The code loading transforms detects this, and uses only ESM in browsers.

trusktr: :+1:

trusktr: There were a few spots where I removed require, but that only that one spot still needs it right?

Issue #2225, opened 2022-03-16
compiler crash when using --use


HerrCai0907:
// builtin.ts
``` typescript
export function abort(message: string | null, fileName: string | null, lineNumber: u32, columnNumber: u32): void {
}
```

// index.ts
``` typescript
export function add(a: i32, b: i32): i32 {
  abort(null, null, 0, 0);
  return 0;
}
```

// asconfig
```json
{
  "targets": {
    "debug": {
      "binaryFile": "build/test.wasm",
      "use": "abort=assembly/builtin/abort"
    }
  },
  "options": {}
}
```

crash report like this:
▌ Whoops, the AssemblyScript compiler has crashed during initialize :-(
▌ 
▌ Here is the stack trace hinting at the problem, perhaps it's useful?
▌ 
▌ Error: no such global element: assembly/builtin/abort
▌     at Program.initialize (/home/ccc/Desktop/opensource/assemblyscript/src/program.ts:1415:19)
▌     at Object.initializeProgram (/home/ccc/Desktop/opensource/assemblyscript/src/index.ts:274:11)
▌     at /home/ccc/Desktop/opensource/assemblyscript/cli/asc.js:886:22
▌     at measure (/home/ccc/Desktop/opensource/assemblyscript/cli/asc.js:1409:3)
▌     at Object.main (/home/ccc/Desktop/opensource/assemblyscript/cli/asc.js:884:27)
▌     at /home/ccc/Desktop/opensource/assemblyscript/bin/asc:19:47
▌ 
▌ If it refers to the dist files, try to 'npm install source-map-support' and
▌ run again, which should then show the actual code location in the sources.
▌ 
▌ If you see where the error is, feel free to send us a pull request. If not,
▌ please let us know: https://github.com/AssemblyScript/assemblyscript/issues
▌ 
▌ Thank you!

PR #2227, opened 2022-03-17
fix: --use cause compile crash (#2225)

HerrCai0907:
⯈ change error to diagnostic message ( issue #2225 )

### DIFF
--- a/src/diagnosticMessages.json
+++ b/src/diagnosticMessages.json
@@ -184,6 +184,7 @@
   "Type '{0}' has no call signatures.": 2757,
 
   "File '{0}' not found.": 6054,
+  "no such global element: '{0}'":6055,
   "Numeric separators are not allowed here.": 6188,
   "Multiple consecutive numeric separators are not permitted.": 6189,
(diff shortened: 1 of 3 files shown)


src/diagnosticMessages.json
```diff
@@ -184,6 +184,7 @@
   "Type '{0}' has no call signatures.": 2757,
 
   "File '{0}' not found.": 6054,
+  "no such global element: '{0}'":6055,
```
dcodeIO, line 187: Diagnostic 6055 is [different in TypeScript](https://github.com/microsoft/TypeScript/blob/073ac920adee8e517350a455908aa420b29f9422/src/compiler/diagnosticMessages.json#L4260-L4263), and we try to prevent collisions there. What seems to fit, though, is to renumber this diagnostic to 110 and move it up to the 1XX range of custom AS compiler errors.

Error ranges are about:
* 100-199: Custom AS compiler errors
* 200-899: Custom AS compilation diagnostics
* 900-999: Pedantic diagnostics
* 1000+: Reused TypeScript errors

PR #2229, opened 2022-03-18
fix setUTCMonth bug

rayshen:
this.month is the real month, but setUTCMonth value is 0 ~ 11, the setUTCMonth function makes an error when this.month == month

for exp:
in JS
const date = new Date(1647585448570);  //2022-03-18
date.getUTCMonth(); // return 2
date.setUTCMonth(3);
date.getUTCMonth(); // return 3

In AS
const date = new Date(1647585448570);  //2022-03-18
date.getUTCMonth(); // return 2
date.setUTCMonth(3); // set failed
date.getUTCMonth(); // return 2  error

### DIFF
--- a/std/assembly/date.ts
+++ b/std/assembly/date.ts
@@ -169,7 +169,7 @@ export class Date {
   }
 
   setUTCMonth(month: i32): void {
-    if (this.month == month) return;
+    if (this.month == month + 1) return;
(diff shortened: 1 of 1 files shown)


dcodeIO:
Thanks, tracked in #2239 including test updates :)

PR #2239, opened 2022-03-25
Fix setUTCMonth check off by one

dcodeIO:
Fixes #2229 and related tests. Not sure what exactly is being tested with these huge timestamps (year 252309), though, so I'm not particularly confident regarding the assertion changes  :) @ColinEberhardt, any idea?

### DIFF
--- a/std/assembly/date.ts
+++ b/std/assembly/date.ts
@@ -168,10 +168,10 @@ export class Date {
     this.setTime(i64(daysSinceEpoch(this.year, this.month, day)) * MILLIS_PER_DAY + ms);
   }
 
-  setUTCMonth(month: i32): void {
-    if (this.month == month) return;
+  setUTCMonth(month: i32, day: i32 = this.day): void {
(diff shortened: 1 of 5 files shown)


ColinEberhardt:
> @ColinEberhardt, any idea?

Hmmm ... good question, it looks like these values are quite close to the largest Date that JavaScript is able to represent, but not quite at the boundary. Not much help I'm afraid!

dcodeIO:
Thanks :) TIL

> It should be noted that the maximum Date is not of the same value as the maximum safe integer (Number.MAX_SAFE_INTEGER is 9,007,199,254,740,991). Instead, it is defined in ECMA-262 that a maximum of ±100,000,000 (one hundred million) days relative to January 1, 1970 UTC (that is, April 20, 271821 BCE ~ September 13, 275760 CE) can be represented by the standard Date object (equivalent to ±8,640,000,000,000,000 milliseconds).

Makes me assume that it's implicitly making sure that huge dates work.

PR #2240, opened 2022-03-26
Add unimplemented error for first-class built-ins

### LINKED ISSUE #2202: Assignment to built-in function crashes the compiler
## Code
```ts
const x = v128;
// or just:
v128
```
[Try it in the playground](https://www.assemblyscript.org/editor.html#IyFvcHRpbWl6ZT1zcGVlZCZydW50aW1lPXN0dWIKY29uc3QgeCA9IHYxMjg7CiMhaHRtbAo8dGV4dGFyZWEgaWQ9Im91dHB1dCIgc3R5bGU9ImhlaWdodDogMTAwJTsgd2lkdGg6IDEwMCUiIHJlYWRvbmx5PjwvdGV4dGFyZWE+CjxzY3JpcHQ+CmxvYWRlci5pbnN0YW50aWF0ZShtb2R1bGVfd2FzbSwgeyAvKiBpbXBvcnRzICovIH0pCiAgLnRoZW4oKHsgZXhwb3J0cyB9KSA9PiB7CiAgICBjb25zdCBvdXRwdXQgPSBkb2N1bWVudC5nZXRFbGVtZW50QnlJZCgnb3V0cHV0JykKICAgIGZvciAobGV0IGkgPSAwOyBpIDw9IDEwOyArK2kpIHsKICAgICAgb3V0cHV0LnZhbHVlICs9IGBmaWIoJHtpfSkgPSAke2V4cG9ydHMuZmliKGkpfVxuYAogICAgfQogIH0pCjwvc2NyaXB0Pg==)

## Stack trace
```
▌ 
▌ 
▌     at assert (D:\assemblyscript\std\portable\index.js:200:11)
▌     at Compiler.ensureRuntimeFunction (D:\assemblyscript\src\compiler.ts:2074:5)
▌     at Compiler.compileIdentifierExpression (D:\assemblyscript\src\compiler.ts:7866:27)
▌     at Compiler.compileExpression (D:\assemblyscript\src\compiler.ts:3511:21)
▌     at Compiler.compileGlobal (D:\assemblyscript\src\compiler.ts:1180:25)
▌     at Compiler.compileTopLevelStatement (D:\assemblyscript\src\compiler.ts:2140:20)
▌     at Compiler.compileFile (D:\assemblyscript\src\compiler.ts:1111:12)
▌     at Compiler.compile (D:\assemblyscript\src\compiler.ts:486:14)
▌     at Object.compile (D:\assemblyscript\src\index.ts:280:32)
▌     at D:\assemblyscript\cli\asc.js:902:31
▌     at measure (D:\assemblyscript\cli\asc.js:1409:3)
▌     at Object.main (D:\assemblyscript\cli\asc.js:900:24)
▌     at runTest (D:\assemblyscript\tests\compiler.js:178:7)
▌     at Array.forEach (<anonymous>)
▌     at D:\assemblyscript\tests\compiler.js:557:16
▌ 
▌ If it refers to the dist files, try to 'npm install source-map-support' and
▌ run again, which should then show the actual code location in the sources.
▌ 
▌ 
▌ Thank you!
```

romdotdog:
Fixes #2202. Another candidate for this error could be functions with `DecoratorFlags.LAZY`, but it seems that none exist.


dcodeIO:
The underlying cause is quite similar to what happens when trying to pass a class by value. On this background, a perhaps more precise error describing what's going on under the hood could be AS234, "Expression does not compile to a value at runtime". Wdyt?

romdotdog:
I chose the unimplemented error because it may be possible to implement first-class "proxy" functions if anyone may need such a thing, i.e.
```ts
function proxy(i: u16): u8 { // generated by the compiler
  return u8(i);
}

function converter<I, R>(f: (i: I) => R, i: I): R {
  return f(i);
}

converter<u16, u8>(proxy, 32767);
```
[That isn't to say that it will be implemented.](https://doc.rust-lang.org/std/macro.todo.html)
> The difference between [unimplemented!](https://doc.rust-lang.org/std/macro.unimplemented.html) and [todo!](https://doc.rust-lang.org/std/macro.todo.html) is that while todo! conveys an intent of implementing the functionality later and the message is “not yet implemented”, unimplemented! makes no such claims. Its message is “not implemented”. Also some IDEs will mark todo!s.

dcodeIO:
My assumption would be that we'd eventually implement such wrappers, yeah. Basically just wrap the builtin's instruction in a function so it can be represented with a value. That would be a case for "todo" as of the quoted paragraph, though, while the idea above is rather to just use the more descriptive message of what's going on. Not feeling strongly about which one, though, mostly just my immediate thoughts. I guess ideally we'd just implement wrappers :)

romdotdog:
> That would be a case for "todo" as of the quoted paragraph, though

`todo!` is a "subset" of `unimplemented!` so an `unimplemented!` can be implemented later. If you want to be specific, we can add a new `TODO: {0}` error.

> I guess ideally we'd just implement wrappers :)

Triaging and a roadmap would help, I think. My objective for this PR and others is to fix all the crashes that I've posted as issues, at least as provisional errors, so that my AS fuzzer can catch more complex and elusive crashes that are not shadowed by trivial and common ones.

dcodeIO:
Sounds like a good plan. I'm OK with adding the suggested error :)

src/compiler.ts
```diff
@@ -7756,6 +7756,13 @@ export class Compiler extends DiagnosticEmitter {
           uniqueMap<string,Type>(flow.contextualTypeArguments)
         );
         if (!functionInstance || !this.compileFunction(functionInstance)) return module.unreachable();
+        if (functionInstance.hasDecorator(DecoratorFlags.BUILTIN)) {
+          this.error(
+            DiagnosticCode.Not_implemented_0,
+            expression.range, "First-class built-ins"
+          );
```
dcodeIO, line 7763: Since `functionInstance` is known here, setting `this.currentType` to the function's type anyhow could prevent further compile errors that only come up due to this one.

romdotdog:
> Sounds like a good plan. I'm OK with adding the suggested error :)

To be consistent, adding the suggested error would be, in my opinion, enough of a change to belong in a separate PR which would involve auditing all `DiagnosticCode.Not_implemented_0` errors, and would require one who is capable of deciding what will be and could do without implementation. My proposition is to leave this diagnostic as `DiagnosticCode.Not_implemented_0` and create the new error in a separate PR along with the aforementioned audits.

dcodeIO:
Thanks! Would agree that it doesn't seem pressing to add a TODO kind error just yet :)

PR #2252, opened 2022-04-07
#1125 Fix parsing of unusual trailing expressions separated by whitespace

technohippy:
Concerning:

- Is "unexpected token error" appropriate for this error?
- Serialized source text still has ";". Is it OK?


src/tokenizer.ts
```diff
@@ -1697,6 +1697,18 @@ export class Tokenizer extends DiagnosticEmitter {
     return this.readHexadecimalEscape(4, startIfTaggedTemplate);
   }
 
+  hasNoMoreExpressionInLine(): bool {
+    let pos = this.pos;
+    const end = this.end;
+    do {
+      const c = this.source.text.charCodeAt(pos);
```
dcodeIO, line 1704: Looks like `pos` can be `== end` here, so the access can fail. A `while` instead would probably work.

src/tokenizer.ts
```diff
@@ -1697,6 +1697,18 @@ export class Tokenizer extends DiagnosticEmitter {
     return this.readHexadecimalEscape(4, startIfTaggedTemplate);
   }
 
+  hasNoMoreExpressionInLine(): bool {
```
dcodeIO, line 1700: Technically not looking for an expression, but for non-whitespace. Not sure what would be a good name, hmm.

dcodeIO:
I wonder whether this is only applicable to `var`iables and `return`s. Another one that comes to mind is `throw`. Do you know what the spec says exactly for this case? If it runs deeper, perhaps a fix could be more general.

technohippy:
> I wonder whether this is only applicable to variables and returns. Another one that comes to mind is throw. Do you know what the spec says exactly for this case? If it runs deeper, perhaps a fix could be more general.

I agree. I don't know the spec well but I should consider more deeply.

Anyway, this change counts some valid codes like below as an error.

```js
function foo(): i32 {
  return 123 }
```

I'll close this pullreq for now and reopen it after making more proper solution.

technohippy:
I read the spec but it's too complecated for me to handle all cases.

- https://tc39.es/ecma262/#sec-automatic-semicolon-insertion

Anyway, this pull req can fix issue #1125. If other cases are found, please replace `tn.skip(Token.SEMICOLON)` to `this.checkRuleCompleted(tn)`.

src/tokenizer.ts
technohippy: fixed

dcodeIO:
Thanks :) There are likely more ASI-related issues, but seems good to have this merged for the (probably most prominent) cases covered herein.

PR #2260, opened 2022-04-13
Fix: Cannot find global type 'Number'

mrchantey:
This PR is to fix the red squigglies that I've noticed appearing in the `tsconfig.json` when upgrading from version `0.19.23` to `0.20.4`. 

The error is `Cannot find global type 'Number'`.

Screenshot 2022-04-13 143750

⯈
⯈
⯈

- [ x] I've read the contributing guidelines
- [ x] I've added my name and email to the NOTICE file


dcodeIO:
Thanks!

Issue #2204, opened 2022-02-13
Declaration that indexes itself crashes the compiler


romdotdog:
## Code

```ts
let a = a[0];
```

[Try it in the playground](https://www.assemblyscript.org/editor.html#IyFvcHRpbWl6ZT1zcGVlZCZydW50aW1lPXN0dWIKbGV0IGEgPSBhWzBdOwojIWh0bWwKPHRleHRhcmVhIGlkPSJvdXRwdXQiIHN0eWxlPSJoZWlnaHQ6IDEwMCU7IHdpZHRoOiAxMDAlIiByZWFkb25seT48L3RleHRhcmVhPgo8c2NyaXB0Pgpsb2FkZXIuaW5zdGFudGlhdGUobW9kdWxlX3dhc20sIHsgLyogaW1wb3J0cyAqLyB9KQogIC50aGVuKCh7IGV4cG9ydHMgfSkgPT4gewogICAgY29uc3Qgb3V0cHV0ID0gZG9jdW1lbnQuZ2V0RWxlbWVudEJ5SWQoJ291dHB1dCcpCiAgICBmb3IgKGxldCBpID0gMDsgaSA8PSAxMDsgKytpKSB7CiAgICAgIG91dHB1dC52YWx1ZSArPSBgZmliKCR7aX0pID0gJHtleHBvcnRzLmZpYihpKX1cbmAKICAgIH0KICB9KQo8L3NjcmlwdD4=)

## Stack trace

```
▌ Whoops, the AssemblyScript compiler has crashed during compile :-(
▌ 
▌ Here is the stack trace hinting at the problem, perhaps it's useful?
▌ 
▌ RangeError: Maximum call stack size exceeded
▌     at Flow.lookup (D:\assemblyscript\src\flow.ts:566:9)
▌     at Resolver.lookupIdentifierExpression (D:\assemblyscript\src\resolver.ts:1139:27)
▌     at Resolver.resolveIdentifierExpression (D:\assemblyscript\src\resolver.ts:1202:24)
▌     at Resolver.resolveExpression (D:\assemblyscript\src\resolver.ts:1054:21)
▌     at Resolver.lookupElementAccessExpression (D:\assemblyscript\src\resolver.ts:1464:27)
▌     at Resolver.resolveElementAccessExpression (D:\assemblyscript\src\resolver.ts:1498:24)
▌     at Resolver.resolveExpression (D:\assemblyscript\src\resolver.ts:1039:21)
[...]

PR #2264, opened 2022-04-15
fix: #2204, #2211 resolve loop

technohippy:
⯈fix: #2204 
⯈fix: #2211


romdotdog (commented):
You should use an O(1) `Set<Expression>` instead of an `Expression[]`

src/resolver.ts
```diff
@@ -1006,6 +1009,21 @@ export class Resolver extends DiagnosticEmitter {
     ctxType: Type = Type.auto,
     /** How to proceed with eventual diagnostics. */
     reportMode: ReportMode = ReportMode.REPORT
+  ): Type | null {
+    if (this.resolvingExpressions.has(node)) return null;
+
+    this.resolvingExpressions.add(node);
+    const resolved = this.unsafeResolveExpression(node, ctxFlow, ctxType, reportMode);
+    this.resolvingExpressions.delete(node);
+    return resolved;
```
dcodeIO, line 1018: Can cache the property access in a local here:
```suggestion
    const resolvingExpressions = this.resolvingExpressions;
    if (resolvingExpressions.has(node)) return null;
    resolvingExpressions.add(node);
    const resolved = this.unsafeResolveExpression(node, ctxFlow, ctxType, reportMode);
    resolvingExpressions.delete(node);
    return resolved;
```

src/resolver.ts
```diff
@@ -1006,6 +1009,21 @@ export class Resolver extends DiagnosticEmitter {
...
+  }
+
+  /** Resolves an expression to its static type. (may cause stack overflow) */
+  private unsafeResolveExpression(
```
dcodeIO, line 1022: Perhaps...?
```suggestion
  private doResolveExpression(
```
I think the Tokenizer uses a similar `unsafe` prefix for a method, but it's not *technically* unsafe.

dcodeIO:
Thanks! :)

Issue #2201, opened 2022-02-12
Compiler enters infinite loop when class extends itself


romdotdog:
## Code
```ts
class Foo extends Foo {
  bar(): void {}
}
```

Strangely if you omit the `void` return type like this
```ts
class Foo extends Foo {
  bar() {}
}
```
only the AS playground enters an infinite loop while the main branch just exits with a parse error
```
  ERROR TS1110: Type expected.

     bar() {}
          ^
   in test.ts(2,8)
```

PR #2266, opened 2022-04-18
fix: class extends itself #2201

technohippy:
⯈fix: #2201 

### DIFF
--- a/src/diagnosticMessages.json
+++ b/src/diagnosticMessages.json
@@ -46,7 +46,6 @@
   "Property '{0}' is always assigned before being used.": 233,
   "Expression does not compile to a value at runtime.": 234,
   "Only variables, functions and enums become WebAssembly module exports.": 235,
-  "Circular inheritance": 236,
 
   "Importing the table disables some indirect call optimizations.": 901,
   "Exporting the table disables some indirect call optimizations.": 902,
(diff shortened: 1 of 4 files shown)


src/diagnosticMessages.json
```diff
@@ -46,6 +46,7 @@
   "Property '{0}' is always assigned before being used.": 233,
   "Expression does not compile to a value at runtime.": 234,
   "Only variables, functions and enums become WebAssembly module exports.": 235,
+  "Circular inheritance": 236,
```
romdotdog, line 49: It's better to use TypeScript's errors if possible; i.e. add and use TS2506 for this
`"'{0}' is referenced directly or indirectly in its own base expression": 2506`

romdotdog (commented):
Thank you for the PR.

src/diagnosticMessages.json
technohippy: Thanks, done.

dcodeIO:
I think `markVirtuals` may not be the ideal place for the check, as it is doing something rather specific at the end of compilation. A better place would be where `.base` is set in a way that the error condition appears, producing a diagnostic then, so `markVirtuals` wouldn't need to check?

dcodeIO:
Thank you!

Issue #2200, opened 2022-02-12
`export default` in namespace crashes compiler


romdotdog:
## Code
```ts
namespace Foo {
  export default foo;
}
```
[Try it in the playground](https://www.assemblyscript.org/editor.html#IyFvcHRpbWl6ZT1zcGVlZCZydW50aW1lPXN0dWIKbmFtZXNwYWNlIEZvbyB7CiAgZXhwb3J0IGRlZmF1bHQgZm9vOwp9CiMhaHRtbAo8dGV4dGFyZWEgaWQ9Im91dHB1dCIgc3R5bGU9ImhlaWdodDogMTAwJTsgd2lkdGg6IDEwMCUiIHJlYWRvbmx5PjwvdGV4dGFyZWE+CjxzY3JpcHQ+CmxvYWRlci5pbnN0YW50aWF0ZShtb2R1bGVfd2FzbSwgeyAvKiBpbXBvcnRzICovIH0pCiAgLnRoZW4oKHsgZXhwb3J0cyB9KSA9PiB7CiAgICBjb25zdCBvdXRwdXQgPSBkb2N1bWVudC5nZXRFbGVtZW50QnlJZCgnb3V0cHV0JykKICAgIGZvciAobGV0IGkgPSAwOyBpIDw9IDEwOyArK2kpIHsKICAgICAgb3V0cHV0LnZhbHVlICs9IGBmaWIoJHtpfSkgPSAke2V4cG9ydHMuZmliKGkpfVxuYAogICAgfQogIH0pCjwvc2NyaXB0Pg==)
## Stack trace
```
▌ Whoops, the AssemblyScript compiler has crashed during initialize :-(
▌ 
▌ Here is the stack trace hinting at the problem, perhaps it's useful?
▌ 
▌ AssertionError: assertion failed
▌     at assert (D:\assemblyscript\std\portable\index.js:200:11)
▌     at Program.initializeNamespace (D:\assemblyscript\src\program.ts:2545:18)
▌     at Program.initialize (D:\assemblyscript\src\program.ts:1114:18)
▌     at Object.initializeProgram (D:\assemblyscript\src\index.ts:274:11)
▌     at D:\assemblyscript\cli\asc.js:886:22
▌     at measure (D:\assemblyscript\cli\asc.js:1409:3)
▌     at Object.main (D:\assemblyscript\cli\asc.js:884:27)
▌     at runTest (D:\assemblyscript\tests\compiler.js:178:7)
[...]

PR #2267, opened 2022-04-19
fix: export default in namespace #2200

technohippy:
⯈fix: #2200 

Now the same error message as TS is used.


romdotdog:
I think changing the message for this error could be beneficial, wdyt? Do you think it's confusing? cc: @dcodeIO

dcodeIO:
I guess the reference to "ECMAScript-style module" can be dropped, as this is all AS supports anyway. Perhaps: "A default export can only be used in a module." Still about the same message, but less over-precise.

Another thought: I wonder that a `NodeKind.EXPORT` even ends up in namespace initialization. Seems brittle. Can this perhaps be checked during parsing already, not appending the export in this place?

romdotdog:
I was thinking something more like `A default export cannot be used in a namespace.` or maybe even better `A default export can only be used in the top level.` unless there is another case where they can be used?

technohippy:
`TS1258: A default export must be at the top level of a file or module declaration.` may be the one we want.

technohippy:
hmm... I choose `A default export can only be used in a module.` for now. Anyone can change this anytime.

technohippy:
@dcodeIO Test / Bootstrap throws a strange error now. `npm run bootstrap` can run on my local env. Do you have any idea?

```
TypeError: WebAssembly.compileStreaming is not a function
    at file:///home/runner/work/assemblyscript/assemblyscript/build/assemblyscript.debug.js:2[23](https://github.com/AssemblyScript/assemblyscript/runs/6088770012?check_suite_focus=true#step:6:23)3:21
    at file:///home/runner/work/assemblyscript/assemblyscript/build/assemblyscript.debug.js:2238:3
    at ModuleJob.run (node:internal/modules/esm/module_job:198:25)
    at async Promise.all (index 0)
    at async ESMLoader.import (node:internal/modules/esm/loader:409:[24](https://github.com/AssemblyScript/assemblyscript/runs/6088770012?check_suite_focus=true#step:6:24))
    at <anonymous> (/home/runner/work/assemblyscript/assemblyscript/cli/index.js:47:20)
```

romdotdog:
Only one thing on the logs caught my eye: `Node v18.0.0` as opposed to the first run which used `Node v17.9.0`. Needs further testing.

dcodeIO:
Fixed the CI error in #2274. Merging main should resolve the error in this PR as well.

dcodeIO:
Thanks!

PR #2272, opened 2022-04-21
fix: the function signatures of two interfaces are the same #1855

### LINKED ISSUE #1855: When the function signatures of two interfaces are the same, they cannot be implemented at the same time
```ts
class A implements Foo, Bar {
    foo(): void {}
}

interface Foo {
    foo(): void;
}

interface Bar {
    foo(): void;
}

new A();
```

```
ERROR TS2394: This overload signature is not compatible with its implementation signature.

     foo(): void;
     ~~~~~~~~~~~
 in index.ts(12,5)

     foo(): void;
     ~~~~~~~~~~~
 in index.ts(8,5)

FAILURE 1 compile error(s)
```

This makes some of my interfaces unable to be reused, and even produces API conflicts.

technohippy:
⯈fix #1855


src/types.ts
```diff
@@ -835,15 +835,17 @@ export class Signature {
...
+
+    if (!isInterface) {
+      // check `this` type
+      var thisThisType = this.thisType;
+      var targetThisType = target.thisType;
+      if (thisThisType) {
+        if (!targetThisType || !thisThisType.isAssignableTo(targetThisType)) return false;
+      } else if (targetThisType) {
+        return false;
+      }
```
dcodeIO, line 848: One case I could imagine where skipping the check entirely is an override of `(this: i32, ...): ...`, where then there would be two classes extending the same interface with an incompatible `this` type. Perhaps keep a check for managed-ness / strict assignability of the types?

dcodeIO:
Has been tackled alongside #2548 iirc. The test used in this PR passes now.

PR #2273, opened 2022-04-21
Add global types recognition for advanced developers

jtenner:
When dealing with adding global types for end users, as-pect and cli tools have always relied on adding something like this to an end user's project:

```ts
/// <reference path="../path/to/types.d.ts" />
```

I instead propose standardizing recognizing types that follow the `../node_modules/**/assembly/global.d.ts` pattern out of the box. This of course is not a perfect solution, and I am amenable to other solutions in the same vein as this one, however, this one seems like the simplest way to fix a long standing problem for cli developers using the compiler programmatically.

⯈
⯈
⯈

### DIFF
--- a/cli/tsconfig.json
+++ b/cli/tsconfig.json
@@ -1,7 +1,6 @@
 {
   "extends": "../tsconfig-base.json",
   "include": [
-    "./**/*.ts",
-    "../node_modules/**/assembly/global.d.ts"
+    "./**/*.ts"
   ]
 }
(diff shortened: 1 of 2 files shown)


dcodeIO:
Is `cli/tsconfig.json` the right file for this change? A typical project would have an `assembly/tsconfig.json` extended from `assemblyscript/std/assembly.json` as per what `asinit` generates.

jtenner:
Did I change the correct file? I'm sorry if it was wrong. I would love some help.

dcodeIO:
It could be that the correct file is the one `asinit` generates. Iirc, as soon as the earliest tsconfig has an `include` entry, it overrides what's in the extended base file, not add to it - but not 100% sure. Could you test? :)

PR #2278, opened 2022-04-29
Fix: Incorrect type declaration about 'Number'

tari404:
In #2260 , `Number` was declared as `typeof F64` (equivalent to `typeof _Float`), which causes all numbers to be considered as a `Class` with some methods that don't belong to it, just like:

image

I updated the declaration of `Number` according to the [document](https://www.assemblyscript.org/stdlib/number.html#instance-members). I think this is a small change, so I submit PR directly.

⯈
⯈
⯈

### DIFF
--- a/std/assembly/index.d.ts
+++ b/std/assembly/index.d.ts
@@ -1425,6 +1425,7 @@ declare const F32: typeof _Float;
 declare const F64: typeof _Float;
 /** Alias of F64. */
 declare const Number: typeof F64;
+declare type Number = _Float;
 
 // User-defined diagnostic macros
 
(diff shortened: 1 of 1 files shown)


dcodeIO:
Wondering if it is sufficient to change

```ts
declare type Number = typeof F64;
```

to

```ts
declare type Number = F64;
```

to make it the desired alias, including all properties?

tari404:
> Wondering if it is sufficient to change
> 
> ```ts
> declare type Number = typeof F64;
> ```
> 
> to
> 
> ```ts
> declare type Number = F64;
> ```
> 
> to make it the desired alias, including all properties?

I think it's better, *but it should be:

```ts
declare type Number = _Float;
```

Issue #2262, opened 2022-04-14
not support windows path


HerrCai0907:
```typescript
// assembly/index.ts
import {a} from "../other/test"
export function add(a: i32, b: i32): i32 {
  return a + b;
}

// other/test.ts
export const a:i32 = 2
```

`npx asc ./assembly/index.ts` works
`npx asc .\assembly\index.ts` does not work

ERROR TS6054: File '/assembly/other/test.ts' not found.

 import {a} from "../other/test"
                 ~~~~~~~~~~~~~~~
 in /assembly//index.ts(2,17)

FAILURE 1 parse error(s)

PR #2279, opened 2022-05-01
fix: support windows format path (#2262)

HerrCai0907:
⯈using `npx asc` will transform the path from `a\b` to `a\\b`, so the entry path replace reg should be changed from `/\\/g` to `/\\+/g`

### DIFF
--- a/cli/index.js
+++ b/cli/index.js
@@ -643,17 +643,18 @@ export async function main(argv, options) {
 
   // Include entry files
   for (let i = 0, k = argv.length; i < k; ++i) {
-    const filename = argv[i];
-    let sourcePath = String(filename)
-      .replace(/\\/g, "/")
-      .replace(extension_re, "")
-      .replace(/[\\/]$/, "");
+    const filename = String(argv[i]);
 
     // Setting the path to relative path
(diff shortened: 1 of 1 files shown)


dcodeIO:
Wouldn't it be more correct to specify `a\b` on Windows, so in `process.argv` one gets a singly-escaaped `"a\\b"`?

HerrCai0907:
> Wouldn't it be more correct to specify `a\b` on Windows, so in `process.argv` one gets a singly-escaped `"a\\b"`?

It seems that npm do something in windows. 
In windows when I use `node .\node_modules\assemblyscript\bin\asc.js .\assembly\index.ts`, I will got argv like `["C:\\Program Files\\nodejs\\node.exe","D:\\as\\node_modules\\assemblyscript\\bin\\asc.js",".\\assembly\\index.ts"]`
but when I use `npx asc .\assembly\index.ts`, I will got argv like `["C:\\Program Files\\nodejs\\node.exe","D:\\as\\node_modules\\assemblyscript\\bin\\asc.js",".\\\\assembly\\\\index.ts"]`

HerrCai0907:
I find `path.normalize` can solve this issue and better readability than RegExp

Issue #2286, opened 2022-05-10
Enhancement: Optimize the build flow, release build and debug build


JesseCodeBones:
We found in the build js, the default settings are include minify and source map feature.
it is a little bit confusing for user want debug build or release build.
so it is better to add an option for build script to define if the build is release build or debug build.
for example:
`npm run build:debug` means debug version.
`npm run build` means release version.

here is the current build configuration in `scripts/build.js`:
```
const common = {
  target: "esnext",
  platform: "node",
  format: "esm",
  external: [
    "assemblyscript",
    "binaryen",
    "long"
  ],
  legalComments: "none",
  bundle: true,
  sourcemap: true,
  treeShaking: true,
  minify: true,
  watch,
  incremental: watch
};
```

JesseCodeBones:
I prepared a PR, please kindly review, thank you

PR #2288, opened 2022-05-10
Optimize build flow, divide default build into release build and debu…

JesseCodeBones:
`npm run build:debug` means debug version (un-minified js files, contains the source maps)
`npm run build:release` means release version (source maps without source content, minified js files)

issue link:
https://github.com/AssemblyScript/assemblyscript/issues/2286

Signed-off-by: Jesse <xiang19890319@gmail.com>

### DIFF
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -10,6 +10,8 @@ import { stdoutColors } from "../util/terminal.js";
 const require = createRequire(import.meta.url);
 const dirname = path.dirname(fileURLToPath(import.meta.url));
 const watch = process.argv[2] === "--watch";
+const debug = process.argv[2] === "--debug";
+const release = process.argv[2] === "--release";
 
 function prelude(name) {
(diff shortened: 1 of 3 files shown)


scripts/build.js
```diff
@@ -10,6 +10,8 @@ import { stdoutColors } from "../util/terminal.js";
 const require = createRequire(import.meta.url);
 const dirname = path.dirname(fileURLToPath(import.meta.url));
 const watch = process.argv[2] === "--watch";
+const debug = process.argv[2] === "--debug";
+const release = process.argv[2] === "--release";
```
dcodeIO, line 14: Should there perhaps be a default here if no build type is specified? Seems that right now this would result in `debug = false` and `release = false`. For example:

```suggestion
const debug = process.argv[2] === "--debug";
const release = process.argv[2] === "--release" || !debug;
```

Issue #2289, opened 2022-05-11
Enhancement: Generate test coverage report after npm test


JesseCodeBones:
Currently, the project do not have test coverage feature, and we donot know if the test is enough or not.
So I suggest if we can have test coverage feature, it is better for us to control test or CI flow in the future.
I will introduce C8 here:
Here is the website and documentation of C8:
https://www.npmjs.com/package/c8
C8 can generate html report, text report, cobertura report and etc.
Here is report screenshot:
![image](https://user-images.githubusercontent.com/56120624/167804775-f2d25814-8256-4333-9733-ef84a43037a4.png)
![image](https://user-images.githubusercontent.com/56120624/167805015-e29b5bf5-8ef0-4655-846f-80d067e9740a.png)
it also report directly during test run:
![image](https://user-images.githubusercontent.com/56120624/167805219-687c1e8e-9568-4ebc-998c-bff6f860eb84.png)
The report will generated under `./coverage` folder, of cause it should be git ignored.

JesseCodeBones:
I created one PR, please kindly review and decide if we need this feature, thank you:
https://github.com/AssemblyScript/assemblyscript/pull/2290

JesseCodeBones:
I skiped the lib folder and created a new PR, please kindly check, thanks.

dcodeIO:
Optional code coverage has been added in #2517 in spirit of #2294. There's some [documentation](https://github.com/AssemblyScript/assemblyscript/tree/main/tests#coverage). Do you want to keep this issue open to track further progress, or can we close?

JesseCodeBones:
@dcodeIO 
I will close it, for the test progress, I think it is suitable to divide them with different functionalities.

PR #2294, opened 2022-05-13
Enhancement: Generate test coverage reports after npm test (#2289)

JesseCodeBones:
Signed-off-by: Jesse <xiang19890319@gmail.com>

Issue details: https://github.com/AssemblyScript/assemblyscript/issues/2289

Test report screen shot:
image


HerrCai0907:
@dcodeIO @MaxGraey. I think add coverage is helpful during feature development. wdyt?

package.json
```diff
@@ -70,7 +71,9 @@
     "lint": "eslint --max-warnings 0 --ext js . && eslint --max-warnings 0 --ext ts .",
     "build": "node scripts/build",
     "watch": "node scripts/build --watch",
+    "coverage": "ASC_FEATURES=\"*\" c8 npm run test:coverage",
     "test": "npm run test:parser && npm run test:compiler -- --parallel && npm run test:browser && npm run test:asconfig && npm run test:transform",
+    "test:coverage": "npm run test:parser && npm run test:compiler && npm run test:browser && npm run test:asconfig && npm run test:transform",
```
dcodeIO, line 76: Looks like the `test:coverage` step can be included in `coverage` as it isn't very useful on its own. Even better perhaps, if the `test` step could detect that it runs under `c8` and then disables parallel execution - so the `test` step can be reused?

JesseCodeBones: Yes, fixed now

JesseCodeBones: Hi Dcode,
Considering the compatibility of Windows and POSIX system. I prefer the `process.argv`. currently the issue is, `c8` need a something like `zygote process` process, so it cannot use `&&` on its command line.
So I create `coverage` and `test:coverage` tasks, the `test:coverage` will have `--coverage` option for specific test `test:compiler`
The other option is we can create a `test.js` file and use it to control the environment variables and argument variables, with this option, we only need one `coverage` task.
What do you think?
Regards

dcodeIO: Toying around with c8 a little, I found that instead of having a separate `--coverage` option, we can detect that `c8` is being used by checking `process.env.NODE_V8_COVERAGE != null` and then turn off parallel testing. Looks like this allows to reuse the `test` step?

tests/compiler.js
```diff
@@ -501,7 +504,7 @@ function evaluateResult(failedTests, skippedTests) {
 }
 
 // Run tests in parallel if requested
-if (args.parallel && coreCount > 2) {
+if (args.parallel && coreCount > 2 && !args.coverage) {
```
dcodeIO, line 507: For example here:

```suggestion
const isCoverage = process.env.NODE_V8_COVERAGE != null;
if (args.parallel && coreCount > 2 && !isCoverage) {
```

dcodeIO:
The slightly modified PR above is based on this PR with a few slight changes so coverage remains light / optional for now. One would manually set `ASC_FEATURES`, and `c8` is installed on demand, whereas all the necessary configuration is provided. PTAL :)

package.json
JesseCodeBones: @dcodeIO 
yes, we can use this option, I finished the modification

dcodeIO:
Initial code coverage in the spirit of this PR has been added in #2517, with the different that it is kept optional for now. Sorry for the strange co-authored-by in the commit message, I used the signed-off credentials from the first post that apparently link to a different account, hrm.

PR #2295, opened 2022-05-16
fix: template literal crash GC(#2178)

HerrCai0907:
⯈Another solution to fix template literal bug, see PR #2285 


dcodeIO:
Referring to https://github.com/AssemblyScript/assemblyscript/issues/2178#issuecomment-1150863187, since marking the `StaticArray` as transparent can be assumed safe as it cannot escape, it seems appropriate to merge this PR to fix the issue for now and do further investigations as a follow-up. Can you sync with `main` and add a comment on the new line why this is done?

PR #2321, opened 2022-06-14
Array destructuring #2

romdotdog:
# Array destructuring
This is a PR to implement a bare-bones version of array destructuring.

## What does this PR implement?

```ts
let [a, b] = arr;
let [, a, b] = arr;
let [[a]] = arr;
[a, b] = arr;
[foo.bar] = arr;
[foo[1]] = arr;
```

## What doesn't it implement?

```ts
const { foo } = arr;
const [a = 10] = arr; // semantics is still an open question
const [a, ...b] = arr;

function foo([a, b]: i32[]) {};
```

## Implementation details

`DeclarationStatement`'s `name` is now `Expression` instead of `IdentifierExpression`. This requires replacing `DeclaredElement`'s `identifierNode` getter which returns `this.declaration.name` with a new `identifierNode` field that is provided through the super constructor by subclasses. The new `Expression`s are semantically validated in the parser before assignment to the declaration.


romdotdog:
Closed for inadequate communication.

dcodeIO:
I'm truly sorry about how this went.

trusktr:
😢

PR #2323, opened 2022-06-14
fix: use correct ctx infer generic types

### LINKED ISSUE #2322: Generic Type Infer Failed
```typescript
// index.ts
import * as lib from "./lib";
export function test(): void {
  lib.test(new lib.Wrapper<i32>());
}
// lib.ts
export function test<T = i32>(t: Wrapper<T>): void {}
export class Wrapper<T> {
  v: T;
}
```

It will receive this error

```
ERROR TS2304: Cannot find name 'Wrapper'.

 export function test<T = i32>(t: Wrapper<T>): void {}
                                  ~~~~~~~
 in assembly/lib.ts(1,34)
```

HerrCai0907:
⯈Fix #2322
⯈`ctxFlow` is the flow of caller, but generic type inference need the context of callee


HerrCai0907:
@dcodeIO Could you have some time to review this PR?

dcodeIO (approved):
I still wonder why the helper previously wanted a flow, but since it doesn't use it, the change seems good :)

PR #2325, opened 2022-06-17
Fix unreachable in block statement

### LINKED ISSUE #2187: compiler crash with unreachable in BlockStatement
code: 
```typescript
export function test(a: i32): i32 {
  let t = [];
  if (t.length == 0) {
    return 2;
  }
  {
    return unreachable();
  }
}
```

Error Message:

```
[wasm-validator error in function assembly/index/test] unexpected false: local.get must have a valid type - check what you provided when you constructed the node, on 
(local.get $4)
[wasm-validator error in function assembly/index/test] unexpected false: vars must be defaultable, on 
unreachable
FAILURE validate error
    at Object.main (/home/ccc/Desktop/test/as/node_modules/assemblyscript/cli/asc.js:953:23)
    at /home/ccc/Desktop/test/as/node_modules/assemblyscript/bin/asc:19:47
```

HerrCai0907:
⯈fix #2187
⯈support `return unreachable()`
⯈avoid compile wat like (return (unreachable)) and then insert shadowstack instrument like (locals.get (unreachable))

### DIFF
--- a/src/passes/shadowstack.ts
+++ b/src/passes/shadowstack.ts
@@ -658,6 +658,7 @@ class InstrumentReturns extends Pass {
     var stmts = new Array<ExpressionRef>();
     if (value) {
       let returnType = _BinaryenExpressionGetType(value);
+      if (returnType == TypeRef.Unreachable) return;
       let temp = this.parentPass.getSharedTemp(this.currentFunction, returnType);
       // t = value
       stmts.push(
--- a/tests/compiler/return-unreachable.debug.wat
+++ b/tests/compiler/return-unreachable.debug.wat
@@ -1,0 +1,2416 @@
+(module
+ (type $i32_i32_=>_none (func (param i32 i32)))
+ (type $i32_=>_i32 (func (param i32) (result i32)))
+ (type $i32_=>_none (func (param i32)))
+ (type $none_=>_none (func))
+ (type $i32_i32_=>_i32 (func (param i32 i32) (result i32)))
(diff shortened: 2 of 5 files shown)


src/passes/shadowstack.ts
```diff
@@ -658,6 +658,7 @@ class InstrumentReturns extends Pass {
     var stmts = new Array<ExpressionRef>();
     if (value) {
       let returnType = _BinaryenExpressionGetType(value);
+      assert(returnType != TypeRef.Unreachable, "shadowstack pass cannot handle return unreachable");
```
dcodeIO, line 661: If the returned value is unreachable, code generated here cannot be reached anyhow it seems:
```ts
let t = unreachableExpr; // aborts execution
__stack_pointer += frameSize
return t
```
As such, wdyt of covering this issue in the shadow pass directly, say by, if `returnType` is unreachable, doing nothing:
```ts
// Nothing to do if execution terminates before a value is returned
if (returnType == TypeRef.Unreachable) return;
```
Then other code can remain as-is, relying on the assumption that an unreachable value can substitute a concrete value.

HerrCai0907: The new commit has fixed it 😄

dcodeIO: With the check now in the shadow stack pass, can the other changes be reverted? :)

HerrCai0907: Yes, testcase is passing after trying to revert other change. 😃

dcodeIO (approved):

PR #2352, opened 2022-06-30
feat: add type narrow to support better type check

HerrCai0907:


src/compiler.ts
```diff
@@ -7903,6 +7915,14 @@ export class Compiler extends DiagnosticEmitter {
       this.currentType = Type.bool;
       return this.module.unreachable();
     }
+    if (expression.expression.kind == NodeKind.IDENTIFIER) {
+      const identifier = <IdentifierExpression>expression.expression;
+      const element = flow.lookup(identifier.text);
+      if (element && element instanceof TypedElement) {
```
MaxGraey, line 7921: `instanceof` is operator so extra check for null/undefined is unnecessary.
```suggestion
      if (element instanceof TypedElement) {
```

HerrCai0907: Do you think this is a good idea to resolve type narrow?

MaxGraey: Yeah, definitely after `instanceof` guard LHS should handle as non-nullable

src/program.ts
```diff
@@ -3017,6 +3018,15 @@ export abstract class TypedElement extends DeclaredElement {
...
     this.set(CommonFlags.RESOLVED);
   }
+  narrowType(type: Type): void {
+    if (type.isClass) {
+      this.typeStack.push(this.type);
+      this.type = type;
+    }
+  }
+  recoverType(): void {
+    this.type = this.typeStack.length > 0 ? this.typeStack.pop() : this.type;
```
MaxGraey, line 3028:
```suggestion
    if (this.typeStack.length) {
      this.type = assert(this.typeStack.pop());
    }
```

HerrCai0907:
> When I thought about narrowing in the past, the mechanism I had in mind was to add a new `Flow#localTypes` in addition to `Flow#localFlags`, that would include narrowed types if available. Checks would then first look if there is a narrowed type in `Flow#localTypes` and otherwise fall back to the annotated type from before (say through a `Flow#getLocalType(index)` helper). Wondering if going this route would have advantages over a separate concept as proposed here?

I have not found `localTypes` in current flow. However it thinks like a good idea that we can reuse current inherit process. I can try to add this feature.

HerrCai0907:
By the way, we should not only support narrow type for variant like `a.b`.
Although TS support this grammar, but it will cause something strange during running.
```typescript
let c = new C();
let d = c;
c.v = new B();
if (c.v instanceof B) {
  d.v = new A(); // c.v has changed, but compiler don't know
  c.v.foo();
}
```

MaxGraey:
```js
d.v = new A(); // c.v has changed, but compiler don't know
```
well. that's problem on TS. I guess they are fixed this later. I see no reason to refuse it if we can handle this case correctly

dcodeIO:
Quite similar to why nullability checks are limited to locals. Also applies to narrowing if we don't want to run into unsoundness.

```ts
...
let cv = c.v;
if (cv instanceof B) {
  // works
}
```

HerrCai0907:
@dcodeIO @MaxGraey Could you kindly review this PR and new feature?

src/compiler.ts
```diff
@@ -4630,13 +4642,21 @@ export class Compiler extends DiagnosticEmitter {
               expr = module.if(leftExpr, module.i32(1), rightExpr);
             }
           }
+          let inheritCondi =
+            condKind == ConditionKind.TRUE
+              ? ConditionKind.FALSE
+              : condKind == ConditionKind.FALSE
+              ? ConditionKind.TRUE
+              : ConditionKind.UNKNOWN;
```
MaxGraey, line 4650: It will be great to create util something like:
```ts
function invertedCondition(kind: ConditionKind): ConditionKind {
  if (kind == ConditionKind.TRUE) return ConditionKind.FALSE;
  if (kind == ConditionKind.FALSE) return ConditionKind.TRUE;
  return kind;
}
```
and later:
```ts
let inheritCondi = invertedCondition(condKind);
```

src/compiler.ts
```diff
@@ -5947,6 +5968,13 @@ export class Compiler extends DiagnosticEmitter {
       elementExpression,
       contextualType != Type.void
     );
+
+    if (target instanceof TypedElement) {
+      let typedTarget = <TypedElement>target;
```
MaxGraey, line 5973: How about this?
```suggestion
      let typedTarget = target;
```
Btw it will be great remove such explicit casts to check type narrowing in compiler's bootsrtap as well. @dcodeIO  wdyt?

tests/compiler/typenarrow-error.ts
```diff
@@ -0,0 +1,47 @@
...
+
+// TS2339: Property 'b4' does not exist on type 'typenarrow-error/A
+if (value instanceof B || value.b4) {
+}
+
+if (value instanceof B && (value = new A())) {
+  // TS2339: Property 'b5' does not exist on type 'typenarrow-error/A
+  value.b5;
+}
+
```
MaxGraey, line 46: Could you also add test when type narrowing can't deduced? Something like:
```ts
declare function externalBool(): bool

// externalBool may return `true` while `value` isn't instanceof C
if (externalBool() || value instanceof C) {
  // TS2339: Property 'b1' does not exist on type 'typenarrow-error/A'.
  value.b1;
}
```

src/compiler.ts
```diff
@@ -3236,6 +3243,7 @@ export class Compiler extends DiagnosticEmitter {
     // Compile the body assuming the condition turned out true
     var bodyFlow = flow.fork();
     bodyFlow.inheritNonnullIfTrue(condExpr);
+    bodyFlow.inheritNarrowedTypeIfTrue(condExpr);
```
dcodeIO, line 3246: It seems to me that once narrowing is in place, that the nonnull machinery becomes obsolete. Hence wondering if it would be sufficient to upgrade the nonnull machinery to narrowing so there is only one concept in the end?

src/compiler.ts
dcodeIO: Hmm, technically, one could even remove the additional `typedTarget` local. In turn, however, this makes me wonder what would happen with Wasm GC, where on the IR level `target` would still be of type `Element` and every use of `target` after narrowing would need to cast, which introduces a hidden cost to the developer, in particular in untouched builds.

What'd need to be done here is that upon narrowing, what `typedTarget = <TypedElement>target` did explicitly before would need to be done implicitly by the compiler, introducing a new local of the now-narrowed type - but we don't have the machinery for this yet. Quite unfortunate :/

MaxGraey: I think Most MVP of GC proposal will have `ref.cast_static` and friends. See this PR: https://github.com/WebAssembly/binaryen/issues/4149

MaxGraey: I don't think this hidden cost should block the semantics of the language. In any case, Wasm GC is in a very unstable phase and we don't yet know what it will end up looking like. It's too early to worry about that imho.

HerrCai0907: I agree @MaxGraey. For TypeScript developer, automatically narrow type after `instanceof` is more clear than adding a local variant. Also, for most of cases, a static_cast / dynamic_cast after `instanceof` is needed because the developer will do it only when they want to use some method / member in the inherit class. After GC feature, this feature maybe make this cast from dynamic_cast to static_cast

src/compiler.ts
HerrCai0907: Yes, and it can resolve the nonull bug mentioned in the last of #2360.
But I think maybe we can do it one by one instead of doing everything in this PR? It maybe make this PR very hard to review

tests/compiler/typenarrow-error.ts
HerrCai0907: done

src/compiler.ts
HerrCai0907: thanks!

tests/compiler/typenarrow-error.ts
```diff
@@ -0,0 +1,57 @@
...
+declare function externalBool(): bool;
+
+// externalBool may return `true` while `value` isn't instanceof C
+if (externalBool() || value instanceof B) {
+  // TS2339: Property 'b1' does not exist on type 'typenarrow-error/A'.
+  value.b6;
```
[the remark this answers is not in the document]
HerrCai0907: sorry for forgetting it.

src/narrow.ts
```diff
@@ -0,0 +1,282 @@
...
+          aMap.delete(akey);
+        }
+      }
+    }
+    for (let i = 0, k = bKeys.length; i < k; i++) {
+      let key = bKeys[i];
+      let aType = aMap.has(key) ? assert(aMap.get(key)) : null;
+      let bType = assert(bMap.get(key));
+      let mergedType =
+        mode == TypeMergeMode.OR ? typeOr(aType, bType) : typeAnd(aType, bType);
```
MaxGraey, line 103:
```suggestion
      let mergedType = mode == TypeMergeMode.OR 
        ? typeOr(aType, bType) 
        : typeAnd(aType, bType);
```

src/narrow.ts
```diff
@@ -0,0 +1,282 @@
...
+  }
+  merge(other: NarrowedTypeMap, mode: TypeMergeMode = TypeMergeMode.OR): void {
+    let aMap = this.typeMap;
+    let bMap = other.typeMap;
+    let bKeys = Map_keys(bMap);
+    if (mode == TypeMergeMode.AND) {
+      let aKeys = Map_keys(aMap);
+      for (let i = 0, k = aKeys.length; i < k; i++) {
+        let akey = aKeys[i];
+        if (!bKeys.includes(akey)) {
```
MaxGraey, line 93:
```suggestion
        if (!bMap.has(akey)) {
```

src/flow.ts
```diff
@@ -237,6 +244,11 @@ export class Flow {
   localFlags: LocalFlags[] = [];
   /** Field flags on `this`. Constructors only. */
   thisFieldFlags: Map<Field,FieldFlags> | null = null;
+  /** type narrow */
+  narrowedTypes: NarrowedTypeMap = new NarrowedTypeMap();
+  /** conditional type narrow, eg if (condi) */
+  conditionalNarrowedType: TypeNarrowChecker = new TypeNarrowChecker();
```
MaxGraey, line 250: it will be great init checker and narrowedTypes lazily. Not every program uses `instanceof`. And it also add overhead for cloning in `fork` method

MaxGraey (approved):

src/compiler.ts
```diff
@@ -3135,7 +3141,7 @@ export class Compiler extends DiagnosticEmitter {
         }
         if (initExpr) {
           initializers.push(
-            this.makeLocalAssignment(local, initExpr, type, false)
+            this.makeLocalAssignment(local, initExpr, initType ? initType : type, false)
```
MaxGraey, line 3144: Doesn't this work `initType || type` ?

tests/compiler/std/math.release.wat
```diff
@@ -49905,7 +49905,7 @@
    call $~lib/builtins/abort
    unreachable
   end
-  f64.const -nan:0x8000000000000
+  f64.const nan:0x8000000000000
```
MaxGraey, line 49908: Hmm, this some red flag. Why this happening?

HerrCai0907: Maybe it's OK because debug.wat is not changed. But actually i don't know what happened.

src/compiler.ts
HerrCai0907: It works. Let me change all the similar situation by this clear and simple way.

src/compiler.ts
```diff
@@ -9264,13 +9296,13 @@ export class Compiler extends DiagnosticEmitter {
 
     var outerFlow = this.currentFlow;
     var ifThenFlow = outerFlow.fork();
-    ifThenFlow.inheritNonnullIfTrue(condExpr);
+    ifThenFlow.inheritNarrowedTypeIfTrue(condExpr);
```
MaxGraey, line 9299:
```suggestion
    ifThenFlow.inheritNarrowedTypeIfTrue(condExpr);
```

tests/compiler/std/math.release.wat
MaxGraey: Could you sync with main and reinstall dependencies? May be you have some outdated binaryen

HerrCai0907: Already sync with main. But onto solve it.

MaxGraey: I found location in tests:
```ts
assert(test_sqrt(-8.06684839057968084, NaN, 0.0, INVALID));
```
and other cases with negative input values.
After binaryen optimizations and inlining and DAE pass it should be equivalent to this:
```ts
check<f64>(-NaN, NaN, 0) &&
check<f64>(JSMath.sqrt(NaN), NaN, 0, INVALID));
```
But instead becomes:
```ts
check<f64>(+NaN, NaN, 0)
```

MaxGraey: Could you check in your node this line?:
```js
function signbit(x) {
    const F64 = new Float64Array(1);
    const U64 = new BigUint64Array(F64.buffer);
    return F64[0] = x, Boolean(U64[0] >> 63n);
}
console.log(signbit(Math.sqrt(-8.06684839057968084)));
```
Does it print `true`?

MaxGraey: Also does you rebuild test fixtures in `main` branch and check if std/math.ts is modified the same way? If not, then it is not a nondeterminism related to your env/platform and we should look for the problem in this PR.

MaxGraey: Yeah, it looks like it's non-determ on you env. I checkout this branch and rebuild std/math.ts. It reverts sign NaN back. May I ask what is your platform & env (os, cpu, node.js version)?

MaxGraey: It's definitely not a blocker for your PR, we can always port update the fixtures on our end. But I really wonder why this is happening on your machine. Maybe you should report the problem to v8 or binaryen

HerrCai0907: I have try following env:
1. macos arm + node v16.9.1
2. macos arm + node v16.16.0

MaxGraey: Hmm, so it's M1. Interesting...

src/compiler.ts
HerrCai0907: In both main and feature branch, it does not work.
It needs to do some special process for 'or' operation

HerrCai0907:
@dcodeIO Nonnull check has been merged into type narrow logic. Could you kindly review it?

src/compiler.ts
```diff
@@ -5947,6 +5960,7 @@ export class Compiler extends DiagnosticEmitter {
       elementExpression,
       contextualType != Type.void
     );
+    return assignmentExpression;
```
MaxGraey, line 5963: Could you revert this to original variant ` return this.makeAssignment(...)`?

src/flow.ts
```diff
@@ -88,9 +82,7 @@ import {
   uniqueMap
 } from "./util";
 
-import {
-  BuiltinNames
-} from "./builtins";
+import { NarrowedTypeMap, TypeNarrowChecker } from "./narrow";
```
MaxGraey, line 85:
```suggestion
import {
  NarrowedTypeMap,
  TypeNarrowChecker
} from "./narrow";
```

src/flow.ts
```diff
@@ -606,6 +614,65 @@ export class Flow {
...
+    if (type && !type.isReference) return;
+    this.setNarrowedType(element, type);
+    this.conditionalNarrowedType.setAssignType(expr, element, type);
+  }
+  /** set type narrow */
+  setNarrowedType(element: TypedElement, type: Type): void {
+    if (!type.isReference) return;
+    if (this.narrowedTypes == null) {
+      this.narrowedTypes = new NarrowedTypeMap();
+    }
```
MaxGraey, line 628: I guess we could use this now:
```suggestion
    this.narrowedTypes = this.narrowedTypes || new NarrowedTypeMap();
```
Which will simplify further when we will support `||=` and `&&=`

src/flow.ts
```diff
@@ -606,6 +614,65 @@ export class Flow {
...
+    let narrowedTypes = assert(this.narrowedTypes);
```
MaxGraey, line 629: Also `assert` should be unnecessary

src/resolver.ts
```diff
@@ -1275,6 +1275,8 @@ export class Resolver extends DiagnosticEmitter {
       case ElementKind.FIELD: { // someVar.prop
         let variableLikeElement = <VariableLikeElement>target;
         let type = variableLikeElement.type;
+        let narrowedType = ctxFlow.getNarrowedType(variableLikeElement);
+        if (narrowedType) type = narrowedType;
```
dcodeIO, line 1279: Applying type narrowing on fields is risky because any intermediate function call (or property access with a getter/setter) can have the side-effect of assigning a different value to the field, so the narrowed type would no longer hold and the program would statically produce failures. So far, we limited such checks to locals, since in the absence of closures, these cannot be mutated as a side-effect. As such, this change is likely unsafe.

src/flow.ts
```diff
@@ -238,6 +236,11 @@ export class Flow {
   localFlags: LocalFlags[] = [];
   /** Field flags on `this`. Constructors only. */
   thisFieldFlags: Map<Field,FieldFlags> | null = null;
+  /** type narrow */
+  narrowedTypes: NarrowedTypeMap | null = null;
+  /** conditional type narrow, eg if (condi) */
+  conditionalNarrowedType: TypeNarrowChecker = new TypeNarrowChecker();
```
dcodeIO, line 242: Hnnm, can you explain why this additional map is necessary? It seems to me that anything that is conditional would be a `narrowedType` limited to the respective flow it is narrowed in. Say there is an `if (cond)`, then the inner flow within the if would the state after `cond` as `narrowedTypes`, and at the end of the if, the inner flow would be merged with the outer one again, updating the successor flow's `narrowedTypes`. As such I don't quite see what `conditionoalNarrowedTypes` does, as within a flow, there wouldn't be conditionals?

src/resolver.ts
MaxGraey: Yes, perhaps we should comment out with some TODO for now, until we have the necessary infra to check the side effects for Flow. Calling a method to instance this reference or passing the reference to a global function between entry and exit points of this path will be considered a side-effect that will either stop type narrowing or start some kind of approximation or deeper analysis.

src/narrow.ts
```diff
@@ -0,0 +1,428 @@
...
+  collectNarrowedTypeIfFalse(expr: ExpressionRef, flow: Flow, typeMap: NarrowedTypeMap | null = null): NarrowedTypeMap {
+    if (typeMap == null) typeMap = new NarrowedTypeMap();
+    switch (getExpressionId(expr)) {
+      case ExpressionId.Unary: {
+        switch (getUnaryOp(expr)) {
+          case UnaryOp.EqzI32:
+          case UnaryOp.EqzI64: {
+            this.collectNarrowedTypeIfTrue(getUnaryValue(expr), flow, typeMap); // !value -> value must have been true
+
+            break;
```
MaxGraey, line 351:
```suggestion
            this.collectNarrowedTypeIfTrue(getUnaryValue(expr), flow, typeMap); // !value -> value must have been true
            break;
```

src/flow.ts
HerrCai0907: The name may cause some misunderstandings. This is conditional type narrowing checker which will collect the info of type narrowing and decide how type narrowed in then and else branch according to the condition expression. 
Do you think renaming it as `conditionalTypeNarrowChecker` is more straightforward?

src/resolver.ts
HerrCai0907: Ok, I will remove the logic of global and field. And then mark it as TODO.

src/flow.ts
dcodeIO: I see, one is a map and the other is a checker object. Would it be possible to abstract the checker object away? Looks like there's always a `new` one assigned to a `Flow` anyway, so the flow could include the respective methods directly perhaps.

src/flow.ts
HerrCai0907: If type narrow does not effect for Field, It should like this:
```typescript
let narrowedTypes = this.narrowedTypes || new NarrowedTypeMap();
this.narrowedTypes = narrowedTypes;
```

src/flow.ts
HerrCai0907: > Would it be possible to abstract the checker object away?
Abstract it away is difficult because map relies on each flow but check is for the flow and the forked flow. Actually it should not create a new checker for each flow. I will move it into constructor and inherit it during fork.

src/flow.ts
```diff
@@ -218,9 +218,10 @@ export class Flow {
 
   private constructor(
     /** Function this flow belongs to. */
-    public parentFunction: Function
+    public parentFunction: Function,
+    typeNarrowChecker: TypeNarrowChecker | null
   ) {
-    /* nop */
+    this.typeNarrowChecker = typeNarrowChecker || new TypeNarrowChecker();
   }
```
MaxGraey, line 225: what about just use default param?
```ts
    ...
    typeNarrowChecker: TypeNarrowChecker = new TypeNarrowChecker();
  ) {
    /* nop */
  }
```

MaxGraey:
Btw such flow mechanics is quite typical:
```ts
  var thenFlow = flow.fork();
  this.currentFlow = thenFlow;
  thenFlow.inheritNarrowedTypeIfTrue(condExpr);
```
I'm wondering could we add some method for Flow which include this routine?
```ts
class Flow {
  ...
  inBranch(condExpr: Expression, inheritForTrue: bool = true): Flow {
    var other = this.fork();
    if (inheritForTrue) {
      other.inheritNarrowedTypeIfTrue(condExpr);
    } else {
      other.inheritNarrowedTypeIfFalse(condExpr);
    }
    return other;
  }
}
```
And later simplify to:
```ts
this.currentFlow = flow.inBranch(condExpr);
```
Not sure about naming. Probably you could call all of this better

src/flow.ts
```diff
@@ -332,6 +332,18 @@ export class Flow {
     return branch;
   }
 
+  /** Fork this flow to a child flow in branch */
+  forkTrueBranch(condExpr: ExpressionRef, resetBreakContext: bool = false): Flow {
+    let branch = this.fork(resetBreakContext);
+    branch.inheritNarrowedTypeIfTrue(condExpr);
+    return branch;
```
MaxGraey, line 339: If `inheritNarrowedTypeIfTrue` refactor to return `this` (Flow) it could be simplified to:
```ts
  forkTrueBranch(condExpr: ExpressionRef, resetBreakContext: bool = false): Flow {
    return this.fork(resetBreakContext).inheritNarrowedTypeIfTrue(condExpr);
 }
```

HerrCai0907: I think it will reduce the readability. Why inherit conditional return a flow? It will cause additional attentions to find what return flow means.

MaxGraey: I see. ok in this case keep this as is

MaxGraey (approved):

HerrCai0907:
@dcodeIO Could this feature merge?

src/compiler.ts
```diff
@@ -2994,6 +2991,8 @@ export class Compiler extends DiagnosticEmitter {
           );
           pendingElements.delete(dummy);
           flow.freeScopedDummyLocal(name);
+
+          initType = this.currentType;
```
dcodeIO, line 2995: Due to compiling with `Constraincts.CONV_IMPLICIT`, I think `initType` is always the same as `type` here. Sure that the additional `initType` variable is needed?

src/compiler.ts
```diff
@@ -4566,13 +4564,15 @@ export class Compiler extends DiagnosticEmitter {
...
           }
+          flow.inheritBranch(rightFlow, condKind);
           this.currentFlow = flow;
           this.currentType = Type.bool;
 
         } else {
           rightExpr = this.compileExpression(right, leftType, inheritedConstraints | Constraints.CONV_IMPLICIT);
           rightType = this.currentType;
           rightFlow.freeScopedLocals();
+          flow.inheritBranch(rightFlow);
```
dcodeIO, line 4575: Let's move this line one up, so it happens before `freeScopedLocals` just in case, that is, avoiding undesired side-effects from freeing.

src/flow.ts
```diff
@@ -905,20 +987,22 @@ export class Flow {
...
-    // local flags
+    // local flags and type
     var thisLocalFlags = this.localFlags;
     if (leftFlags & FlowFlags.TERMINATES) {
       if (!(rightFlags & FlowFlags.TERMINATES)) {
         let rightLocalFlags = right.localFlags;
         for (let i = 0, k = rightLocalFlags.length; i < k; ++i) {
           thisLocalFlags[i] = rightLocalFlags[i];
         }
+        this.narrowedTypes = right.narrowedTypes;
```
dcodeIO, line 998: I wonder if assigning the narrowed types map by reference here can lead to issues. Say, one of the flows is updated later on, then the other would be updated as well, which might be incorrect? Copying instead would likely be safer.

src/flow.ts
```diff
@@ -905,20 +987,22 @@ export class Flow {
...
       }
     } else if (rightFlags & FlowFlags.TERMINATES) {
       let leftLocalFlags = left.localFlags;
       for (let i = 0, k = leftLocalFlags.length; i < k; ++i) {
         thisLocalFlags[i] = leftLocalFlags[i];
       }
+      this.narrowedTypes = left.narrowedTypes;
```
dcodeIO, line 1005: Same here: Perhaps copy?

src/flow.ts
```diff
@@ -991,237 +1089,39 @@ export class Flow {
...
+    let thisNarrowedTypes = this.narrowedTypes;
     // below, only teeLocal/getLocal are relevant because these are the only expressions that
     // depend on a dynamic nullable state (flag = LocalFlags.NONNULL), while everything else
     // has already been handled by the nullable type check above.
     switch (getExpressionId(expr)) {
       case ExpressionId.LocalSet: {
         if (!isLocalTee(expr)) break;
         let local = this.parentFunction.localsByIndex[getLocalSetIndex(expr)];
-        return !local.type.isNullableReference || this.isLocalFlag(local.index, LocalFlags.NONNULL, false);
+        let localType = thisNarrowedTypes ? thisNarrowedTypes.get(local) : null;
```
dcodeIO, line 1111: Can the `.get` fail here? If so, it would return `undefined` in JS and abort in Wasm. Looks like this needs another check that the `local` in fact exists?

src/flow.ts
```diff
@@ -991,237 +1089,39 @@ export class Flow {
...
+        localType = localType || local.type;
+        return !localType.isNullableReference;
       }
       case ExpressionId.LocalGet: {
         let local = this.parentFunction.localsByIndex[getLocalGetIndex(expr)];
-        return !local.type.isNullableReference || this.isLocalFlag(local.index, LocalFlags.NONNULL, false);
+        let localType = thisNarrowedTypes ? thisNarrowedTypes.get(local) : null;
```
dcodeIO, line 1117: Same as above: What if the `.get` fails?

src/compiler.ts
HerrCai0907: consider this testcase. Actually `initType` and `type` is not always same
```tyepscript
export function testInit(a: Ref): void {
  let c: Ref | null = requireNonNull(a);
  if (isNullable(c)) ERROR("should be non-nullable");
}
```

src/flow.ts
HerrCai0907: For detail scenes to be used, copy by reference in current compile is no problem because it is only used in `TernaryExpression` and `IfStatement`. But for better scalability clone will be safer.

HerrCai0907:
@dcodeIO Could you kindly review this PR?

src/compiler.ts
```diff
@@ -2668,9 +2669,9 @@ export class Compiler extends DiagnosticEmitter {
 
     // Compile ifTrue assuming the condition turned out true
     var thenStmts = new Array<ExpressionRef>();
-    var thenFlow = flow.fork();
+    var thenFlow = flow.forkTrueBranch(condExpr);
     this.currentFlow = thenFlow;
-    thenFlow.inheritNonnullIfTrue(condExpr);
+
```
MaxGraey, line 2674: Btw we pretty often have this pattern:
```ts
var someFlow = flow.forkTrueBranch(condExpr);
this.currentFlow = someFlow;
...
someFlow.freeScopedLocals();
this.currentFlow = flow; // restore old flow
```
What if make this less error-prone? And automatically call `freeScopedLocals`. For that we should refator `currentFlow` to get / set accessor and use this impl:
```ts
set currentFlow(newFlow: Flow) {
  if (this._currentFlow) this._currentFlow.freeScopedLocals(); // It already have guard if locals is empty
  this._currentFlow = newFlow;
}
```
also modify `forkTrueBranch`:
```ts
forkTrueBranch(...): Flow {
   ...
   this._currentFlow = forkedFlow;
   return forkedFlow;
}
```
Then code below will be a little bit simpler and more safety:
```ts
var someFlow = flow.forkTrueBranch(condExpr);
// don't need set this.currentFlow = someFlow;
...
this.currentFlow = flow; // restore old flow
// don't need to call someFlow.freeScopedLocals();
```
@dcodeIO wdyt?

MaxGraey: Another alternative solution may look like this:
```
this.currentFlow.lock(); // save prev flow
var someFlow = flow.forkTrueBranch(condExpr);
// implicitly set currentFlow to `someFlow` in forkTrueBranch

this.currentFlow.unlock(); // call someFlow.freeScopedLocals() and resore previousle saved `flow` to `currentFlow`
```

MaxGraey: From the discussion with Daniel, we decided picked a second option, but to do it in a separate PR. So mark this as resolved

dcodeIO:
I'm having a hard time reviewing the PR, and would like to suggest to split it up a little. As a start, there has been the refactoring from `fork + X` to `forkTrueBranch` and `forkFalseBranch`. Can you make this a separate PR, so we have this in already as it seems useful independently? There also appear to be a few fixes here, which would be good to merge independently as well? Unless I missed another refactoring, afterwards we'd be left with just the narrowing changes.

From there on it gets a little tricky, as I am not confident about having an out-of-line `src/narrow.ts`. Typically, what we do there is to place utility data structures into `util/`, which here could be an `util/narrowing.ts` with the specialized data structure. However, the data structure now absorbs the logic from the `inheritXY` methods that previously lived in `Flow`, but I'd like to keep the logic in `Flow`. Suggesting to do the other ones first, so I can look at this again separately and see. Would this be OK for you?

HerrCai0907:
> I'm having a hard time reviewing the PR, and would like to suggest to split it up a little.

Yes, I will split it into several PR. But it needs some time because I am preparing my graduation thesis now😄

PR #2360, opened 2022-07-04
fix: Fix `||` and `&&` will not inherit branch status

### LINKED ISSUE #2359: `&&` and `||` expression will not change the `NONNULL` flag
This code should not pass the compile.
```
  let value: A | null = new A();
  if (value != null) {
    // value = null;  // cause compile error correctly
    true && (value = null); // compile passed, but value is changed to zero now
    let a1 = value.a1;
  }
}
```

HerrCai0907:
⯈Fix #2359 

### DIFF
--- a/src/compiler.ts
+++ b/src/compiler.ts
@@ -4566,14 +4566,15 @@ export class Compiler extends DiagnosticEmitter {
               expr = module.if(leftExpr, rightExpr, module.i32(0));
             }
           }
+          flow.inheritBranch(rightFlow, condKind);
           this.currentFlow = flow;
-          this.currentFlow.inheritBranch(rightFlow, condKind);
           this.currentType = Type.bool;
 
         } else {
           rightExpr = this.compileExpression(right, leftType, inheritedConstraints | Constraints.CONV_IMPLICIT);
           rightType = this.currentType;
           rightFlow.freeScopedLocals();
+          flow.inheritBranch(rightFlow);
           this.currentFlow = flow;
 
           // simplify if copying left is trivial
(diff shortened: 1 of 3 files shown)


src/compiler.ts
```diff
@@ -4597,6 +4597,7 @@ export class Compiler extends DiagnosticEmitter {
           }
           this.currentType = leftType;
         }
+        this.currentFlow.inheritBranch(rightFlow);
```
dcodeIO, line 4600: It's a little more complex than this I guess. For instance, if the LHS is always `true`, the parent flow would inherit the right flow unconditionally (not a branch), while if it is always `false`, the parent flow would not inherit the right flow since it doesn't execute.

src/compiler.ts
```diff
@@ -4661,6 +4662,7 @@ export class Compiler extends DiagnosticEmitter {
...
```
dcodeIO, line 4665: Similarly, if the LHS is always `false`, the parent flow would inherit the right flow unconditionally, while if it is always `true`, it would not inherit anything since it never executes.

src/compiler.ts
HerrCai0907: Yes, `inheritBranch` is a little conservative. It can spilt with difference LHS/RHS `ConditionKind`

HerrCai0907:
@dcodeIO It seems like a little bit complex. After this fix, It still have problem.

```
  let a: string| null = "hhh"
  if (a && !(a = null)) {
    trace(a.length.toString())
  }
}
```

wasm://wasm/a4e78d1a:1

RuntimeError: memory access out of bounds
    at wasm://wasm/a4e78d1a:wasm-function[2]:0x18c
    at wasm://wasm/a4e78d1a:wasm-function[58]:0x14df
    at file:///home/ccc/Desktop/demo/as/tests/index.js:3:1

PR #2439, opened 2022-08-17
BREAKING CHANGE: Remove support for WASI

dcodeIO:
For us, WASI's exclusionary design is a regression and even though we voiced our needs, suggestions and concerns many times over the last years, its subgroup keeps ignoring, attacking and excluding us. Since there is no reasonable path forward anymore and future expectations are solely negative, the reasonable consequence is to stop officially supporting WASI's harmful efforts.

For those who have no other choice than to use such interfaces and related products, we are working on a custom shim that will become available under https://github.com/AssemblyScript/wasi-shim somewhen after 0.21 has been released.


trusktr (approved):
Anyone who needs WASI support can polyfill web APIs on that side. 👍

JairusSW:
I totally agree. 👍🏼

dcodeIO:
Thanks for the feedback! Still working on the shim, but it's almost ready so it should be released soon™ 🙂

JairusSW:
Awesome! I have been looking forward to seeing these changes appear in AS for quite a while now :)

PR #2449, opened 2022-08-19
fix: override keyword shouldn't throw error in multilevel inherit 

HerrCai0907:
fix override keyword in following case

```
class A {
  foo(): void {}
}
class B extends A {
  bar(): void {}
}
class C extends B {
  override foo(): void {}. // should not throw error
}
```


dcodeIO:
Wondering why this didn't work before. Can you provide some context of what was missing exactly? Would assume that `B` inhereits `foo` from `A`, and then `C` upon extending sees `foo`, overriding it.

HerrCai0907:
> Wondering why this didn't work before. Can you provide some context of what was missing exactly? Would assume that `B` inhereits `foo` from `A`, and then `C` upon extending sees `foo`, overriding it.

But B's `baseInstanceMembers` don't have A's member, so need to search all the node in inherit chain and throw the error at the end.

dcodeIO:
Oh, I see. Makes me think this is either an oversight, or perhaps we should look up from `instanceMembers` if these have all of them? I wonder what I was thinking when I wrote that code :)

dcodeIO:
Looking more closely, I think `markVirtuals` isn't the right place to perform the check / emit the diagnostic. That's a rather specific method, and it was likely an oversight that the prior check was added there. Looks like the check should rather be performed when resolving a class.

dcodeIO (approved):

PR #2495, opened 2022-09-02
Add two to four bytes utf8 character tests

JesseCodeBones:
Currently the repo missed the tests for 2 bytes utf-8 character, 3 bytes utf-8 character and 4 bytes utf-8 character (4 bytes utf-8 character cannot bypass the parser phase, not sure if it is by design, for example charactor: 𝆑).

This PR attampts to add tests for global parameter name with utf-8 character, 3 bytes utf-8 character, AS will call allocString and alloc this utf-8 string to binary.
image


dcodeIO:
Regarding the 4-byte character problem, see #2522.

Issue #2508, opened 2022-09-06
Multi implements / extends issue


JesseCodeBones:
Hi Maintainers,
I found the class resolver for implements only check the name of the property or function, with will cause compile error during multi implements or extends.
here is demo code:
```
interface TwoDimensionalCoordinates {
    x: i32;
    y: i32;
}

interface ThreeDimensionalCoordinates {
    x: i32;
    y: i32;
    z: i32;
}

class RealCoordinates implements TwoDimensionalCoordinates, ThreeDimensionalCoordinates {
    x: i32 = 2;
    y: i32 = 2;
    z: i32 = 2;
}
let a = new RealCoordinates();
```
I think if the parents vitual function or parameter has same name should bypass this check.
WDYT?

the main check logic should be in `isCompatibleOverride` of `reslover`.

JesseCodeBones:
Proposal:
1. parameter
1.1 with different type, raise error. ❌
1.2 with same type, bypass :heavy_check_mark: 
2. function
2.1 with different signature, raise error (typescript seems also do this) ❌
2.2 with same signature, bypass :heavy_check_mark:

JesseCodeBones:
Also for the `FunctionPrototype`:
This is one Demo code:
```
interface Car {
    getHeight():i32;
}

interface Mechanical {
    getHeight():i32;
}

class Nissan implements Mechanical, Car {
    getHeight():i32{return 1;}
}
let b = new Nissan();
```

PR #2510, opened 2022-09-07
Fix: multi implements issue

JesseCodeBones:
From issue [2508](https://github.com/AssemblyScript/assemblyscript/issues/2508)


src/program.ts
```diff
@@ -3037,6 +3055,26 @@ export abstract class DeclaredElement extends Element {
...
+    //   if (baseGetterInsance && baseSetterInsance) {
+    //     if (!selfField.internalGetterSignature.isAssignableTo(baseGetterInsance.signature)
+    //       || !selfField.internalSetterSignature.isAssignableTo(baseSetterInsance.signature)) {
+    //       return false;
+    //     }
+    //   } else {
+    //     return false;
+    //   }
+    //   return true;
+    // }
```
dcodeIO, line 3077: This is most likely not going to work because fields and properties have different code generation. In particular, getters and setters or properties are function calls, whereas a field is just load/store to a memory location. Getter and setter functions can be overloaded, yielding a virtual stub, while fields do not participate in overloading. Suggesting to revert what I commented out here, and update the tests accordingly so fields and properties are not mixed for now.

To give an outlook for the future, we'll likely want to remove the concept of fields and transparently make these properties with automatically generated getter and setter functions (a few bits into this direction already exist, but it's not yet finished), where these getters and setters are consequently used whenever the field is accessed. Conceptionally, fields then become properties with generated accessors (that can participate in virtual stubs) and a storage location behind. Hence we'll likely not need the `Field` kind / class anymore, and instead have `Property` with a flag `IS_FIELD` or so. Out of scope for this PR, though :)

src/types.ts
```diff
@@ -926,17 +926,27 @@ export class Signature {
...
+        }
+      } else if (targetThisType) {
+        return false;
+      }
+    } else {
+      // check kind of `this` type
+      if (thisThisType) {
+        if (!targetThisType || thisThisType.kind != targetThisType.kind || thisThisType.isReference != targetThisType.isReference) {
+          return false;
+        }
```
dcodeIO, line 946: Here we still need to make sure that the `this` type matches enough so having an overload will always result in a functioning callable signature. Using the `kind` here, so `this` is both times `usize` for example, and both times either a reference or not. Likely needs more checks, so managed and unmanaged aren't mixed etc., which could then become a separate helper function.

tests/compiler/class-implements.ts
```diff
@@ -18,3 +22,51 @@ class C extends B implements I {
...
+
+let h = new H();
+assert(h.foo() == 3);
+
+class K extends D implements I {
+  foo(): i32 { return 4; }
+}
+
+let k = new K();
+assert(k.foo() == 4);
```
dcodeIO, line 72: With the commented out field/property check, some of the tests don't work anymore. Suggesting to use properties exclusively so these pass again.

dcodeIO (commented):
Took a look at the branch, and made a couple changes (remember to `git pull` on your end). A few observations below. I think with those addressed the PR can be merged, even though it then does not cover all cases.

src/types.ts
JesseCodeBones: Hi Dcode,
how about the below function? for the multi implements / overrides cases, do we only need to consider with reference type?
```
  isLenientlyAssignableTo(target: Type) {
    if (target) {
      if (this.isReference && target.isReference) {
        let thisClass = this.getClass();
        let targetClass = target.getClass();
        if (thisClass && targetClass) {
          // extends ThisClass implements TargetInterface
          // implements ThisInterface, TargetInterface
          if (thisClass.kind == ElementKind.CLASS || thisClass.kind == ElementKind.INTERFACE) {
            if (targetClass.kind == ElementKind.INTERFACE) {
              return true;
            }
          }
        }
      }
    }
    return false;
  }
```
@dcodeIO

dcodeIO: Needs to be a little more complex I think. Perhaps `.isInternalReference`, so external and internal aren't mixed. We'd also need to check that unmanaged and managed aren't mixed (`type.isManaged == type.isManaged`), since both are classes. And the method in question might have an annotated `this` type on its own, so it makes sense to check the method, not just the class, hmm.

JesseCodeBones: Hi Dcode,
yeah, I changed the name from `isLenientlyAssignableTo` to `isCompatibleTo`, so it make sense now, please have a look the below method:
```
  /** Tests if a value of this class is compatible to the target class / interface in the multi extends / implements situation. */
  isCompatibleTo(target: Type) {
    if (target) {
      if (this.isInternalReference && target.isInternalReference && this.isManaged && target.isManaged) {
        let thisClass = this.getClass();
        let targetClass = target.getClass();
        if (thisClass && targetClass) {
          // extends ThisClass implements TargetInterface
          // implements ThisInterface, TargetInterface
          if (thisClass.kind == ElementKind.CLASS || thisClass.kind == ElementKind.INTERFACE) {
            if (targetClass.kind == ElementKind.INTERFACE) {
              return true;
            }
          }
        }
      }
    }
    return false;
  }
```

JesseCodeBones: Hi Dcode,
I pushed the changes, can you take a look? thank you.
@dcodeIO

dcodeIO:
Thanks!

PR #2515, opened 2022-09-15
fix: add more test coverage for binary expression

JesseCodeBones:
Hi Maintainer,
I found the some tests are missed for binary expression, please decide if we need to coverage this part? and please review the below tests are suitable?
image

Regards

### DIFF
--- a/tests/compiler/binary-error.json
+++ b/tests/compiler/binary-error.json
@@ -1,0 +1,24 @@
+{
+  "asc_flags": [
+  ],
+  "stderr": [
+    "TS2365: Operator '<' cannot be applied to types 'binary-error/A | null' and 'i8'.",
+    "TS2365: Operator '>' cannot be applied to types 'binary-error/A | null' and 'i8'.",
+    "TS2365: Operator '<=' cannot be applied to types 'binary-error/A | null' and 'i8'.",
+    "TS2365: Operator '>=' cannot be applied to types 'binary-error/A | null' and 'i8'.",
+    "TS2365: Operator '==' cannot be applied to types 'binary-error/A | null' and 'i8'.",
+    "TS2365: Operator '!=' cannot be applied to types 'binary-error/A | null' and 'i8'.",
+    "TS2365: Operator '+' cannot be applied to types 'i8' and '() => void'.",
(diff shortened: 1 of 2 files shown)


JesseCodeBones:
@dcodeIO @MaxGraey
Hi Maintainers,
Actually we got a plan that help ASC to improve the test coverage, I understand we prefer the fuzz test as ASC is a compiler, but we also found some issues during the tests adding, so I think it is also benefit if we could review or cover the un-tested code block.
for example: [2403](https://github.com/AssemblyScript/assemblyscript/pull/2403),[2428](https://github.com/AssemblyScript/assemblyscript/pull/2428),[2379](https://github.com/AssemblyScript/assemblyscript/pull/2379) and [2510](https://github.com/AssemblyScript/assemblyscript/pull/2510).
If your also think it is valuable to continue this work, I will devide the test coverage PR with different functionalities, so it could be easy to review and find if they are meaningful tests.
So can you give an advice? Thank you.
Here are the coverage PRs:
[2515](https://github.com/AssemblyScript/assemblyscript/pull/2515),[2495](https://github.com/AssemblyScript/assemblyscript/pull/2495),[2467](https://github.com/AssemblyScript/assemblyscript/pull/2467),[2444](https://github.com/AssemblyScript/assemblyscript/pull/2444),[2294](https://github.com/AssemblyScript/assemblyscript/pull/2294).
Regards

dcodeIO:
Hey @JesseCodeBones :)

I agree that improving test coverage is a good idea, and appreciate that you are working on this, thanks! There is one somewhat subtle aspect on my mind in addition: There are two builds of the AS compiler, one compiled with `tsc` to JS and one compiled with `asc` to AS. The test coverage right now relies on the JS variant to function IIUC, which makes sense since that's what we use in the test suite today. On this background, it would be good if coverage reporting could eventually include Wasm testing as well - just for your information, though, as right now I think the JS variant is sufficient.

I guess we'd start with #2294 and then merge the concrete additions to tests on top, am I seeing this correctly?

JesseCodeBones:
@dcodeIO 
Thank you for your rapid reply:)
From my understanding, yes, corrent tests relies on JS variant and the map file will do the source code mapping to the correct `ts` source file, if the test coverage report is the first step, I think it is a great start.
Regards

PR #2532, opened 2022-10-09
use LinkedList solution for minimal runtime build

JesseCodeBones:
Hi Maintainer,
I prepared a lighter memory allocator for ASC, the algorithm is following [LibMemory](https://github.com/embeddedartistry/libmemory).
Comparing with current TLSF solution, it has below advantages:

1. It can use `page fault` ([Link](https://en.wikipedia.org/wiki/Page_fault)) feature of OS / Runtime. 
2. Linearly memory grow
3. Smaller size of target files (wasm, wast and map file)
4. Easier logic for `free`, which is an advantage when manage smaller amount of object at same time.
5. It is compatible to current GC solution (`mmInfo` not needed for LibMemory, but I did not remove it yet)

Disadvantages:

1. O(n) logic for alloc and free, according to my [simple bench](https://github.com/JesseCodeBones/WastDemo/tree/main/02assemblyscript/03malloc_bench), when managing about 100~200 objects at the same time, TLSF will be more efficient than LibMemory solution.

Bench Result:
image
image

Target file size:
image

### DIFF
--- a/std/assembly/rt/index-minimal.ts
+++ b/std/assembly/rt/index-minimal.ts
@@ -1,2 +1,2 @@
-import "rt/tlsf";
+import "rt/lm";
 import "rt/tcms";
--- a/std/assembly/rt/lm.ts
+++ b/std/assembly/rt/lm.ts
@@ -1,0 +1,194 @@
+import { E_ALLOCATION_TOO_LARGE } from "../util/error";
+
+// This allocator is follow the algorithm from https://github.com/embeddedartistry/libmemory
+// FreeList is a linked list that manage all free blocks.
+
+// @ts-ignore: decorator
+@unmanaged class LinkedList {
+    prev: LinkedList;
+    next: LinkedList;
+}
+// @ts-ignore: decorator
+@inline export const BLOCK_MAXSIZE: usize = (1 << 30) - sizeof<LinkedList>();
+// @ts-ignore: decorator
+@inline const AL_BITS: u32 = 4; // 16 bytes to fit up to v128
+// @ts-ignore: decorator
+@inline const AL_SIZE: usize = 1 << <usize>AL_BITS;
+// @ts-ignore: decorator
+@inline const AL_MASK: usize = AL_SIZE - 1;
+// @ts-ignore: decorator
+@unmanaged class Block extends LinkedList {
+    size: usize;
+}
+// @ts-ignore: decorator
+@inline const BLOCK_SIZE: usize = offsetof<Block>();
+// @ts-ignore: decorator
+@inline const MIN_BLOCK_SIZE: usize = offsetof<usize>() + BLOCK_SIZE;
+const freeListPtr: usize = memory.data(sizeof<LinkedList>());
+// @ts-ignore: decorator
+@lazy var freelist: LinkedList;
+// @ts-ignore: decorator
+@inline function alignUp(num: usize): usize {
+    return ((num) + ((1 << alignof<usize>()) - 1)) & ~((1 << alignof<usize>()) - 1);
+}
+// @ts-ignore: decorator
+@inline function insertItem(newItem: LinkedList, preItem: LinkedList, nextItem: LinkedList): void {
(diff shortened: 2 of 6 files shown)


dcodeIO:
One good test case we have is bootstrapping the compiler with a different allocator. Would be interesting to see what's the difference there.

MaxGraey:
> It can use page fault ([Link](https://en.wikipedia.org/wiki/Page_fault)) feature of OS / Runtime.

It doesn't make sense for WebAssembly

> Linearly memory grow

That's that how any memory allocator works. Well, maybe except for "Buddy alloc" scheme

> Smaller size of target files

Only 400 bytes for release

But this allocator isn't battle tested in our environment. It has always O(N) for alloc and free. Linked list -baased allocators usually have very bad fragmentation behavior, while TLSF has the proven lowest fragmentation of all known allocators
pragm-perf

dcodeIO:
Tried bootstrapping with the allocator. Good news is that it appears to work. Bad news is that it is extremely slow:

```
╭───────────────────────────────────╮
│ Stats                             │
╞════════════╤════════════════╤═════╡
│ parse      │ 2284818.057 ms │ 162 │
│ initialize │  175197.964 ms │   1 │
│ compile    │ 6242447.154 ms │   1 │
│ emit       │   29783.141 ms │   4 │
│ validate   │      30.426 ms │   1 │
│ optimize   │       0.301 ms │   1 │
│ transform  │            n/a │   0 │
├────────────┴────────────────┴─────┤
│ Took 8732944.730 ms               │
│ 82 reads, 5 writes                │
╰───────────────────────────────────╯
```

For comparison, typical stats are:

```
╭───────────────────────────────╮
│ Stats                         │
╞════════════╤════════════╤═════╡
│ parse      │ 266.474 ms │ 162 │
│ initialize │  27.732 ms │   1 │
│ compile    │ 970.060 ms │   1 │
│ emit       │ 711.229 ms │   4 │
│ validate   │  29.179 ms │   1 │
│ optimize   │   0.442 ms │   1 │
│ transform  │        n/a │   0 │
├────────────┴────────────┴─────┤
│ Took 2103.953 ms              │
│ 82 reads, 5 writes            │
╰───────────────────────────────╯
```

JesseCodeBones:
@dcodeIO @MaxGraey 
Hi Maintainers,
thank you for the reply.
I am trying to findout another option for memory management, it is not a replacement fot tlsf, tlsf is a good option with o(1) cost.
yes for big project tlsf is much more better, and if we can bring it as another option for user?

For the buddy malloc, I still not investigate it but plan to look into it later.

MaxGraey:
> For the buddy malloc, I still not investigate it but plan to look into it later.

We used buddy allocator before TLSF. The latter came out much more economical and seems to be faster

MaxGraey:
I don't really understand why you are trying to replace TLSF? It is one of the most compact allocators that has excellent features

JesseCodeBones:
> I don't really understand why you are trying to replace TLSF? It is one of the most compact allocators that has excellent features

Actually the main target is our script is quite small and the runtime support page fault to grow the linear memory (not OS level).
this is main reason😁
if it is not the target of community I will close this PR

JesseCodeBones:
@MaxGraey @dcodeIO 
you know, the TLSF has a tail at the end, so after page grow, the runtime will actually allocate the full space (at least 64KB). Thank you for the check, it is a small request, reasonable to decline or merge it:)

MaxGraey:
Did you try `lowMemoryLimit`?
```
--lowMemoryLimit      Enforces very low (<64k) memory constraints.
```

JesseCodeBones:
> Did you try `lowMemoryLimit`?
> 
> ```
> --lowMemoryLimit      Enforces very low (<64k) memory constraints.
> ```

I did not investigate very detail about this option, is it allocate memory from `stack`? I will check it later.
 Our runtime is very special one.
the memory usage can larger than 64KB (for example process serialization), but normally it is tiny script with very limited resources.

PR #2548, opened 2022-11-04
BREAKING CHANGE: Make class fields a special kind of property

dcodeIO:
Removes the hard distinction between fields, i.e.

```ts
class Foo {
  a: i32;
}
```

and properties, i.e.

```ts
class Foo {
  get a(): i32 { return 42; }
  set a(value: i32) { }
}
```

by making fields special kinds of properties, differing in having an associated memory offset. As a result, fields (which get built-in getters and setters) and properties (with explicit getters and setters) can override each other, lifting the annoying limitation that interface fields can only be implemented with properties. Since interchangeable fields and properties are more powerful than what is possible in JS/TS, compatibility with TS becomes a matter of disallowing what wouldn't be supported there due to JS nature.


dcodeIO:
Last commit bakes in static `instanceof` helpers for both classes and interfaces that no longer rely on the RTTI blob. Together with host bindings that also generate static code, bundling the RTTI blob now is no longer necessary. I did not yet remove the RTTI blob, since doing so in turn renders the loader non-functional, so should likely get its own PR.

tests/compiler/rt/instanceof.debug.wat
```diff
@@ -2805,6 +2742,39 @@
...
+   i32.sub
+   i32.load $0
+   i32.const 5
+   i32.eq
+   br_if $is_instance
+   i32.const 0
+   return
+  end
+  i32.const 1
+ )
```
dcodeIO, line 2777: An alternative here is to bake all possible ids into one function instead of calling through to extendees / implementers. Would benefit perf (possibly allowing `br_table`s), but duplicates checks for ids. Might be worth it.

dcodeIO: Decided to make this change

dcodeIO:
(for some reason this one doesn't show as merged as usual but is still open, so closing manually)

PR #2550, opened 2022-11-04
Move module creation to program

dcodeIO:
Moves creation of the module to the program, which is the common link between all of parser, resolver, compiler *and* module. With the change, it for example becomes possible to assign names of program elements when building compound types for classes and signatures, where previously there was a missing reference to the module (type => program !=> module) the names are supposed to be added to.


CountBleck:
This change also allows `CompiledExpression` to be usable in transforms, I think.

PR #2555, opened 2022-11-10
Remove no longer used half of RTTI

dcodeIO:
The compiler no longer depends on embedded RTTI, in particular base class ids (see #2548, where `instanceof` helpers are now compiler-generated). There are still two uses of RTTI outside of the compiler itself, though: One is to determine pointer-free objects during GC, and the other is the loader, that both rely on type flags. The pointer-free use case is important enough to at least keep the flags, so this PR slashes RTTI in half, removing the base class id alongside the now unused programmatic `__instanceof` helper.


jtenner:
This is going to break ASON. Why is `__instanceof` being removed? There doesn't seem like a logical reason for this to occur.

dcodeIO:
It's a change resulting from #2548, where remaining open issues with interfaces have been tackled as well that required a more sophisticated approach to implement `instanceof` to also support interfaces. It's now compiler generated. In turn, the old mechanism to perform these checks became unused, and as such redundant. Can you explain the use case a little? Perhaps there are ways to support it otherwise.

jtenner:
```ts
abstract class BaseClass {}

class MyClass extends BaseClass {}

ASON.deserialize<BaseClass>(buffer); // no longer can validate a serialized MyClass

```


This breaks the entirety of the as-lunatic package and I would have to ask end users of ASON to do this:

```ts
let result = ASON.deserialize<BaseClass>(); // oh no it's an Array
assert(result instanceof BaseClass); // not ergonomic and backwards uncompatible
```

dcodeIO:
Can `ASON.deserialize<T>` perhaps do:

```ts
function deserialize<T>(...) {
  let ret = ...;
  if (ret instanceof T) ... // AS allows this
}
```

jtenner:
The assert would panic if it isn't an instanceof  T... let me try that.

dcodeIO:
I guess one question there is what is used as the type of `ret` so this doesn't become precomputed, hmm.

PR #2558, opened 2022-11-11
Simplify compileCallExpression

### LINKED ISSUE #2525: Error when call function from array of functions

dcodeIO:
An alternative to #2530 that instead simplifies the logic in `compileCallExpression`: Any call that is not a direct call compiles the target expression for use as the function argument of an indirect call.

cc @CountBleck 

Fixes #2530, fixes #2525 


CountBleck:
If it works, it looks good to me. You should link #2525 to this PR.

PR #2559, opened 2022-11-12
Implicitly inherit from Object

dcodeIO:
Classes that are not unmanaged and don't extend other classes now implicitly extend `Object`. As a result, the type `Object` can be used to reference and pass along any normal class instance.

```ts
let obj: Object = new Anything();
```

```ts
function fn(arg: Object): void {
  if (arg instanceof Anything) {
    ...
  }
}
fn(new Anything());
```

Note: Does not apply to basic value types like `i32`, which are not objects.

With `Object` becoming the general base class, a breaking change to constant class ids was necessary, in that foundational classes move one up. Relevant for external integration with the runtime, where these ids may be hardcoded.

| Class | New id | Old id |
|-------|---------|--------|
| Object | 0 | dynamic
| ArrayBuffer | 1 | 0
| String | 2 | 1
| ... | |


jtenner:
Thank you for this! I was able to test the branch with as-pect and it works after I made some coding changes. I also needed to use the new loader because the runtime type ids changed. Other than that, everything looks good to me!

PR #2563, opened 2022-11-14
This needs to be fixed

jtenner:
Hey it looks like we missed something here in the loader.


jtenner:
Oh never mind. I was using @assemblyscript/loader instead of the lib loader. This is okay.

PR #2582, opened 2022-11-30
Improve Signature lookups, Make Signature immutable

jtenner:
In this pull request, the Signature class becomes immutable. The reason being is because signature type lookups have become a big bottleneck for libraries that have a lot of type signatures. We can uniquely identify these signatures by calculating their string representation using signature.toString(). This causes a lot of not so very straightforward problems within the compiler, the program, and the resolver.  For instance, the compiler now has to call the `newStub` function with the maximum arguments, and when resolving a type in the resolver, we are not allowed to simply modify the return type of signatures, and instead we create a new one to obtain the proper function type id at instantiation time.

Changes involved:
- compiler tests fixtures updated
- Signatures are now immutable
- Some changes involving signatureReference equality have changed

⯈
⯈
⯈


jtenner:
After some local testing, I got some really amazing speed increases. I tested nearly 250k function signatures in that list, and the time it took to deal with them reduced to thirty seconds instead of a whopping 25 minutes (or more.) This can be regarded as a generalized improvement.

src/types.ts
```diff
@@ -421,13 +421,20 @@ export class Type {
...
-        this.isNullableReference == other.isNullableReference
-      );
+      if (this.classReference != other.classReference) return false;
+      let selfSignatureReference = this.signatureReference;
+      let otherSignatureReference = other.signatureReference;
+      if (selfSignatureReference && otherSignatureReference) {
+        if (!selfSignatureReference.equals(otherSignatureReference)) return false;
+      } else {
+        if (selfSignatureReference != otherSignatureReference) return false;
+      }
```
dcodeIO, line 435: Shouldn't equality be guaranteed by pointer equality here (as checked before), since if a signature would be equal to another, the pre-existing other would be returned from the constructor already so there cannot be two equal signatures with different pointers?

src/types.ts
```diff
@@ -904,14 +907,16 @@ export class Signature {
     program: Program,
     parameterTypes: Type[] | null = null,
     returnType: Type | null = null,
-    thisType: Type | null = null
+    thisType: Type | null = null,
+    /** Number of required parameters excluding `this`. Other parameters are considered optional. */
+    public readonly requiredParameters: i32 = parameterTypes ? parameterTypes.length : 0,
+    /** Whether the last parameter is a rest parameter. */
+    public readonly hasRest: bool = false,
```
dcodeIO, line 914: Suggesting to unify these to also use explicit fields to fit in with the other parameters. Alternatively, the old ones could also be changed to become constructor parameters like the new ones, if that's possible. Either way is fine for me as long as code becomes uniform.

src/types.ts
```diff
@@ -922,16 +927,20 @@ export class Signature {
...
     let signatureTypes = program.uniqueSignatures;
-    let length = signatureTypes.length;
-    for (let i = 0; i < length; i++) {
-      let compare = unchecked(signatureTypes[i]);
-      if (this.equals(compare)) {
-        this.id = compare.id;
-        return this;
+    let compareString = this.toString();
+    if (signatureTypes.has(compareString)) {
+      let existing = signatureTypes.get(compareString)!;
```
dcodeIO, line 932: Nit: When compiling with `asc`, this line yields a "Value is never null" kind of warning. Is somewhat a design wart resulting from JS being typed with `| undefined` here. A way around for now is
```suggestion
      let existing = changetype<Signature>(signatureTypes.get(compareString));
```

src/types.ts
```diff
@@ -922,16 +927,20 @@ export class Signature {
...
+      if (!this.equals(existing)) {
+        trace(compareString);
+        trace(existing.toString());
```
dcodeIO, line 935: The debug traces here and below should be removed before a merge :)

src/types.ts
```diff
@@ -922,16 +927,20 @@ export class Signature {
     type.signatureReference = this;
 
     let signatureTypes = program.uniqueSignatures;
-    let length = signatureTypes.length;
-    for (let i = 0; i < length; i++) {
-      let compare = unchecked(signatureTypes[i]);
-      if (this.equals(compare)) {
-        this.id = compare.id;
-        return this;
+    let compareString = this.toString();
```
dcodeIO, line 930: Perhaps
```suggestion
    let uniqueKey = this.toString();
```

src/types.ts
```diff
@@ -1105,7 +1127,8 @@ export class Signature {
     let thisType = this.thisType;
     if (thisType) {
       sb.push(validWat ? "this:" : "this: ");
-      assert(!thisType.signatureReference);
+      // sometimes "this" can be a function type
+      // assert(!thisType.signatureReference);
```
dcodeIO, line 1131:
```suggestion
```

src/types.ts
```diff
@@ -1127,7 +1150,7 @@ export class Signature {
   }
 
   /** Creates a clone of this signature that is safe to modify. */
-  clone(): Signature {
+  clone(requieredParameters: i32 = this.requiredParameters, hasRest: bool = this.hasRest): Signature {
```
dcodeIO, line 1153: Typo: "required"

dcodeIO (commented):
Overall looks good to me, modulo comments :)

src/types.ts
jtenner: For some reason I kept hitting a runtime assertion here. Perhaps I should revisit this?

dcodeIO: Hitting an assertion likely indicates that signatures are not properly deduplicated, that there is another late modification of a signature that should not exist, that `toString` is missing info, or similar. Finding out "why" would help to properly understand the nature of the problem.

src/types.ts
jtenner: This will require a refactor rename.

src/types.ts
jtenner: These properties are now constructor parameters.

src/types.ts
jtenner: I determined this to be because I was returning the wrong signature in the constructor. This was later refactored into a `static new(): Signature` function.

src/compiler.ts
```diff
@@ -474,7 +474,7 @@ export class Compiler extends DiagnosticEmitter {
     module.setFeatures(featureFlags);
 
     // set up the main start function
-    let startFunctionInstance = program.makeNativeFunction(BuiltinNames.start, new Signature(program, [], Type.void));
+    let startFunctionInstance = program.makeNativeFunction(BuiltinNames.start, Signature.new(program, [], Type.void));
```
CountBleck, line 477:
```suggestion
    let startFunctionInstance = program.makeNativeFunction(BuiltinNames.start, Signature.create(program, [], Type.void));
```

src/compiler.ts
```diff
@@ -7102,8 +7102,7 @@ export class Compiler extends DiagnosticEmitter {
         }
       }
 
-      let signature = new Signature(this.program, parameterTypes, returnType, thisType);
-      signature.requiredParameters = numParameters; // !
+      let signature = Signature.new(this.program, parameterTypes, returnType, thisType, numParameters);
```
CountBleck, line 7105:
```suggestion
      let signature = Signature.create(this.program, parameterTypes, returnType, thisType, numParameters);
```

src/compiler.ts
```diff
@@ -8669,7 +8668,7 @@ export class Compiler extends DiagnosticEmitter {
             )
           ),
           null,
-          new Signature(this.program, null, classInstance.type, classInstance.type),
+          Signature.new(this.program, [], classInstance.type, classInstance.type),
```
CountBleck, line 8671:
```suggestion
          Signature.create(this.program, [], classInstance.type, classInstance.type),
```

src/program.ts
```diff
@@ -3217,7 +3217,7 @@ export class File extends Element {
     program.filesByName.set(this.internalName, this);
     let startFunction = this.program.makeNativeFunction(
       `start:${this.internalName}`,
-      new Signature(program, null, Type.void),
+      Signature.new(program, [], Type.void),
```
CountBleck, line 3220:
```suggestion
      Signature.create(program, [], Type.void),
```

src/resolver.ts
```diff
@@ -422,9 +422,7 @@ export class Resolver extends DiagnosticEmitter {
       );
       if (!returnType) return null;
     }
-    let signature = new Signature(this.program, parameterTypes, returnType, thisType);
-    signature.requiredParameters = requiredParameters;
-    signature.hasRest = hasRest;
+    let signature = Signature.new(this.program, parameterTypes, returnType, thisType, requiredParameters, hasRest);
```
CountBleck, line 425:
```suggestion
    let signature = Signature.create(this.program, parameterTypes, returnType, thisType, requiredParameters, hasRest);
```

src/resolver.ts
```diff
@@ -2706,7 +2704,14 @@ export class Resolver extends DiagnosticEmitter {
       }
       const type = this.resolveExpression(expr, tempFlow, ctxType, reportMode);
       if (type) {
-        signatureReference.returnType = type;
+        functionType.signatureReference = Signature.new(
```
CountBleck, line 2707:
```suggestion
        functionType.signatureReference = Signature.create(
```

src/resolver.ts
```diff
@@ -2853,8 +2858,7 @@ export class Resolver extends DiagnosticEmitter {
       returnType = type;
     }
 
-    let signature = new Signature(this.program, parameterTypes, returnType, thisType);
-    signature.requiredParameters = requiredParameters;
+    let signature = Signature.new(this.program, parameterTypes, returnType, thisType, requiredParameters);
```
CountBleck, line 2861:
```suggestion
    let signature = Signature.create(this.program, parameterTypes, returnType, thisType, requiredParameters);
```

src/types.ts
```diff
@@ -882,58 +887,72 @@ export function typesToString(types: Type[]): string {
...
-  hasRest: bool;
-  /** Respective function type. */
-  type: Type;
-  /** The program that created this signature. */
-  program: Program;
+  
 
-  /** Constructs a new signature. */
-  constructor(
+  public static new(
```
CountBleck, line 892:
```suggestion
  public static create(
```

src/types.ts
```diff
@@ -1127,18 +1145,20 @@ export class Signature {
...
-  clone(): Signature {
+  clone(requiredParameters: i32 = this.requiredParameters, hasRest: bool = this.hasRest): Signature {
     let parameterTypes = this.parameterTypes;
     let numParameterTypes = parameterTypes.length;
     let cloneParameterTypes = new Array<Type>(numParameterTypes);
     for (let i = 0; i < numParameterTypes; ++i) {
       unchecked(cloneParameterTypes[i] = parameterTypes[i]);
     }
-    return new Signature(
+    return Signature.new(
```
CountBleck, line 1155: Is this actually going to create a clone?
```suggestion
    return Signature.create(
```

CountBleck (changes requested):
I didn't think of this, but `.new` isn't the best name for this. `.create` would be better.
Also, there is a stray `output.txt` file.

PR #2587, opened 2022-12-06
Use constructor signature for class decorators

bullno1:
Use constructor signature for `@unmanaged` and `@final` instead of `Function`.
This removes the need to use `// @ts-ignore: decorator` to suppress error from tsserver (used for code completion) while using the `assembly.json` config.

`Function` in `assembly` is different from `portable` so a class type signature inferred by Typescript is not assignable to `Function`.


dcodeIO:
Thanks!

Issue #2577, opened 2022-11-23
instanceof check for interfaces is static, not dynamic


CountBleck:
Compare and contrast the following two snippets of code: [`<Object> instanceof <class>`](https://www.assemblyscript.org/editor#IyFvcHRpbWl6ZT1ub25lJnJ1bnRpbWU9aW5jcmVtZW50YWwKY2xhc3MgRm9vIHt9DQpjbGFzcyBCYXIgZXh0ZW5kcyBGb28ge30NCg0KY29uc3QgdGVzdCA9IGNoYW5nZXR5cGU8QmFyPigwKQ0KDQpleHBvcnQgZnVuY3Rpb24gaGV5KCk6IGJvb2wgew0KICByZXR1cm4gdGVzdCBpbnN0YW5jZW9mIEZvbw0KfQ0KDQpleHBvcnQgZnVuY3Rpb24gaGl5YSgpOiBib29sIHsNCiAgcmV0dXJuIHRlc3QgYXMgT2JqZWN0IGluc3RhbmNlb2YgRm9vDQp9CiMhaHRtbAo=) and [`<Object> instanceof <interface>`](https://www.assemblyscript.org/editor#IyFvcHRpbWl6ZT1ub25lJnJ1bnRpbWU9aW5jcmVtZW50YWwKY2xhc3MgRm9vIHt9DQpjbGFzcyBCYXIgZXh0ZW5kcyBGb28ge30NCg0KY29uc3QgdGVzdCA9IGNoYW5nZXR5cGU8QmFyPigwKQ0KDQpleHBvcnQgZnVuY3Rpb24gaGV5KCk6IGJvb2wgew0KICByZXR1cm4gdGVzdCBpbnN0YW5jZW9mIEZvbw0KfQ0KDQpleHBvcnQgZnVuY3Rpb24gaGl5YSgpOiBib29sIHsNCiAgcmV0dXJuIHRlc3QgYXMgT2JqZWN0IGluc3RhbmNlb2YgRm9vDQp9CiMhaHRtbAo=)

The former snippet works as expected: the `hiya` export calls an `instanceof` helper, which then checks the ID at runtime.
The latter snippet does not work as expected: the `hiya` export returns `false`, because the `instanceof` is incorrectly checked at compile-time.

[...]

PR #2588, opened 2022-12-08
Do precise subtype tests in instanceof

dcodeIO:
Addresses #2577. One rather trivial part of the issue was that previously, since interfaces are not derived from `Object` but their implementing classes are, an `someInterface as Object instanceof SomeInterface` would fail. Another was that the check for when a cast may possibly succeed was overly simplistic, thus required more elaborate infrastructure.

The seemingly obvious alternative that doesn't work is to simply fall back to do dynamic checks if the relationship between the types is arbitrarily complicated (and hence possibly costly to compute), since that breaks assumptions about such checks being precomputable.

Making a draft for now while I untie the knot in my brain before figuring out how to best test the changes. Is perfectly possible that I missed something, hence I'd appreciate any amount of eyes on the new logic :)

### DIFF
--- a/src/program.ts
+++ b/src/program.ts
@@ -4459,7 +4459,7 @@ export class Class extends TypedElement {
     } while (nextBase);
   }
 
-  /** Propagates an extendee to this class as its base classes. */
+  /** Propagates an extendee to this class and its base classes. */
   private propagateExtendeeUp(extendee: Class): void {
     let nextBase: Class | null = this;
     do {
--- a/src/program.ts
+++ b/src/program.ts
@@ -4466,7 +4466,8 @@ export class Class extends TypedElement {
               let thisExtendee = extendeesArray[i];
               baseInterfaceImplementers.add(thisExtendee);
             }
-            baseInterface = baseInterface.base;
+            baseInterface = <Interface>baseInterface.base;
+            assert(!baseInterface || baseInterface.isInterface);
           } while (baseInterface);
         }
       }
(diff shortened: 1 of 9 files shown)


src/program.ts
```diff
@@ -4743,6 +4768,80 @@ export class Class extends TypedElement {
     }
     return out;
   }
+
+  /** Tests if this class or interface extends the given classs or interface. */
```
CountBleck, line 4772: Typo:
```suggestion
  /** Tests if this class or interface extends the given class or interface. */
```

src/program.ts
```diff
@@ -4743,6 +4768,80 @@ export class Class extends TypedElement {
...
+    let base = this.base;
+    while (base) {
+      if (base == other) return true;
+      base = base.base;
+    }
+    return false;
+  }
+
+  /** Tests if this class has a direct or indirect extendee matching the given class. */
+  hasExtendee(other: Class): bool {
```
CountBleck, line 4783: Wouldn't this method be the same thing as `other.extends(this)`?

src/program.ts
```diff
@@ -4743,6 +4768,80 @@ export class Class extends TypedElement {
     }
     return out;
   }
+
+  /** Tests if this class or interface extends the given classs or interface. */
+  extends(other: Class): bool {
+    let base = this.base;
```
CountBleck, line 4774: Shouldn't the method check if `this == other` too? Like so:
```suggestion
    let base = this;
```

CountBleck (commented):
I don't really get this stuff, so I can't help much :)

dcodeIO:
So I wasn't quite happy with the previous take and decided to refactor this more. Now, `extendees` and `implementers` are populated early using a bunch of ugly loops that are probably less ugly than frequent recursion, so the checks, whereof there may be many, are just hash map lookups.

src/program.ts
dcodeIO: That would mean that a class extends itself. Probably better to check that manually with an `a == b` where applicable.

Issue #1299, opened 2020-05-27
Support WeakRef / FinalizationRegistry


dcodeIO:
An [earlier issue](https://github.com/AssemblyScript/assemblyscript/issues/383) suggested the implementation of a `@destructor` annotation, but it turned out that is is likely not feasible and we should instead aim for a more JS-like approach, i.e. along the lines of [WeakRef and FinalizationRegistry](https://github.com/tc39/proposal-weakrefs) since that's much safer. Also, since `__finalize` has been implemented recently, it should be possible now to roll custom solutions, or to investigate a JS-like approach further.

WenheLI:
Hi interested in this one! Any starting material for this issue?

jtenner:
I have a very specific use case for this in lunatic where a finalization registry might not be feasible either. I could ask the creator to implement rtrace host functions... But that might be a bad idea in general.

Any ideas?

dcodeIO:
Can you describe the concrete use case you have? That would help me to give you a better answer :)

jtenner:
When using lunatic, a socket can can be obtained like this:

```ts
let server = TCPServer.bind(ip, port);
let socket: TCPSocket;

while (socket = server.accept()) {
  // socket is now referenced once
  Process.spawn(socket, (theSocket: TCPSocket) => {
    // socket is now referenced twice, only now it's on another process
    socket.close();
    // now only once
  });
  // if we don't call socket.close() here, it will forever remain open, especially if it falls out of scope
}
```

As you can see, in languages like rust, we can automatically call `socket.close()` to effectively dereference it when it gets cleaned up by the memory manager.

Currently, it requires the user to make sure that they call `socket.close()` on every single process that touches a socket, which is not very ergonomic.

jtenner:
I also realize that the garbage collector could never run, and the reference to the socket could still never be freed. It seems like there's no good solution here.

dcodeIO:
Yeah, hooking into the GC is most likely not a sufficient device for closing system resources, and may never be. One mechanism that comes to mind here, that's also present is JS, is to "transfer" a resource, so the current one becomes "detached". Perhaps something along these lines could help?

jtenner:
In the majority of cases, we actually want to reference a socket on multiple processes. In fact, without that feature, it becomes impossible to read and write to a socket asynchronously.

dcodeIO:
Would some sort of reference count on the socket work, so it is only actually closed once all threads have closed it?

[...]

PR #2593, opened 2022-12-20
Implement FinalizationRegistry and WeakRef

bullno1:
Resolve https://github.com/AssemblyScript/assemblyscript/issues/1299

All tests fail right now because fixtures are not recreated.
This is created to gather feedbacks.

One debug output is attached to see the impact of this change.
AFAIK, there is no impact on release builds when the feature is not used.

Current quirks of `FinalizationRegistry`:

* ~`register` only takes 2 arguments and the 3rd argument is assumed to be the object.~
  `register` always take 3 arguments because undefined is not representable.
* ~Each object can only be registered once per registry.~
   ~Simplify the implementation for what I believe is the most typical use case (see below).~
   Each `token` can only be registered once.
* Repeated `register` and `unregister` will consume extra memory until the target object is collected.
   This is because new book keeping entries are created in `register` but not removed in `unregister`.
   The reason is again, to keep the implementation simple and a typical use case should only `unregister` once, typically in an explicit "close" method (see below).

   Even in the case of duplicated entries, `finalizer` is still guaranteed to be called once per token.
  
```typescript
declare function closeFd(fd: u32); void; // some native function

class Socket {
    private static FINALIZATION_REGISTRY = new FinalizationRegistry<u32>(closeFd);

    constructor(private fd: u32) {
        FINALIZATION_REGISTRY.register(this, fd, this);
    }

    close() { // explicit clean up
        if (FINALIZATION_REGISTRY.unregister(this)) {  // can call close multiple times safely
            closeFd(fd);
        }
    }
}
```


dcodeIO:
Regarding making FRs GC-able, would something like this work?

```ts
let ptrToRegistry = new Map<usize,FinalizationRegistry[]>();

function finalizeImpl(ptr: usize): void {
  if (!ptrToRegistry.has(ptr)) return;
  let registries = ptrToRegistry.get(ptr);
  for (let i = 0, k = registries.length; i < k; ++i) {
    registries[i].finalize(ptr);
  }
  ptrToRegistry.delete(ptr);
}
```

so when `finalizeImpl` is invoked with a pointer, it can forward to the FRs handling the pointer, and afterwards delete the pointer so the respective FRs can be collected? Any registry's `register` would then register with `ptrToRegistry`.

bullno1:
I think that's better.

Implementing a WeakMap is actually not easy without per-instance GC-able FR.

std/assembly/finalizationregistry.ts
```diff
@@ -0,0 +1,76 @@
...
+
+    if (!INITIALIZED) {
+      PREVIOUS_FINALIZER = __finalize;
+      __finalize = BaseRegistry.finalizeAll.index;
+      INITIALIZED = true;
+    }
+  }
+
+  register<U>(key: U, value: T): void {
+    assert(isManaged<U>());
```
dcodeIO, line 46: Could this be a compile-time error so using an invalid key type won't compile?
```suggestion
    if (!isManaged<U>()) ERROR("invalid target");
```
The corresponding error in Node seems to be "Uncaught TypeError: FinalizationRegistry.prototype.register: invalid target".

bullno1:
I guess my problem with this is that GC pressure is kind of amplified.
`register` creates an array.
`__finalize` drops the array for the next collection cycle.

It might be worth pooling those arrays.

std/assembly/finalizationregistry.ts
bullno1: I see, I didn't know about `ERROR`

jtenner:
Fwiw finalization was implemented using as-disposable already and it had to deal with quite a lot of problems

bullno1:
@dcodeIO There are still a lot of tests to write but this change will be huge due to the code change to GCs to support `__finalize` as an index.
Thankfully, it should only affect debug fixtures.
Should I make a separate PR for that?

I'd argue it's a better GC hook API than `__finalize` as a symbol.
It allows the stdlib hook to be improved without affecting fixtures.
Moreover, multiple hooks can be set at the same time by chaining to the previous hook.

However, it will break as-disposable or anything that relies on `__finalize` being an optional symbol.
It is possible to keep compatibility by putting something like this in GC initialization:

```
// now call it __finalize_index instead
export let __finalize_index: u32 = 0;

// When initializing GC
if (isDefined(__finalize)) {
    __finalize_index = __finalize.index; 
}
```

After the `__finalize` change, I guess I will only submit `FinalizationRegistry` first.
`WeakRef` and `WeakMap` have many little details.

jtenner:
I'm glad this is at least being talked about. Can we not break something that lunatic devs rely on please? It would be nice to rely on __finalize as a feature.

PR #2595, opened 2022-12-22
Integrate Rtrace with ESM bindings

dcodeIO:
The Rtrace utility to inspect the inner workings of the runtime must be installed on the respective imports objects, which did not work as-is with ESM bindings

Hence, add a special case to ESM bindings generation to install Rtrace when used, exclude its imports from instrumentation, and import it with a private package specifier so it can be mapped via `imports` in package.json. Also unbreaks `bootstrap:rtraced` in so far as that it executes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,6 +59,12 @@
     },
     "./*": "./*"
   },
+  "imports": {
+    "#rtrace": {
+      "import": "./lib/rtrace/index.js",
+      "types": "./lib/rtrace/index.d.ts"
+    }
+  },
   "bin": {
     "asc": "./bin/asc.js",
     "asinit": "./bin/asinit.js"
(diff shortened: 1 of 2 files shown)


CountBleck:
I spent way too long yesterday (at least two/three hours) trying to hack around ESM/raw bindings to get rtrace to work. I'm glad it's working now.

Should rtrace be added to CI in case it catches errors before faulty PRs are merged? Or is this bug a special case?

dcodeIO:
Rtrace is used for some of the tests (when `asc_trace` is specified), but using it on the compiler itself it is slooow / uses too much memory. Perhaps if we could speed it up somehow, but in its current form it'll likely blow every budged imposed by GH actions.

PR #2596, opened 2022-12-22
Fix lifetime confusion in shadow stack pass

dcodeIO:
Potential fix for the errors observed in #2594, and prior in #2575.

```ts
let outer = null;
while (condition) {
  let inner = value;
  outer = inner; // here
}
```

Previously, at `// here`, no shadow stack slot was populated, assuming that `inner` respectively `value` is already kept alive by another slot. In a loop, this assumption doesn't hold obviously, since if `inner` is reassigned a different value in a subsequent iteration, the previous value copied to `outer` is, due to skipping the stack slot, no longer kept alive.


CountBleck (approved):
Looks good to me. It would be nice if slots could still be omitted by checking lifetimes somehow.

PR #2597, opened 2022-12-23
Little problem in the comments

Dan-k391:
⯈
⯈
⯈


CountBleck:
No, that's correct. It refers to `i31ref`, which a type introduced in the GC proposal if I remember correctly.

CountBleck:
See [here](https://github.com/WebAssembly/gc/blob/main/proposals/gc/MVP.md#reference-types).

Dan-k391:
ok

PR #2607, opened 2023-01-06
Don't implicitly call exportStart function in generated host bindings

logandavies181:
Hi there. Great project - imo it's what TS should have been :)

This PR removes the call to the target of --exportStart in the `instantiate` function from the generated JS host bindings. The current behavior causes issues for non-browser JS hosts as per the below.

I'm not sure how this affects browser-based users of --exportStart (nor what their use-case is for it) - but regardless, the current behaviour of the generated bindings does seem to be at odds with [the documentation](https://www.assemblyscript.org/compiler.html#compiler-options), which says that the function is not called implicitly, except that `instantiate` calls it anyway.

Consider this example from [the Nodejs wasi docs](https://nodejs.org/api/wasi.html), which I've modified to use generated host bindings:

```typescript
import { instantiate } from './build/release.js';

import { readFile } from 'node:fs/promises';
import { WASI } from 'wasi';
import { argv, env } from 'node:process';

const wasi = new WASI({
    args: argv,
    env,
});

const importObject = { wasi_snapshot_preview1: wasi.wasiImport };

const wasm = await WebAssembly.compile(await readFile(new URL('./build/release.wasm', import.meta.url)));

const instance = await instantiate(wasm, importObject); // error raised here

wasi.start({ exports: instance });
```

Nodejs exits with this error:
```
file:///some/path/build/release.js:11
        return __module0.fd_write(fd, iovs, iovs_len, nwritten);
                         ^

Error: wasi.start() has not been called
    at fd_write (file:///some/path/build/release.js:11:26)
    at wasm://wasm/bbc26f1e:wasm-function[12]:0xfcf
    at wasm://wasm/bbc26f1e:wasm-function[13]:0x1081
    at wasm://wasm/bbc26f1e:wasm-function[9]:0xa03
    at instantiate (file:///some/path/build/release.js:21:11)
    at async file:///some/path/index.js:16:18 {
  code: 'ERR_WASI_NOT_STARTED'
}

Node.js v19.2.0
```

Deno has a [nearly identical API](https://deno.land/std@0.170.0/wasi/snapshot_preview1.ts) and quietly exits instead of erroring like Node does

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -55,6 +55,7 @@ under the licensing terms detailed in LICENSE:
 * CountBleck <Mr.YouKnowWhoIAm@protonmail.com>
 * Abdul Rauf <abdulraufmujahid@gmail.com>
 * Bach Le <bach@bullno1.com>
+* Logan Davies <logan.davies181@gmail.com>
 
 Portions of this software are derived from third-party works licensed under
 the following terms:
--- a/src/bindings/js.ts
+++ b/src/bindings/js.ts
@@ -931,11 +931,6 @@ export class JSBuilder extends ExportsWalker {
     if (this.needsGetF32) sb.push(makeCheckedGetter("F32", "getFloat32"));
     if (this.needsGetF64) sb.push(makeCheckedGetter("F64", "getFloat64"));
 
-    let exportStart = options.exportStart;
-    if (exportStart) {
-      sb.push(`  exports.${exportStart}();\n`);
-    }
-
     if (hasAdaptedExports) {
       sb.push("  return adaptedExports;\n}\n");
     } else {
--- a/tests/compiler/bindings/noExportRuntime.debug.js
+++ b/tests/compiler/bindings/noExportRuntime.debug.js
@@ -141,7 +141,6 @@ async function instantiate(module, imports = {}) {
       return __dataview.getUint32(pointer, true);
     }
   }
-  exports._start();
   return adaptedExports;
 }
 export const {


dcodeIO:
There are two mutually exclusive use cases here it seems:

1. Use an AS module from JavaScript. Here, one would generate and use bindings, where it is [sensible](https://www.assemblyscript.org/concepts.html#accessing-memory-during-instantiation) to call the start function so the module is ready to go when `import`ed. Such a module integrates directly with JS via ESM. No WASI here.
2. Use an AS module in a Non-Web host. This use case involves a different set of assumptions since standard Web APIs are not available. In the case of a WASI host, one would typically use [wasi-shim](https://github.com/AssemblyScript/wasi-shim), which essentially replaces calls to Web APIs with calls to WASI imports. No bindings here.

Generally, there should not be a need to use the WASI modules in Node and Deno to *load* AS modules. Ideally, AS would instead one day expose these loader modules so Non-Web modules can be instantiated from within AS.

logandavies181:
> 1. Use an AS module from JavaScript. Here, one would generate and use bindings, where it is [sensible](https://www.assemblyscript.org/concepts.html#accessing-memory-during-instantiation) to call the start function so the module is ready to go when `import`ed. Such a module integrates directly with JS via ESM. No WASI here.

Perhaps it's my lack of understanding, but would that mean to avoid having a web-AS module's top-level statements from being run, you'd have to use the --exportStart option and then not use the generated bindings?

> 2. Use an AS module in a Non-Web host. This use case involves a different set of assumptions since standard Web APIs are not available. In the case of a WASI host, one would typically use [wasi-shim](https://github.com/AssemblyScript/wasi-shim), which essentially replaces calls to Web APIs with calls to WASI imports. No bindings here.

Basically I'm experimenting with using wasi and supplementing it with other host functions as wasi-shim doesn't support things like fetch (wasi_snapshot_preview1 doesn't really provide enough to support it anyway). I was previously looking at other wasm runtimes, (wasmer, wasmtime, wasmedge) which are all wasi-first so I guess that's how I've ended up in this rabbit hole. Seeing as AS generates imports for console.log and such I see now that the web compile target is the better option for AS on node and deno.

Thanks for clearing that up!

> Generally, there should not be a need to use the WASI modules in Node and Deno to _load_ AS modules. Ideally, AS would instead one day expose these loader modules so Non-Web modules can be instantiated from within AS.

Now that's an interesting topic. What would this look like to you, given the AS project's perspective on the Component Model?

dcodeIO:
> Perhaps it's my lack of understanding, but would that mean to avoid having a web-AS module's top-level statements from being run, you'd have to use the --exportStart option and then not use the generated bindings?

The intended behavior here is to do what one would expect from a normal JS module. When it is `import`ed, its top-level statements run (with `--exportStart` that implies invoking the start function), so the result of the import is an initialized module. The general idea is that JS and AS modules should one day be interchangeable. In the Web use case, whether or not `--exportStart` is required during compilation depends on what the top-level statements do. If these call out to JS, and JS synchronously calls back into Wasm, that's sometimes only possible with `--exportStart`, so exported memory is already available when the indirection happens. But perhaps I'm missing the forest for the trees here. Could you elaborate a bit on the issue you have encountered?

> Now that's an interesting topic. What would this look like to you, given the AS project's perspective on the Component Model?

The idea is that one can import any built-in Node or Deno module normally in AS eventually. Similar to `import { ... } from "fs"` to use the built-in FS module, one could also `import { ... } from "wasi"` to use the built-in WASI module - not to instantiate the AS module itself, but any other WASI module. In essence, AS simply provides what's in the environment, not itself needing WASI, or any other redundant (for AS) ABI. More generally, the direction Wasm is taking with WASI and the Component Model seems like a blunder, as these needlessly introduce fragmentation and redundancy, so there's not really a perspective at this point in time. Hopes are that the Wasm spec eventually realizes its mistake.

logandavies181:
> Could you elaborate a bit on the issue you have encountered?

I think I just didn't understand what the flag was doing for the web use case as the docs only directly mention the WASI use-case. Now I see that using the bindings to lift strings (which includes console.log) involves exported memory so exportStart is needed. Maybe if the docs mentioned that it would be more straightforward for newcomers?

> import { ... } from "wasi" to use the built-in WASI module

That sounds like something that could work today by generating bindings and imports for the import object. I might look into that some more.

Thanks for your insight on this, it's been very informative!

Issue #2605, opened 2023-01-04
Several parse / resolve issues


dcodeIO:
These have been reported by a throwaway account lately, with some trolling attached. Collecting the useful bits here.

Parser issues:

```ts
a ? (b) : c
// ERROR TS1005: '=>' expected.
```

```ts
let y: ((x: i32) => i32);
// ERROR TS1005: '|' expected.
```

```ts
{ x: 1 };
// TS1109: Expression expected.
```

Assertion hit:

```ts
let a: void;
a.b;
// crash
```

Endless loop in override discovery:

```ts
class A extends B {
    c(): i32 {}
}

class B extends A {
    c(): i32 {}
}
```

PR #2620, opened 2023-01-17
fix: analysis `ParenthesizedExpression` or `FunctionExpression` correctly in parser (#2605)

HerrCai0907:
fix parser error for `a ? (b) : c` in #2605


src/parser.ts
```diff
@@ -3780,14 +3786,20 @@ export class Parser extends DiagnosticEmitter {
...
+                      break;
+                    }
                   }
-                  // fall-through
+                  if (tn.skip(Token.Equals_GreaterThan)) {
+                    tn.reset(state);
+                    return this.parseFunctionExpression(tn);
+                  }
+                  again = false;
+                  break;
```
dcodeIO, line 3802: Seems that when inverting the condition, the fall-through to parsing a function expression can be maintained instead of duplicating it?
```suggestion
                  if (!tn.skip(Token.Equals_GreaterThan)) {
                    again = false;
                    break;
                  }
                  // fall-through
```

dcodeIO (approved):

dcodeIO:
Thanks :)

PR #2621, opened 2023-01-19
typo

HerrCai0907:
fix some type in `compiler.ts` and `parser.ts`

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -4530,7 +4530,7 @@ export const enum Precedence {
   Grouping
 }
 
-/** Determines the precende of a non-starting token. */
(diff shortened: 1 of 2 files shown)


src/parser.ts
```diff
@@ -4530,7 +4530,7 @@ export const enum Precedence {
   Grouping
 }
 
-/** Determines the precende of a non-starting token. */
+/** Determines the precede of a non-starting token. */
```
dcodeIO, line 4533:
```suggestion
/** Determines the precedence of a non-starting token. */
```

dcodeIO (approved):

PR #2624, opened 2023-01-19
BREAKING CHANGE: fix: detect the premature usage of `@global` variant 

### LINKED ISSUE #2622: `@global` causes compile error
```ts
// index.ts
import {} from "./a"
import {} from "./b"

// a.ts
let a = t;

// b.ts
class T {}
@global export const t = new T();
```

it causes compile error
```
ERROR TS2304: Cannot find name 'T'.
   :
 2 │ @global export const t = new T();
   │                              ~
   └─ in assembly/b.ts(2,30)

FAILURE 1 compile error(s)
```

HerrCai0907:
fix #2622


HerrCai0907:
The bug will cause compile result is incorrect.

```ts
// index.ts
import {} from "./a";
import {} from "./b";

// a.ts
class T {
  a: i32;
}
let a = t;

// b.ts
class T {}
@global export const t = new T();
const b = new T();
```

the compile result is 
```wat
 (func $start:assembly/a
  global.get $~lib/memory/__heap_base
  i32.const 4
  i32.add
  i32.const 15
  i32.add
  i32.const 15
  i32.const -1
  i32.xor
  i32.and
  i32.const 4
  i32.sub
  global.set $~lib/rt/stub/startOffset
  global.get $~lib/rt/stub/startOffset
  global.set $~lib/rt/stub/offset
  i32.const 0
  call $assembly/a/T#constructor
  global.set $assembly/b/t
  global.get $assembly/b/t
  global.set $assembly/a/a
 )
```

It use assign `a/a` and `b/t` with `call $assembly/a/T#constructor` but actually it should be `call $assembly/b/T#constructor`

HerrCai0907:
> Hence I wonder if it would be preferable to detect this situation and raise an error that `t` is used too early, say before its declaration? The fix would then be to import `b.ts` before `a.ts`, instead of permitting potentially undesirable side-effects?

I am wondering how to handle `@lazy` `@global`. I have no idea to detect this case because the `@lazy` decorator make the variant not compile until used. So maybe we can make an exception for this case.

dcodeIO:
If both `@lazy` and `@global` are present together, it seems OK to permit potential side-effects for now, given that `@lazy` permits it already.

src/compiler.ts
```diff
@@ -7339,6 +7339,14 @@ export class Compiler extends DiagnosticEmitter {
...
       case ElementKind.Global: {
         let global = <Global>target;
+        if (!global.hasDecorator(DecoratorFlags.Lazy) && global.hasDecorator(DecoratorFlags.Global) && !global.is(CommonFlags.Compiled)) {
+          this.error(
+            DiagnosticCode.Variable_0_used_before_its_declaration,
+            expression.range,
+            global.internalName
+          );
+          return module.unreachable();
+        }
```
dcodeIO, line 7349: Does this cover all possible cases? For example, if the global is not read (`a = t1`), but assigned (`t1 = something`), it looks like this branch alone would not suffice.

HerrCai0907:
replaced by #2632

Issue #2653, opened 2023-02-16
`addMemory` in rt cannot handle more 65536 wasm page memory.


HerrCai0907:
### Bug description

Meet runtime `addMemory` error.

RuntimeError: unreachable
    at ~lib/rt/tlsf/addMemory (wasm://wasm/b69e6f5a:wasm-function[11]:0x47b)
    at ~lib/rt/tlsf/growMemory (wasm://wasm/b69e6f5a:wasm-function[16]:0x853)
    at ~lib/rt/tlsf/allocateBlock (wasm://wasm/b69e6f5a:wasm-function[18]:0x918)
    at ~lib/rt/tlsf/__alloc (wasm://wasm/b69e6f5a:wasm-function[19]:0x965)
    at ~lib/memory/heap.alloc (wasm://wasm/b69e6f5a:wasm-function[20]:0x970)
    at start:assembly/index (wasm://wasm/b69e6f5a:wasm-function[23]:0x9fe)
    at ~start (wasm://wasm/b69e6f5a:wasm-function[21]:0x976)
    at file:///Users/q540239/as/tests/index.js:4:31

### Steps to reproduce

```ts
// align memory page to 2**n
while (memory.size() != 128) {
  heap.alloc(16);
}
// alloc memory to more than 65536/2 wasm page
for (let i = 0; i < 32763; i++) {
  heap.alloc(64 * 1024);
}
```

### AssemblyScript version

0.27.0

PR #2654, opened 2023-02-16
allow `addMemory` to support 65536 wasm pages memory

HerrCai0907:
If wasm uses 65536 pages memory, memory.size() << 16 will be 0 due to overflow.
It causes `assert(start <= end)` failed. (end is 0)
Actually due to assemblyscript runtime will double the memory each time. It will happened when wasm uses 32768 page and want to grow memory.

This PR wants to avoid this assert failed when wasm use 65536 pages.

fixed: #2653


HerrCai0907:
@dcodeIO @MaxGraey Could you kindly review it?

std/assembly/rt/tlsf.ts
```diff
@@ -374,7 +374,7 @@ function prepareBlock(root: Root, block: Block, size: usize): void {
 
 /** Adds more memory to the pool. */
 function addMemory(root: Root, start: usize, end: usize): bool {
-  if (DEBUG) assert(start <= end); // must be valid
+  if (DEBUG) assert(start <= end || end == 0); // must be valid
```
dcodeIO, line 377: In a potential future where `usize` might be 64-bit, it seems like this check does nothing. For 32-bit `usize`, the check feels brittle. Would it perhaps make sense to change this to `end64: u64`, then deriving `let end: usize = <usize>end64` from it after the check has been performed?

HerrCai0907: looks great. I will change it.

dcodeIO (approved):

Issue #2660, opened 2023-03-01
generate null instanceof function


HerrCai0907:
### Bug description

instanceof function only contain unreachable instruction

```wat
 (func $~instanceof|assembly/index/B1 (param $0 i32) (result i32)
  unreachable
 )
```

### Steps to reproduce

```ts
class A {
  foo(): void {}
}
let a: A = new A();
class B1 extends A {
  foo(): void {
    if (a instanceof B1) {
    }
  }
}

let b = new B1();
(b as A).foo();
```

### AssemblyScript version

0.27.0

PR #2661, opened 2023-03-01
fix: `instanceof` in override function compile as unreachable

HerrCai0907:
fix: #2660

switch the order of `compile pending instanceof helpers` and `set up override stubs`


dcodeIO (approved):

dcodeIO:
Good catch, thanks!

Issue #2662, opened 2023-03-03
Could not compare null with nullable type


yjhmelody:
### Bug description

```
$ asc ./foo.ts 
WARNING AS226: Expression resolves to unusual type 'usize'.
   :
 3 │ assert(null != x);
   │        ~~~~
   └─ in foo.ts(3,8)

ERROR TS2365: Operator '!=' cannot be applied to types 'usize' and '~lib/string/String | null'.
   :
 3 │ assert(null != x);
   │        ~~~~~~~~~
   └─ in foo.ts(3,8)

FAILURE 1 compile error(s)
```

### Steps to reproduce

```ts
let x: string | null = "foo";

assert(null != x);
```

### AssemblyScript version

v0.25 ~ v0.27

yjhmelody:
@HerrCai0907 Hi, there is another kind type error

yjhmelody:
I always feel that we always need non-overloadable reference comparison and overloadable comparison

HerrCai0907:
The second issue seems like another problem related with operator overload.

PR #2663, opened 2023-03-04
fix: introduce resolver type before compiling a not eq expresssion

HerrCai0907:
fix: #2662

The idea is compiler should infer the left type according to right type before compiling it.

### DIFF
--- a/src/compiler.ts
+++ b/src/compiler.ts
@@ -3967,7 +3967,16 @@ export class Compiler extends DiagnosticEmitter {
       }
       case Token.Exclamation_Equals_Equals:
       case Token.Exclamation_Equals: {
-        leftExpr = this.compileExpression(left, contextualType);
+        let resolver = this.resolver;
+        let leftType = resolver.resolveExpression(left, this.currentFlow, Type.auto, ReportMode.Swallow);
+        if (leftType == this.options.usizeType || leftType == Type.auto || leftType == null) {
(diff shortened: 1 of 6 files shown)


dcodeIO:
A more comprehensive fix for such cases could be, now that we have `AnyClass <: Object`, to compile `null` as `Object` instead of `usize` if there isn't a more specific context. Haven't tried, but if this works we'd also get rid of the "compiles to unusual type" warning.

HerrCai0907:
@dcodeIO It is a break change, see this test code.
 Of course, maybe we can do it, I do not think it is a good idea to make null as `<usize>0`. it just likes nullptr in cpp and NULL in c.

```ts
{ // only nulls infers as usize[]
  let arr1 = [null];
  assert(isInteger(arr1[0]));
  assert(!isNullable(arr1[0]));
  let arr2 = [null, null];
  assert(isInteger(arr2[0]));
  assert(!isNullable(arr2[0]));
}
{ // null in integer contexts infers as usize
  let arr1 = [1, null];
  assert(isInteger(arr1[0]));
  assert(!isNullable(arr1[0]));
  let arr2 = [null, 1];
  assert(isInteger(arr2[0]));
  assert(!isNullable(arr2[0]));
}
```

HerrCai0907:
After trying it, null should not be `~lib/object/Object`, consider this case, `[null, "str"]` should be inferred as `Array<string>`, but `[object, "str"]` should be inferred as `Array<object>`

CountBleck (commented):
I can't say anything meaningful about big changes to the type system.

PR #2664, opened 2023-03-07
chore: add prettier to format

HerrCai0907:
Because prettier v3 will support ESM plugin, it make possible for assemblyscript prettier plugin to format code.
This PR want to introduce the prettier as formatting tooling.


HerrCai0907:
@dcodeIO @MaxGraey 
Add a formatter is help us to avoid some code format issue.
WDYT?

PR #2669, opened 2023-03-10
feat: add getOrDefault to Map

yjhmelody:
Add `getOrDefault` to prevent panic when get value of map.

- [ x ] I've read the contributing guidelines
- [ x ] I've added my name and email to the NOTICE file

### DIFF
--- a/std/assembly/index.d.ts
+++ b/std/assembly/index.d.ts
@@ -2043,6 +2043,7 @@ declare class Map<K,V> {
   has(key: K): bool;
   set(key: K, value: V): this;
   get(key: K): V;
+  getOrDefault(key: K, def: V): V;
   delete(key: K): bool;
   clear(): void;
   keys(): K[]; // preliminary
--- a/std/assembly/map.ts
+++ b/std/assembly/map.ts
@@ -106,6 +106,12 @@ export class Map<K,V> {
     return entry.value;
   }
 
+  getOrDefault(key: K, def: V): V {
+    let entry = this.find(key, HASH<K>(key));
+    if (!entry) return def;
+    return entry.value;
+  }
+
   @operator("[]=")
   set(key: K, value: V): this {
     let hashCode = HASH<K>(key);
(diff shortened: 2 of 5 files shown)


yjhmelody:
I do not know to update the fixture

tests/compiler/std/math.release.wat
```diff
@@ -49900,7 +49900,7 @@
    call $~lib/builtins/abort
    unreachable
   end
-  f64.const -nan:0x8000000000000
+  f64.const nan:0x8000000000000
```
HerrCai0907, line 49903: Change in this file should be reverted. I guess you use mac arm64 as develop machine. I have the same problem.

yjhmelody: Ok

yjhmelody:
Hi @dcodeIO Could this be merged? Though It's not a method of Map in JS, but it's very helpful to reduce twice lookup to once.

dcodeIO:
Generally agree with the observation that it would be good to do something about double lookups. Feels somewhat suboptimal that we'd have to add a non-standard method for this, hence I'm wondering if there are alternatives.

Starting from the underlying issue, that is that `Map#get` traps when a respective entry does not exist, which is already non-standard (since `undefined` doesn't exist), thinkable approaches could be:

* Add a non-standard method `Map#getOrDefault` as proposed, adding additional non-standardness
* Add a second argument to `Map#get`, which is an optional default value (traps if not provided), changing where the non-standardness materializes
* Something else?

Thoughts?

dcodeIO:
Regarding the alternative of modifying `Map#get` with a second parameter, a potential implementation could use `defaultValue: V = changetype<V>(0)`, and only trap when `V` is a non-nullable type and `defaultValue == 0` - the default if `defaultValue` is omitted. This would be a slight breaking change, in that if `V` is nullable, `get` would never trap but return `null` if a key does not exist. One could argue that this breaking change makes it more JS-y in the typical case.

yjhmelody:
Well, should I close this PR?

PR #2678, opened 2023-03-30
Start adding alignment hints

Heath123:
This adds some alignment hints to speed up unaligned access on some (admittedly obscure) platforms (but it's probably still less obscure than `lowMemoryLimit`). I'm not 100% sure that they are correct especially for the one where I modified the compiler but I think it's right. There will definitely be more cases of unaligned access that need annotating, to get them all I'll need to find a way to run the test cases in a way that errors on unaligned access like the Binaryen SafeHeap pass does: https://github.com/WebAssembly/binaryen/blob/main/src/passes/SafeHeap.cpp#L303.


Heath123:
Done

PR #2687, opened 2023-04-19
Implement stringref operations

dcodeIO:
Implements the current state of the Reference-Typed Strings proposal (currently phase 1) as supported by Binaryen, for now as built-ins to enable experimentation. A constant reference-typed string can be created using a string literal in a `ref_string` context.

Enable with `--enable stringref` in AS, `--experimental-wasm-stringref` in V8.

| Instruction | Binaryen status
|-------------|------------------
| string.const | ✔️
| string.new_utf8 | ⏳ Assumes single memory
| string.new_utf8_array | ✔️
| string.new_lossy_utf8 | ⏳ Assumes single memory
| string.new_lossy_utf8_array | ✔️
| string.new_wtf8 | ⏳ Assumes single memory
| string.new_wtf8_array | ✔️
| string.new_wtf16 | ⏳ Assumes single memory
| string.new_wtf16_array | ✔️
| string.from_code_point | ✔️
| string.hash | ✔️
| string.measure_utf8 | ✔️
| string.measure_wtf8 | ✔️
| string.measure_wtf16 | ✔️
| string.is_usv_sequence | ✔️
| string.encode_utf8 | ⏳ Assumes single memory
| string.encode_utf8_array | ✔️
| string.encode_lossy_utf8 | ❌
| string.encode_lossy_utf8_array | ❌
| string.encode_wtf8 | ⏳ Assumes single memory
| string.encode_wtf8_array | ✔️
| string.encode_wtf16 | ⏳ Assumes single memory
| string.encode_wtf16_array | ✔️
| string.concat | ✔️
| string.eq | ✔️
| string.compare | ✔️
| string.as_wtf8 | ✔️
| string.as_wtf16 | ✔️
| string.as_iter | ✔️
| stringview_wtf8.advance | ✔️
| stringview_wtf8.encode_utf8 | ❌
| stringview_wtf8.encode_lossy_utf8 | ❌
| stringview_wtf8.encode_wtf8 | ❌
| stringview_wtf8.slice | ✔️
| stringview_wtf16.length | ✔️
| stringview_wtf16.slice | ✔️
| stringview_wtf16.get_codeunit | ✔️
| stringview_wtf16.encode | ❌
| stringview_iter.next | ✔️
| stringview_iter.advance | ✔️
| stringview_iter.rewind | ✔️
| stringview_iter.slice | ✔️

✔️ Available &nbsp; ⏳ Available but incomplete &nbsp; ❌ Not available

Possible follow-up work in case there is interest: Populate the `RefString` wrapper class using the instructions to mimic `String`, so it can eventually replace `String`. So far this PR implements a few string functions as a POC.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -74,16 +74,16 @@ jobs:
     - uses: dcodeIO/setup-node-nvm@master
       with:
         node-mirror: https://nodejs.org/download/v8-canary/
-        node-version: 19.0.0-v8-canary202209029fc5a9347b
+        node-version: 21.0.0-v8-canary20230419061e93e884
     - name: Install dependencies
       run: npm ci --no-audit
     - name: Build
       run: npm run build
     - name: Test experimental features
       env:
-        ASC_FEATURES: threads,reference-types,gc,exception-handling
+        ASC_FEATURES: threads,reference-types,gc,exception-handling,stringref
       run: |
-        npm run test:compiler features/threads features/reference-types features/gc features/exception-handling
+        npm run test:compiler features/threads features/reference-types features/gc features/exception-handling features/stringref
   runtimes:
     name: "Runtimes"
     runs-on: ubuntu-latest
--- a/src/builtins.ts
+++ b/src/builtins.ts
@@ -715,7 +715,6 @@ export namespace BuiltinNames {
   export const i31_new = "~lib/builtins/i31.new";
   export const i31_get = "~lib/builtins/i31.get";
 
-  export const string_const = "~lib/builtins/string.const_";
   export const string_new_utf8 = "~lib/builtins/string.new_utf8";
   export const string_new_utf8_array = "~lib/builtins/string.new_utf8_array";
   export const string_new_lossy_utf8 = "~lib/builtins/string.new_lossy_utf8";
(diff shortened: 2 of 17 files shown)


dcodeIO:
With #2689 in place, instructions that technically accept but would throw when provided a `stringref` that is `null` now take `ref_string` (non-nullable `stringref`), effectively eliminating the possibility of such traps statically per the type system. Similarly, instructions that cannot produce `null` use `ref_string` as their return type, with tests showing that V8 materializes a non-nullable type there as well, i.e. respective assignments succeeding. One exception is `string.eq`, as it can deal with `null` inputs.

dcodeIO:
There is some proof of concept code now, showing how `RefString` can be populated with functionality utilizing the stringref instruction set. Basically, in a `ref_string` type context, a string literal compiles to a `string.const`, making a reference-typed string. Such a value of type `ref_string` is backed by the `RefString` wrapper class (in std/assembly/references.ts) providing the JS-like API that mostly vanishes after optimizations, leaving primarily the string instructions. So the underlying Wasm type in AS is simply `(ref string)` that can be exchanged with JS and Web APIs by reference.

dcodeIO:
Quick update: There has been a more general [string discussion](https://github.com/WebAssembly/meetings/pull/1330/files#diff-2d9a18ed4699ee5ee6e3dacf178049832a6d36108ceabee5e720613a3d570c3cR94-R227) in CG context meanwhile. Iiuc, an alternative to `stringref` that instead uses (non-portable?) compile time imports for common string operations has been discussed. Also noteworthy:

> The Wasmtime project would prefer not to implement `stringref`. We don't have a managed host string to expose to Wasm guests, and we don't want to have to implement one. We would rather that Wasm guests implement their own strings on top of the GC proposal, where they can build exactly what they need, which can be very different across languages. As far as interoperability of strings across different languages goes, we believe this should be addressed at the component model and interface types layer. For tight integration with JS strings on the Web, string imports seem like a good solution.

Translated: Our thing that intentionally does not address the thing should be off-label responsible for the thing so we can not address the thing and eventually the Web becomes the special case - it's <del>Business</del>Wasmtime!

PR #2688, opened 2023-04-20
Issue 2686

### LINKED ISSUE #2686: Part of debug info missed for template function
### Bug description

After compilation, some debug info (sourcemap) missed for `$assembly/index/testFoo<assembly/index/AA|null>` funciton.

I prepared a reproduce repo: https://github.com/JesseCodeBones/miss_debug_info
for this function, this part of debug info will be missed:  
```
  if(v.length > 10) {
    return 2;
  } else {
    return 3;
  }
```

below is the output for function `$assembly/index/testFoo<assembly/index/AA|null>` and `$assembly/index/testFoo<assembly/index/AA>` 

```WAT
(func $assembly/index/testFoo<assembly/index/AA|null> (param $v i32) (result i32)
  (local $1 i32)
  (global.set $~lib/memory/__stack_pointer
   (i32.sub
    (global.get $~lib/memory/__stack_pointer)
    (i32.const 4)
   )
  )
  (call $~stack_check)
  (i32.store
   (global.get $~lib/memory/__stack_pointer)
   (i32.const 0)
  )
  ;;@ assembly/index.ts:6:28
  (if
   (i32.eq
    ;;@ assembly/index.ts:6:23
    (local.get $v)
    ;;@ assembly/index.ts:6:28
    (i32.const 0)
   )
   (block
    ;;@ assembly/index.ts:7:11
    (local.set $1
     (i32.const 1)
    )
    (global.set $~lib/memory/__stack_pointer
     (i32.add
      (global.get $~lib/memory/__stack_pointer)
      (i32.const 4)
     )
    )
    ;;@ assembly/index.ts:7:4
    (return
     (local.get $1)
    )
   )
  )
  (local.set $1
   (local.get $v)
  )
  (i32.store
   (global.get $~lib/memory/__stack_pointer)
   (local.get $1)
  )
  (if
   (i32.gt_s
    (call $assembly/index/AA#get:length <--------------------**from here the debug info missed ** 
     (local.get $1)
    )
    (i32.const 10)
   )
   (block
    (local.set $1
     (i32.const 2)
    )
    (global.set $~lib/memory/__stack_pointer
     (i32.add
      (global.get $~lib/memory/__stack_pointer)
      (i32.const 4)
     )
    )
    (return
     (local.get $1)
    )
   )
   (block
    (local.set $1
     (i32.const 3)
    )
    (global.set $~lib/memory/__stack_pointer
     (i32.add
      (global.get $~lib/memory/__stack_pointer)
      (i32.const 4)
     )
    )
    (return
     (local.get $1)
    )
   )
  )
 )
```

but debug info for `$assembly/index/testFoo<assembly/index/AA>` is good:  
```WAT
(func $assembly/index/testFoo<assembly/index/AA> (param $v i32) (result i32)
  (local $1 i32)
  (global.set $~lib/memory/__stack_pointer
   (i32.sub
    (global.get $~lib/memory/__stack_pointer)
    (i32.const 4)
   )
  )
  (call $~stack_check)
  (i32.store
   (global.get $~lib/memory/__stack_pointer)
   (i32.const 0)
  )
  ;;@ assembly/index.ts:6:6
  (drop
   (i32.const 0)
  )
  ;;@ assembly/index.ts:9:6
  (local.set $1
   (local.get $v)
  )
  (i32.store
   (global.get $~lib/memory/__stack_pointer)
   (local.get $1)
  )
  ;;@ assembly/index.ts:9:17
  (if
   (i32.gt_s
    ;;@ assembly/index.ts:9:6
    (call $assembly/index/AA#get:length
     (local.get $1)
    )
    ;;@ assembly/index.ts:9:17
    (i32.const 10)
   )
   (block
    ;;@ assembly/index.ts:10:11
    (local.set $1
     (i32.const 2)
    )
    (global.set $~lib/memory/__stack_pointer
     (i32.add
      (global.get $~lib/memory/__stack_pointer)
      (i32.const 4)
     )
    )
    ;;@ assembly/index.ts:10:4
    (return
     (local.get $1)
    )
   )
   (block
    ;;@ assembly/index.ts:12:11
    (local.set $1
     (i32.const 3)
    )
    (global.set $~lib/memory/__stack_pointer
     (i32.add
      (global.get $~lib/memory/__stack_pointer)
      (i32.const 4)
     )
    )
    ;;@ assembly/index.ts:12:4
    (return
     (local.get $1)
    )
   )
  )
 )
```

### Steps to reproduce

1. create AS file with below code:
```
 class AA {
  length: i32 = 1
}

function testFoo<T>(v: T): i32 {
  if (isNullable(v) && v == null) {
    return 1;
  }
  if (v.length > 10) {
    return 2;
  } else {
    return 3;
  }
}
export function testExport(): void {
  testFoo<AA | null>(null);
  const a = new AA();
  testFoo(a);
}

 ```
2. run compile command: `asc assembly/index.ts --debug --disable bulk-memory --sourceMap --textFile --outFile dist/test.wasm`
3. run `[wasm-opt executable] dist/test.wasm -ism dist/test.wasm.map --print`

### AssemblyScript version

v0.27.2

### PR DESCRIPTION
**Pull request by JesseCodeBones**
Fixes #2686

* optimize data struct of `range` and `expressRef`, `range 1 --> n expressionRef`.
* add capacity for complier test to verify the sourceMap output

- [X] I've read the contributing guidelines
- [X] I've added my name and email to the NOTICE file

### DIFF
--- a/src/compiler.ts
+++ b/src/compiler.ts
@@ -9830,7 +9830,7 @@ export class Compiler extends DiagnosticEmitter {
     let targetFunction = this.currentFlow.targetFunction;
     let source = range.source;
     if (source.debugInfoIndex < 0) source.debugInfoIndex = this.module.addDebugInfoFile(source.normalizedPath);
-    range.debugInfoRef = expr;
+    range.debugInfo.set(targetFunction.internalName, expr);
     targetFunction.debugLocations.push(range);
   }
 
--- a/src/compiler.ts
+++ b/src/compiler.ts
@@ -93,7 +93,8 @@ import {
   PropertyPrototype,
   IndexSignature,
   File,
-  mangleInternalName
+  mangleInternalName,
+  DebugLocation
 } from "./program";
 
 import {
@@ -9830,8 +9831,7 @@ export class Compiler extends DiagnosticEmitter {
     let targetFunction = this.currentFlow.targetFunction;
     let source = range.source;
     if (source.debugInfoIndex < 0) source.debugInfoIndex = this.module.addDebugInfoFile(source.normalizedPath);
-    range.debugInfo.set(targetFunction.internalName, expr);
-    targetFunction.debugLocations.push(range);
+    targetFunction.debugLocations.push(new DebugLocation(expr, range));
   }
 
   /** Checks whether a particular function signature is supported. */
--- a/src/diagnostics.ts
+++ b/src/diagnostics.ts
@@ -11,6 +11,7 @@ import {
   DiagnosticCode,
   diagnosticCodeToString
 } from "./diagnosticMessages.generated";
+import { ExpressionRef } from "./module";
 
 import {
   isLineBreak,
@@ -45,7 +46,7 @@ export const enum DiagnosticCategory {
 export class Range {
 
   source!: Source;
-  debugInfoRef: usize = 0;
+  debugInfo: Map<string, ExpressionRef> = new Map<string, ExpressionRef>();
 
   constructor(public start: i32, public end: i32) {}
 
--- a/src/diagnostics.ts
+++ b/src/diagnostics.ts
@@ -11,7 +11,6 @@ import {
   DiagnosticCode,
   diagnosticCodeToString
 } from "./diagnosticMessages.generated";
-import { ExpressionRef } from "./module";
 
 import {
   isLineBreak,
@@ -46,7 +45,6 @@ export const enum DiagnosticCategory {
 export class Range {
 
   source!: Source;
-  debugInfo: Map<string, ExpressionRef> = new Map<string, ExpressionRef>();
 
   constructor(public start: i32, public end: i32) {}
 
--- a/src/program.ts
+++ b/src/program.ts
@@ -118,7 +118,8 @@ import {
   Module,
   FunctionRef,
   MemorySegment,
-  getFunctionName
+  getFunctionName,
+  ExpressionRef
 } from "./module";
 
 import {
@@ -3916,9 +3917,11 @@ export class Function extends TypedElement {
       for (let i = 0, k = debugLocations.length; i < k; ++i) {
         let range = debugLocations[i];
         let source = range.source;
+        const debugInfo = range.debugInfo.get(this.internalName);
+        assert(debugInfo);
         module.setDebugLocation(
           ref,
-          range.debugInfoRef,
+          debugInfo as ExpressionRef,
           source.debugInfoIndex,
           source.lineAt(range.start),
           source.columnAt() - 1 // source maps are 0-based
--- a/src/program.ts
+++ b/src/program.ts
@@ -422,6 +422,13 @@ export namespace OperatorKind {
     return OperatorKind.Invalid;
   }
 }
+/** DebugLocation for debug info generation */
+export class DebugLocation {
+  /** Constructs a new Debug Location*/
+  constructor(
+    public expressionRef: ExpressionRef,
+    public range: Range) {}
+}
 
 /** Represents an AssemblyScript program. */
 export class Program extends DiagnosticEmitter {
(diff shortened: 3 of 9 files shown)

### CODE REVIEW

**Comment by dcodeIO** [comment]
Having one map per range seems excessive, memory-wise. Am I assuming correctly that the underlying issue was that the same debug info reference cannot be reused on multiple output expressions, hence there must be multiple copies? If so, is there perhaps a better way to make such copies?

**Comment by JesseCodeBones** [comment]
> Having one map per range seems excessive, memory-wise. Am I assuming correctly that the underlying issue was that the same debug info reference cannot be reused on multiple output expressions, hence there must be multiple copies? If so, is there perhaps a better way to make such copies?

yeah, logically same range can generate multi `expressionRef`, from the trace we found it can happened that `debugLocations` contains the expression reference that donot belong to current function, we found this issue with template function.

** debug info insert trace ** ->
```
targetFunctionName=assembly/index/testFoo<assembly/index/AA|null>, debugFunctionName=assembly/index/testFoo<assembly/index/AA|null>, expressionRef=317368
targetFunctionName=assembly/index/testFoo<assembly/index/AA|null>, debugFunctionName=assembly/index/testFoo<assembly/index/AA>, expressionRef=496956
targetFunctionName=assembly/index/testFoo<assembly/index/AA|null>, debugFunctionName=assembly/index/testFoo<assembly/index/AA|null>, expressionRef=317384
targetFunctionName=assembly/index/testFoo<assembly/index/AA|null>, debugFunctionName=assembly/index/testFoo<assembly/index/AA>, expressionRef=496968
```
I suppose it is better to correct the data structure of `range` and `expressionRef`.
solutions could be:
1. create new structure for `debugLocation` like: {range, functionRef, expressionRef}
2. current solution, range debug info should be a map {functionRef, expressionRef}
for the memory usage issue, normally `debugLocation` belongs to `Function` class, so the data has been splited with function level, I think it is acceptable.
WDYT

**Comment by dcodeIO** [comment]
Hmm, what if we'd instead clone the range when calling `addDebugLocation`, so `debugLocations` is unique for each function, with each range having its own `debugInfoRef`? Or perhaps remove the caching entirely, and add `debugInfoRef` on demand? Haven't looked into this in detail yet, so I might be missing something, it's just that additional maps seem not needed.

**Comment by JesseCodeBones** [comment]
> Hmm, what if we'd instead clone the range when calling `addDebugLocation`, so `debugLocations` is unique for each function, with each range having its own `debugInfoRef`? Or perhaps remove the caching entirely, and add `debugInfoRef` on demand? Haven't looked into this in detail yet, so I might be missing something, it's just that additional maps seem not needed.

yeah, explicit producer and consumer mode is doable, that means functions should manage their range reference and process sequence carefully.
I didnot get the root cause why template function could impact each other, but it is better to increase Robust with datastructure level😁

**Comment by JesseCodeBones** [comment]
@dcodeIO 
we have four options for fixing this issue
* copy `Range` to each function
  * disadvantage: previously `Range` follows the AST, this solution will break this principle, and copy action will generate more CPU cost
* create new data structure for debug line info {fileIndex, line, column, expressionPtr, functionExpr}
  * disadvantage: then the debug info will miss the range information, maybe not convenient for developer in future
* change the `Range` data structure from `expressionRef` to `Map<string, ExpressionRef>`
  * disadvantage: new map will bring more memory cost (I think it is OK)
* check why the template function will impact each other and fix it
  * disadvantage: developer need to control the consumer and producer flow carefully, the issue still may happen in future and very hard to debug the root cause
 Above is the solutions I can imagine, can you help to figure out which direction we should follow and if you have comments for the direction, please let me know.

**Comment by dcodeIO** [comment]
Perhaps, would a change like this achieve the desired result?

```diff
   /** Adds the debug location of the specified expression at the specified range to the source map. */
   addDebugLocation(expr: ExpressionRef, range: Range): void {
     let targetFunction = this.currentFlow.targetFunction;
     let source = range.source;
     if (source.debugInfoIndex < 0) source.debugInfoIndex = this.module.addDebugInfoFile(source.normalizedPath);
-    range.debugInfoRef = expr;
-    targetFunction.debugLocations.push(range);
+    targetFunction.debugLocations.set(range, expr); // here
   }
```

So there is no longer a `debugInfoRef` on ranges (can be removed), but instead a per-`targetFunction` mapping of ranges to expressions (here replacing `debugLocations` with a `Map`)?

Changqing-JING:
If debugLocations is changed to Map, the debugInfoRef of Range can also be remove.

JesseCodeBones:
@dcodeIO 
yeap, this is could be a better solution, can you review the change again? Thank you

Issue #2646, opened 2023-02-09
Cannot find global type 'CallableFunction'


ckaznocha:
### Bug description

Opening an assemblyscript project with vscode results in the error "Cannot find global type 'CallableFunction'" from the `tsconfig.json`. It looks like this is caused by the `noLib` being true and no global definition for `CallableFunction` in the assemblyscript d.ts files. This only started for me after I updated to the latest version of vscode which I'm under the impression comes with a bundled version of typescript.

I found some folks on so are having the same problem with no resolution but I wasn't able to find an issue here so creating one.

https://stackoverflow.com/questions/74438208/cannot-find-global-type-callablefunction-error-in-tsconfig-json-file

### Steps to reproduce

Open an assemblyscript project with vscode v1.75.0 then edit any ts file.

### AssemblyScript version

v0.26.7

dcodeIO:
Seems we'll need to declare a mock `CallableFunction` in [std/index.d.ts](https://github.com/AssemblyScript/assemblyscript/blob/main/std/assembly/index.d.ts), where the `noLib` environment is declared, so TS tooling can pick it up. Might be as simple as copy-pasting the interface TS uses.

ckaznocha:
Thanks for posting this workaround @neodon! I tried it in a project and can confirm it suppresses the new errors from the `tsconfig.json`.

PR #2712, opened 2023-06-21
fix: Cannot find global type 'CallableFunction'

XMadrid:
Fixes [#2646 ](https://github.com/AssemblyScript/assemblyscript/issues/2646)

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -55,6 +55,7 @@ under the licensing terms detailed in LICENSE:
 * CountBleck <Mr.YouKnowWhoIAm@protonmail.com>
 * Abdul Rauf <abdulraufmujahid@gmail.com>
 * Bach Le <bach@bullno1.com>
+* Xinquan Xu <xinquan0203@163.com>
 
 Portions of this software are derived from third-party works licensed under
 the following terms:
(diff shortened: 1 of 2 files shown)


CountBleck (changes requested):
I was going to manually specify some changes to improve the `call`/`bind`/`apply` typings, but it looks like https://github.com/microsoft/TypeScript/commit/33ab6fd0d5eceb7715000398382b60d64dde1c67 has exactly what I mean.

Using `A extends any[]` and `...args: A` allows functions with any number and type of arguments to match, which removes unnecessary bloat.

XMadrid:
> I was going to manually specify some changes to improve the `call`/`bind`/`apply` typings, but it looks like [microsoft/TypeScript@33ab6fd](https://github.com/microsoft/TypeScript/commit/33ab6fd0d5eceb7715000398382b60d64dde1c67) has exactly what I mean.
> 
> Using `A extends any[]` and `...args: A` allows functions with any number and type of arguments to match, which removes unnecessary bloat.
> 
> Edit: sorry for the three-month delay!

Okay, I aligned the declaration with the newest `typescript/src/lib/es5.d.ts`.

CountBleck (approved):

PR #2718, opened 2023-07-20
Support string literal export names

CountBleck:
⯈ Add support for `export {foo as "b a\r"}`
⯈ Add support for `import {"b a\r" as foo} from "baz";`

### DIFF
--- a/src/ast.ts
+++ b/src/ast.ts
@@ -531,7 +531,7 @@ export abstract class Node {
 
   static createExportMember(
     localName: IdentifierExpression,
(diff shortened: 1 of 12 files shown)


CountBleck:
Currently, bootstrap is broken due to some sort of oversight with inherited fields not becoming part of overrides (at least, that's what I think it is).

As a result, I'll have to make some changes.

CountBleck:
Never mind, I'm stupid and TypeScript doesn't support this syntax for whatever reason.

PR #2722, opened 2023-07-21
Small fixes for tests and CI

CountBleck:
⯈ Ensure all tests are run in CI, including those involving experimental features
⯈ Fix the tokenizer test, as useless as it may be
⯈ Ensure the release fixtures and bindings fixtures are checked
⯈ Fix the typings for the `writeFile` CLI option, since strings are passed in too
⯈ Make VS Code complain less about variables in tests being redeclared in other files

Closes #2723.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -83,7 +83,7 @@ jobs:
       env:
         ASC_FEATURES: threads,reference-types,gc,exception-handling
       run: |
-        npm run test:compiler features/threads features/reference-types features/gc features/exception-handling
+        npm run test:compiler features/threads features/reference-types features/gc features/exception-handling bindings/esm bindings/raw
   runtimes:
     name: "Runtimes"
     runs-on: ubuntu-latest
--- a/cli/index.d.ts
+++ b/cli/index.d.ts
@@ -165,7 +165,7 @@ export interface APIOptions {
   /** Reads a file from disk (or memory). */
   readFile?: (filename: string, baseDir: string) => (string | null) | Promise<string | null>;
   /** Writes a file to disk (or memory). */
-  writeFile?: (filename: string, contents: Uint8Array, baseDir: string) => void | Promise<void>;
+  writeFile?: (filename: string, contents: Uint8Array | string, baseDir: string) => void | Promise<void>;
   /** Lists all files within a directory. */
   listFiles?: (dirname: string, baseDir: string) => (string[] | null) | Promise<string[] | null>;
   /** Handler for diagnostic messages. */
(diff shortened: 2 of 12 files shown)


.github/workflows/test.yml
```diff
@@ -81,9 +81,9 @@ jobs:
       run: npm run build
     - name: Test experimental features
       env:
-        ASC_FEATURES: threads,reference-types,gc,exception-handling
+        ASC_FEATURES: "*"
```
dcodeIO, line 84: Idea here was that we can add features that we know should work on Node canary, or disable them again if something breaks. Given that there is no `--disable` here, perhaps keeping a long list works better, but make sure we update the list more often?

dcodeIO: In the past, what would happen was that we implemented milestone X, and things worked with it, but not with milestone X+1. So during development, it was useful to run specific feature tests locally, but not run them on CI just yet, for example.

.github/workflows/test.yml
```diff
@@ -81,9 +81,9 @@ jobs:
...
       run: |
-        npm run test:compiler features/threads features/reference-types features/gc features/exception-handling
+        npm run test:compiler
```
HerrCai0907, line 86: Does it repeat to run lots of test which has already done in `npm test`?

CountBleck: Sadly, yes. Maybe this can be changed to somehow run only tests requiring `ASC_FEATURES=whatever`.

.github/workflows/test.yml
CountBleck: > In the past, what would happen was that we implemented milestone X, and things worked with it, but not with milestone X+1. So during development, it was useful to run specific feature tests locally, but not run them on CI just yet, for example.

I see why you want to whitelist the features...Is it perhaps better if the regular test jobs compiled every test to match against the fixtures but only this job instantiated the tests, and only according to ASC_FEATURES?

dcodeIO: Perhaps, while prompting us to at least keep Binaryen and early feature tests in sync. What might happen is that we implemented something that Binaryen does't yet have interpreter support for, but not sure it that's relevant as it wouldn't work in manual local testing either and would rather be commented out in the test case itself.

CountBleck:
I just changed the functionality of `ASC_FEATURES` so it only affects whether the relevant tests are instantiated or not. I'll go to sleep for now, but please do let me know if this is a robust solution.

CountBleck:
@dcodeIO @HerrCai0907 This PR should be fully ready as far as I know.

HerrCai0907 (approved):

dcodeIO (approved):

CountBleck:
Okay, @MaxGraey, what do you think?

MaxGraey (approved):

PR #2723, opened 2023-07-21
ci: add missing feature test case in action

HerrCai0907:
add missing feature test case

@CountBleck Could we fix test like this?


CountBleck:
Maaaaaaaayyyybeeeeeee, but I think *maybe* I could try something else...

CountBleck:
The main disadvantage is how there could be more tests utilizing ASC_FEATURES in the future, and if I added them, I doubt I would remember to add them to that list in CI 😛. Then, someone might wonder why a bunch of unrelated changes were included when they implement/touch that feature and run those tests.

PR #2729, opened 2023-07-31
Use unique debug info per concrete function

dcodeIO:
A slightly simpler variant of #2688 using a `Map`, so debug info entries are automatically deduplicated.

fixes #2688, see also #2686


lortonx:
release ?

PR #2756, opened 2023-10-01
fix: interface inheritance cause crash

### LINKED ISSUE #2714: Compiler crash caused by interface inheritance
### Bug description

When an interface extends from a parent interface, any functions/methods that receive or return the BASE interface type will cause the compiler to crash.

### Steps to reproduce

The following example code should compile fine in the typescript playground, but will crash the assemblyscript playground:

```ts
interface BaseInt {
  m(): BaseInt;
}

interface MyInt extends BaseInt {
  s: string;
  m(): MyInt;
}

class MyClass implements MyInt {
  s: string = 'foo'
  m(): MyClass { return this }
}

export function main(): void {
  const c = new MyClass()
  test(c)
}

function test(t: BaseInt): BaseInt {
  return t.m()
}
```

NB: change the arg/return types of the `test()` function to `MyInt` or `MyClass` and it compiles OK. But on the base interface the following exception is thrown:

```test
▌ 
▌ 
▌     at Y.assert (file:///Users/x/Dev/x/node_modules/assemblyscript/dist/assemblyscript.js:9:3369)
▌     at eu.resolveOverrides (file:///Users/x/Dev/x/node_modules/assemblyscript/dist/asc.js:22622:94904)
▌     at As.compile (file:///Users/x/Dev/x/node_modules/assemblyscript/dist/asc.js:22872:5096)
▌     at Object.q1 (file:///Users/x/Dev/x/node_modules/assemblyscript/dist/asc.js:22910:2148)
▌     at Object.nh (file:///Users/x/Dev/x/node_modules/assemblyscript/dist/asc.js:22942:9224)
▌     at async compile (file:///Users/x/Dev/x/packages/compiler/dist/compiler.js:96:46)
▌     at async Command.compileCode (file:///Users/x/Dev/x/packages/cli/dist/cmds/compile.js:28:17)
▌     at async Command.parseAsync (/Users/x/Dev/x/node_modules/commander/lib/command.js:935:5)
▌     at async file:///Users/x/Dev/x/packages/cli/dist/cli.js:35:9
▌ 
▌ 
▌ Thank you!
```

The error above can specifically be isolated to this LOC:

https://github.com/AssemblyScript/assemblyscript/blob/a0c27fadd027a32542420160deff04dc31e10f59/src/resolver.ts#L3063

### AssemblyScript version

#main

HerrCai0907:
Fixes #2714

### DIFF
--- a/tests/compiler/class-implements.release.wat
+++ b/tests/compiler/class-implements.release.wat
@@ -25,7 +25,7 @@
  (global $class-implements/f (mut i32) (i32.const 0))
  (global $class-implements/g (mut i32) (i32.const 0))
  (global $class-implements/h (mut i32) (i32.const 0))
- (global $~lib/memory/__stack_pointer (mut i32) (i32.const 34352))
+ (global $~lib/memory/__stack_pointer (mut i32) (i32.const 34360))
  (memory $0 1)
  (data $0 (i32.const 1036) "<")
  (data $0.1 (i32.const 1048) "\02\00\00\00(\00\00\00A\00l\00l\00o\00c\00a\00t\00i\00o\00n\00 \00t\00o\00o\00 \00l\00a\00r\00g\00e")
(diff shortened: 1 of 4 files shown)


tests/compiler/class-implements.release.wat
```diff
@@ -1703,37 +1703,43 @@
...
-                  block $~lib/arraybuffer/ArrayBufferView
-                   block $~lib/string/String
-                    block $~lib/arraybuffer/ArrayBuffer
-                     block $~lib/object/Object
-                      local.get $0
-                      i32.const 8
-                      i32.sub
-                      i32.load $0
-                      br_table $~lib/object/Object $~lib/arraybuffer/ArrayBuffer $~lib/string/String $~lib/arraybuffer/ArrayBufferView $class-implements/A $class-implements/I $class-implements/C $class-implements/B $class-implements/D $class-implements/J $class-implements/E $class-implements/F $class-implements/A2 $class-implements/I2 $class-implements/A3 $class-implements/A4 $class-implements/B2 $class-implements/B3 $class-implements/B4 $invalid
+          block $class-implements/K
```
CountBleck, line 1706: This isn't really caused by this PR (alone), but it would be nice if these cases for interfaces could be removed someday.

CountBleck (approved):
Seems like it works!

PR #2757, opened 2023-10-01
General cleanup

CountBleck:
Changes proposed in this pull request:
⯈ Skip fixture comparisons in tests if compiling them failed
⯈ Update esbuild
⯈ Clean up the DTS generator
⯈ Update glob
⯈ Marginally clean up the web build thingy
⯈ Add a coverage job to CI; maybe Codecov can be used on it in the future?

- [x] I've read the contributing guidelines
- [x] I've added my name and email to the NOTICE file

PR #2759, opened 2023-10-07
test: avoid test:compiler crash when unexpected compiling error

HerrCai0907:


CountBleck (commented):
Looks like we both solved the same problem: see [here](https://github.com/AssemblyScript/assemblyscript/pull/2757/commits/41c66ce4c76a214efb3c6f1ecdf9e10bd62b10d1)

PR #2760, opened 2023-10-07
BREAKING CHANGE: Improve null comparison

### LINKED ISSUE #2662: Could not compare null with nullable type

HerrCai0907:
Fixes #2662.


src/resolver.ts
```diff
@@ -1061,6 +1061,37 @@ export class Resolver extends DiagnosticEmitter {
...
+    /** Contextual flow. */
+    ctxFlow: Flow
+  ): Type | null {
+    const left = expression.left;
+    const right = expression.right;
+
+    const leftType = this.resolveExpression(left, ctxFlow, Type.auto, ReportMode.Report);
+    if (!leftType) return null;
+    const rightType = this.resolveExpression(right, ctxFlow, Type.auto, ReportMode.Report);
+    if (!rightType) return null;
```
CountBleck, line 1076: This should make sure diagnostics are reported for both expressions like (I think it was) before:
```suggestion
    const leftType = this.resolveExpression(left, ctxFlow, Type.auto, ReportMode.Report);
    const rightType = this.resolveExpression(right, ctxFlow, Type.auto, ReportMode.Report);
    if (!leftType || !rightType) return null;
```

src/resolver.ts
```diff
@@ -1061,6 +1061,37 @@ export class Resolver extends DiagnosticEmitter {
...
+
+    if (left.kind == NodeKind.Null && rightType.isReference) {
+      return rightType.asNullable();
+    }
+    if (right.kind == NodeKind.Null && leftType.isReference) {
+      return leftType.asNullable();
+    }
```
CountBleck, line 1083: Some style:
```suggestion
    if (left.kind == NodeKind.Null && rightType.isReference) return rightType.asNullable();
    if (right.kind == NodeKind.Null && leftType.isReference) return leftType.asNullable();
```

tests/compiler/compare-null.ts
```diff
@@ -0,0 +1,51 @@
...
+  null == "aa";
+  null != "aa";
+  "aa" == null;
+  "aa" != null;
+
+  null === "aa";
+  null !== "aa";
+  "aa" === null;
+  "aa" !== null;
+}
```
CountBleck, line 27: This is something to be very careful with. Your current code is perfectly fine for String, but I have to wonder what happens with classes with operator overloads that have signatures with non-null parameters. In that case, you would need to add some null-checking code for `==` that's equivalent to `a && b ? eqOverload(a, b) : a == b`, where `a` and `b` are the pointers of the LHS and RHS respectively, and `eqOverload` represents the overload function. Similar null-checking code applies to `!=`. You could also opt to error if `==`/`!=` is used on nullable types that have an overload without nullable parameters.

CountBleck (changes requested):
The big question is in my last comment.

CountBleck:
Whoops, sorry, I didn't realize you removed the request to review.

HerrCai0907:
> Whoops, sorry, I didn't realize you removed the request to review.

I find it should handle float and integer promotion more carefully. `commonType` will not promotion i32 to f32.

tests/compiler/compare-null.debug.wat
```diff
@@ -0,0 +1,468 @@
...
+  i32.const 0
+  i32.store $0
+  i32.const 0
+  i32.const 32
+  i32.eq
+  drop
+  i32.const 0
+  i32.const 32
+  i32.ne
+  drop
```
CountBleck, line 386: This doesn't give me a good feeling, because this means `null == overloaded` isn't being compiled correctly, while `overloaded == null` is.

tests/compiler/compare-null.debug.wat
```diff
@@ -0,0 +1,468 @@
...
+  call $~lib/string/String.__ne
+  drop
+  i32.const 0
+  i32.const 32
+  i32.eq
+  drop
+  i32.const 0
+  i32.const 32
+  i32.ne
+  drop
```
CountBleck, line 412: Same here.

CountBleck (changes requested):
This isn't quite right either.

tests/compiler/compare-null.debug.wat
HerrCai0907: But I think it should not block this fix.
The same thing will happen when comparing `new A() == new B()`
We should introduce more strictly check for operator overload.

CountBleck: I suppose so. I don't like having different semantics for `null ==` vs `== null` as well as for `A ==` and `== A`.

HerrCai0907: See #2761

HerrCai0907: > I don't like having different semantics for `null ==` vs `== null` as well as for `A ==` and `== A`.

Agree. This patch should land after we implement better operator overload resolution.

CountBleck: > implement better operator overload resolution

We would still need to special-case `null` somehow. There needs to be some way to differentiate between a value of type `Object | null` and a `null` literal, which has the same type in this PR.

HerrCai0907: Why we need to differentiate between a value of type Object | null and a null literal?

PR #2765, opened 2023-10-12
fix: top level statement compile within namespace

### LINKED ISSUE #2764: Diagnose statement in namespace
### Bug description

statement in namespace does not have any effect. It should be diagnosed.

### Steps to reproduce

```ts
let a = 0;

namespace NS {
  a = 100; // here should be diagnosed
  export function foo(): void {
    trace("a", 1, a);
  }
}

  trace("a", 1, a);
  NS.foo();
}
```

compile and run this code will get
```
a 0
a 0
```

### AssemblyScript version

0.27.x

JesseCodeBones:
Fixes #2764.


tests/parser/namespace.ts
```diff
@@ -13,4 +15,5 @@ declare namespace A {
       var aVar: i32;
     }
   }
+  return; // // ERROR 1108: A 'return' statement can only be used within a function body.
```
HerrCai0907, line 18:
```suggestion
  return; // 1108: A 'return' statement can only be used within a function body.
```

HerrCai0907 (approved):
LGTM

src/parser.ts
```diff
@@ -398,9 +398,7 @@ export class Parser extends DiagnosticEmitter {
               tn.range(declareStart, declareEnd), "declare"
             ); // recoverable
           }
-          if (!namespace) {
```
CountBleck, line 401: This looks like an intentional check.

@dcodeIO Is there something behind this check, or can it be safely removed?

tests/parser/namespace.ts
```diff
@@ -1,4 +1,6 @@
+let outsideVariable:i32 = 0;
```
CountBleck, line 1:
```suggestion
let outsideVariable: i32 = 0;
```

src/parser.ts
HerrCai0907: I have taken a look for compiler part. It handles statements in namespace correctly.

HerrCai0907 (changes requested):
AS and TS has difference behavior for this case.
TS will throw diagnose `Block-scoped variable 'a' used before its declaration` but AS will not.

But I think it is different topic with this PR. It is a generic issue both in namespace and function.
Do you think we should introduce variable scope analysis to handle these cases.
@dcodeIO @CountBleck WDYT.

```ts
let a: number = 0;

namespace NS {
  a = 100;
  export function foo(): void {
    trace("a", 1, a);
  }
  let a = 200;
}

  trace("aaaaa", 1, a);
  NS.foo();
}
```

JesseCodeBones:
> AS and TS has difference behavior for this case. TS will throw diagnose `Block-scoped variable 'a' used before its declaration` but AS will not.
> 
> @dcodeIO @CountBleck WDYT.
> 
> ```ts
> let a: number = 0;
> 
> namespace NS {
>   a = 100;
>   export function foo(): void {
>     trace("a", 1, a);
>   }
>   let a = 200;
> }
> 
> export function _start(): void {
>   trace("aaaaa", 1, a);
>   NS.foo();
> }
> ```



> AS and TS has difference behavior for this case. TS will throw diagnose `Block-scoped variable 'a' used before its declaration` but AS will not.
> 
> But I think it is different topic with this PR. It is a generic issue both in namespace and function. Should we introduce variable scope analysis to handle these cases. @dcodeIO @CountBleck WDYT.
> 
> ```ts
> let a: number = 0;
> 
> namespace NS {
>   a = 100;
>   export function foo(): void {
>     trace("a", 1, a);
>   }
>   let a = 200;
> }
> 
> export function _start(): void {
>   trace("aaaaa", 1, a);
>   NS.foo();
> }
> ```

This issue could happened in any scope: top level, function body, namespace and etc.
More information:
1. ts pass and invalid cases
pass case:
```
a = 42;
var a = 0;
```
invalid case:
```
a = 42;
let a = 0;
```
invalid case:
```
a = 42;
const b = ()=>{
  var a = 0;
}
```

2. asc test case and its compile output
```typescript
let a:i32 = 1;
function foo():void{
  a = 42;
  let a: i32 = 0;
}
foo();
```

```WAT
(module
 (type $none_=>_none (func))
 (global $namespace/a (mut i32) (i32.const 1))
 (global $~lib/memory/__data_end i32 (i32.const 8))
 (global $~lib/memory/__stack_pointer (mut i32) (i32.const 32776))
 (global $~lib/memory/__heap_base i32 (i32.const 32776))
 (memory $0 0)
 (table $0 1 1 funcref)
 (elem $0 (i32.const 1))
 (export "memory" (memory $0))
 (start $~start)
 (func $namespace/foo
  (local $a i32)
  i32.const 42
  global.set $namespace/a
  i32.const 0
  local.set $a
 )
 (func $start:namespace
  call $namespace/foo
 )
 (func $~start
  call $start:namespace
 )
)

```

CountBleck:
> ```webassembly
>  (func $namespace/foo
>   (local $a i32)
>   i32.const 42
>   global.set $namespace/a
>   i32.const 0
>   local.set $a
>  )
> ```
Okay, this one thing is unacceptable. There should be a temporal deadzone above that `let a: i32 = 0;`, meaning a compile-time error (as @HerrCai0907 mentioned) should be emitted. AS should not compile that code, and *never* compile it to that. The meaning of `a` shouldn't change halfway into the function.

CountBleck:
> But I think it is different topic with this PR. It is a generic issue both in namespace and function.

Sure, but this PR shouldn't be merged (yet) if it allows weird stuff as shown above.

CountBleck:
> Totally ignoring statement in namespace is also weird stuff.

Yeah, that `// TODO: else?` is a good place to raise an error.

JesseCodeBones:
In typescript, it has `check file` process before enter the `compile` phase.
the logic can be found in function `isBlockScopedNameDeclaredBeforeUse` of https://github.com/microsoft/TypeScript/blob/main/src/compiler/checker.ts.
In assembly script, we do not have the `check` phase, so it is hard to predict that there is a `declaration expression` after this `assignment expression`.
could we have the same `check` functionality for `Diagnose` purpose?

HerrCai0907 (commented):
Since AS cannot fully support namesapce as TS, maybe diagnose "not implement" as temporary solution.

JesseCodeBones:
> Since AS cannot fully support namesapce as TS, maybe diagnose "not implement" as temporary solution.

Hi Congcong,
finished the diagnose message for parser, please help to review again.
Regards

HerrCai0907 (approved):

PR #2768, opened 2023-10-16
improve default argument const check

HerrCai0907:
Using binaryen provided `ExpressionRunner` to run compilesToConst for more accurate evaluation results.


src/compiler.ts
```diff
@@ -6834,6 +6826,19 @@ export class Compiler extends DiagnosticEmitter {
...
+            initializer,
+            parameterTypes[i],
+            Constraints.ConvImplicit
+          );
+          this.currentFlow = previousFlow;
+          const precomp = module.runExpression(expr, ExpressionRunnerFlags.Default);
+          if (precomp) {
+            operands.push(precomp);
+            continue;
+          }
```
dcodeIO, line 6841: Previously, since the expression was checked before compiling it, its side effects (like updating local states) would only be applied when indeed using the expression. With this change, its side-effects occur always due to compiling it unconditionally, even when the expression is not used, which is likely undesirable.

src/compiler.ts
```diff
@@ -6834,6 +6826,19 @@ export class Compiler extends DiagnosticEmitter {
...
           }
+          let previousFlow = this.currentFlow;
+          this.currentFlow = instance.flow;
+          const expr = this.compileExpression(
+            initializer,
+            parameterTypes[i],
+            Constraints.ConvImplicit
+          );
+          this.currentFlow = previousFlow;
+          const precomp = module.runExpression(expr, ExpressionRunnerFlags.Default);
```
HerrCai0907, line 6837:
```suggestion
          const precomp = module.runExpression(expr, ExpressionRunnerFlags.PreserveSideeffects);
```

src/compiler.ts
HerrCai0907: > With this change, its side-effects occur always due to compiling it unconditionally, even when the expression is not used, which is likely undesirable.

Which kind of side-effects do you mean?
If it is side effect for code, it can be detected by `module.runExpression`.
If it is side effect for compiler, here is in  `instance.flow`, default argument definitely needs to be consider.

PR #2772, opened 2023-10-17
fix: use correct valueType for compileAssignment[NFC]

HerrCai0907:
In `makeAssignment`, `valueType` doesn't match the type of `valueExpr`.
But there are no actual influence for current generated wasm.


CountBleck:
I'm going to close this since it's already part of #2773.

PR #2773, opened 2023-10-17
fix: return the assigned value in assignments

### LINKED ISSUE #2770: assignment chain cause compatibility issue
### Bug description

In Typescript, it will print 
```
set 1
set 1
```

But In Assemblyscript, it will print
```
set 1
get
set 123
```
Which will cause compatibility issue when port code.

```ts
class X {
  get y() {
    console.log("get");
    return 123;
  }
  set y(z) {
    console.log("set", z);
  }
}

let x = new X();
x.y = x.y = 1;
```

### Steps to reproduce

compile and execute below code in both ts and as

### AssemblyScript version

v0.27.x

HerrCai0907:
In TS, assignment chain will assign result to each variable. But in AS, assignment chain will assign result to second variable and assign the value of second variable to the first variable.
It will cause incompatibility problem.
This patch assign result to a temporary local and assign it to each variable.
Fixes #2770.


HerrCai0907:
deps: #2772

HerrCai0907:
WDYT @CountBleck @dcodeIO @MaxGraey ?

CountBleck (approved):
Everything seems good here

Issue #2751, opened 2023-09-28
Compiling release without sourceMap create a sourcemap named "false"


DaGaMx:
### Bug description

I am new to AssemblyScript and followed your tutorial:
https://www.assemblyscript.org/getting-started.html#setting-up-a-new-project

I checked the output files in a hex editor and was wondering that the release ends with "sourceMappingURL" "false".
Then I saw that a file named "false" is generated in the output folder.

### Steps to reproduce

Execute this:
npx asc "C:\test\assembly\index.ts" --target release -o "C:\test\assembly\output\release\index.wasm" -Ospeed --noAssert --optimizeLevel 3

In the output folder "C:\test\assembly\output\release\" will be a file named "false".
The output file "C:\test\assembly\output\release\index.wasm" will contain a reference to that source map.

As far as I understand the compiler manual, the source map should only be generated if I specify --sourceMap.

### AssemblyScript version

v0.27.10

woruyu:
Ask a question:
I follows your step to run `npx asc "C:\test-as\assembly\index.ts"  --target release -o "C:\test-as\assembly\output\release\index.wasm" -Ospeed --noAssert --optimizeLevel 3` but I can't get a file named "false". And I guess the reason that you get the source map is in asconfig.json, the sourceMap field is true.

DaGaMx:
Sorry, there was a minor thing I changed in comparison to your "Getting started" manual.
If I do the same steps like in the manual, then I get the "./index.wasm.map" entry like you.
But if I update the asconfig.json to this:
`{
  "targets": {
    "debug": {
      "outFile": "build/debug.wasm",
      "textFile": "build/debug.wat",
      "sourceMap": true,
      "debug": true
    },
    "release": {
      "outFile": "build/release.wasm",
      "textFile": "build/release.wat",
      "sourceMap": false,
      "optimizeLevel": 3,
      "shrinkLevel": 0,
      "converge": false,
      "noAssert": false
    }
  },
  "options": {
    "bindings": "esm"
  }
}`
Then I get the false entry and the false file.

This also happens if I compile using "npm run asbuild".

PR #2775, opened 2023-10-24
fix: string field with false value in asconfig.js

woruyu:
Fixes https://github.com/AssemblyScript/assemblyscript/issues/2751

### DIFF
--- a/util/options.d.ts
+++ b/util/options.d.ts
@@ -68,3 +68,6 @@ export function resolvePath(path: string, baseDir: string, useNodeResolution?: b
 
 /** Populates default values on a parsed options result. */
 export function addDefaults(config: Config, options: OptionSet): void;
+
+/** Sanitizes an option value to be a valid value of the option's type. */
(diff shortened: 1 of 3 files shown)


util/options.d.ts
```diff
@@ -68,3 +68,6 @@ export function resolvePath(path: string, baseDir: string, useNodeResolution?: b
 
 /** Populates default values on a parsed options result. */
 export function addDefaults(config: Config, options: OptionSet): void;
+
+/** Sanitizes an option value to be a valid value of the option's type. */
+export function sanitizeValue(value: any, type: string): boolean | number | string | null | number[] | string[]; // eslint-disable-line @typescript-eslint/no-explicit-any
```
HerrCai0907, line 73: unrelated change

CountBleck: Hm, this is only necessary because this function is tested directly...is there a way to avoid that?

HerrCai0907: Could you test this case by using `merge` API?

HerrCai0907:
fixed in #2802

PR #2779, opened 2023-11-02
ci: build action-node to install nodejs

HerrCai0907:
since all of issue mentioned in `dcodeIO/setup-node-nvm` are resolved.
Now `actions/setup-node` is available.


CountBleck (approved):
This change also happens to allow caching npm dependencies in CI.

PR #2781, opened 2023-11-06
fix: improve diagnose for invalid generic function call

### LINKED ISSUE #2780: Parser error, i think
### Bug description

I understand that I am doing the wrong things in the assembly script, but it seems to me that this error should not look like what is shown below. Hope this helps improve Assemblyscript.
image

### Steps to reproduce

run

```js
let tries = 64;
const aliof = alignof<typeof tries>();
// or
const siof = sizeof<typeof tries>();
```

### AssemblyScript version

v0.27.9

HerrCai0907:
Original implement treat invalid generic function call `Expr '<' Expr '>' '(' ')'` as less than and greater expression.
This patch detect it and throw "expect type" diagnose.

Fixes #2780


CountBleck (commented):
Arguably, this seems like a hack, since it's special-casing those sets of tokens.

Also, `a<12345>()` is valid TS, since `12345` is a type (but we don't support that!).

The proper fix would be implementing the `typeof` syntax in types (with a "Not yet implemented" diagnostic).

Also, TS seems to dislike [this](https://www.typescriptlang.org/play?#code/IYGwpgTgLgPMAEBqeAjAfACgEQAsCWWAlEA) bit of code, but this patch would erroneously treat it as a call to a generic function.

HerrCai0907:
> Also, TS seems to dislike this bit of code, but this patch would erroneously treat it as a call to a generic function.

This code will not match this pattern. It only match empty parentheses and it is definitely not a comparison.

PR #2783, opened 2023-11-09
fix failed test

HerrCai0907:


HerrCai0907:
It looks like arm64 mac will get difference result in CI

Issue #2790, opened 2023-11-14
incorrect variable resolving for class default initializer


HerrCai0907:
### Bug description

incorrect variable resolving.
```ts
let a = 10;
class A {
  v: i32 = a;
  constructor(a: f64) {}
}

new A(1.0);
```
```bash

ERROR AS200: Conversion from type 'f64' to 'i32' requires an explicit cast.
   :
 3 │ v: i32 = a;
   │          ~
   └─ in assembly/index.ts(3,12)
```

### Steps to reproduce

```bash
npm run asbuild
```

### AssemblyScript version

latest

PR #2791, opened 2023-11-14
fix: incorrect variable resolving for class default initializer

HerrCai0907:
Introducing new default constructor to avoid name conflict when do default initializer.
fixes: #2790

### DIFF
--- a/src/common.ts
+++ b/src/common.ts
@@ -165,6 +165,7 @@ export namespace CommonNames {
   // objects
   export const this_ = "this";
   export const super_ = "super";
+  export const defaultConstructor = "constructor|default";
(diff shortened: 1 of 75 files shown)


CountBleck (commented):
I can't review in depth at the moment (I've got overdue homework and all), but why does this PR touch so many fixtures?

HerrCai0907:
> I can't review in depth at the moment (I've got overdue homework and all), but why does this PR touch so many fixtures?

It introduce new default constructor

CountBleck:
> It introduce new default constructor

I don't follow. Isn't this just an issue with scoping?

PR #2794, opened 2023-11-14
feat: support class member type infer

HerrCai0907:
Introduce new `ResolvedType` to make lazy type inference possible.


CountBleck (commented):
It looks like the compiler requires an AST node *somewhere* because, minus the requirement in the parser, the root of this issue is [`program.ts:3995`](https://github.com/AssemblyScript/assemblyscript/blob/v0.27.17/src/program.ts#L3995) and `program.ts:4009`.

Here, stub getter/setter declarations are made with the type node from the field declaration, and the compiler has no way to obtain the initializer or its true type when the getter/setter is resolved. So some sort of hack (like this) is necessary (or a big refactor for type resolution).

By the way, does the following (using generic type parameters in initializers) work?

```ts
class Box<T> { value: T | null = null; }
class Example<T> {
    blaBlaBla = new Box<T>();
}
```

HerrCai0907:
> So some sort of hack (like this) is necessary (or a big refactor for type resolution).

Agree. I don't want to refactor the whole type resolution at this time.

> By the way, does the following (using generic type parameters in initializers) work?

Unfortunately it doesn't work. At current time, only `Function` can contain a `contextualTypeArguments`. But It can be extend later.

HerrCai0907:
> By the way, does the following (using generic type parameters in initializers) work?
> 
> ```ts
> class Box<T> { value: T | null = null; }
> class Example<T> {
>     blaBlaBla = new Box<T>();
> }
> ```

Not support yet.

HerrCai0907:
This is not a good implement, it introduces new AST node, but still cannot cover all cases. I need to think about it again.

Issue #2795, opened 2023-11-14
missing diagnose for interface with initialized field


HerrCai0907:
### Bug description

ASC does not diagnose for these code which will diagnose `An interface property cannot have an initializer.` in TS
```ts
interface NS {
  v:string = ""
}

class T implements NS {
  v:string = "";
}

new T();
```

### Steps to reproduce

`npm run build`

### AssemblyScript version

v0.27.x

PR #2797, opened 2023-11-14
fix: add diagnose when initializing field in interface

HerrCai0907:
Fixes: #2795

### DIFF
--- a/tests/compiler/interface-with-initializer.ts
+++ b/tests/compiler/interface-with-initializer.ts
@@ -1,0 +1,9 @@
+interface I {
+  v:string = ""
+}
+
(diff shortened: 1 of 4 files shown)


tests/compiler/interface-with-initializer.ts
```diff
@@ -0,0 +1,9 @@
+interface I {
+  v:string = ""
+}
+
+class C implements I {
+  v:string = "";
```
CountBleck, line 6:
```suggestion
  v: string = "";
}

class C implements I {
  v: string = "";
```

CountBleck (approved):
LGTM

PR #2800, opened 2023-11-15
fix: diagnose when accessing setter only property

HerrCai0907:
This will introduce some incompatible with TS. since TS is support this but report undefined error in runtime.

### DIFF
--- a/src/resolver.ts
+++ b/src/resolver.ts
@@ -1363,7 +1363,15 @@ export class Resolver extends DiagnosticEmitter {
       }
       case ElementKind.Property: { // someInstance.prop
         let propertyInstance = <Property>target;
-        let getterInstance = assert(propertyInstance.getterInstance); // must have a getter
+        let getterInstance = propertyInstance.getterInstance;
+        if (!getterInstance) {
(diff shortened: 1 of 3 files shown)


src/resolver.ts
```diff
@@ -1363,7 +1363,17 @@ export class Resolver extends DiagnosticEmitter {
       }
       case ElementKind.Property: { // someInstance.prop
         let propertyInstance = <Property>target;
-        let getterInstance = assert(propertyInstance.getterInstance); // must have a getter
+        let getterInstance = propertyInstance.getterInstance;
+        if (!getterInstance) {
+          // It should compile as (undefined) for compatibility with ts.
+          // Since AS don't have undefined, diagnose it in compile time would be better although it does cause compatibility issues
```
CountBleck, line 1369:
```suggestion
          // TS allows this behavior, but at runtime, gets return `undefined`. Since
          // AS doesn't have undefined, we instead diagnose it at compile time, but
          // this isn't compatible with TS.
```

CountBleck (approved):
Seems reasonable.

PR #2805, opened 2023-11-21
Forbid duplicate namespace/static class members

### LINKED ISSUE #2793: asc crash when use same name in class and namespace

CountBleck:
(Barely) fixes #2793.

⯈ Check diagnostics after initialize (VERY BAD)
⯈ Forbid duplicate namespace/static class members

Requiring the compiler to check diagnostics after initialize is VERY BAD, and ideally someone should find a way to prevent the compiler from attempting to compile the duplicate global (which it shouldn't see to begin with).


HerrCai0907 (commented):
I wonder why do we need to split lots of test.

HerrCai0907:
It will reject this valid code
```ts
class A {
  static a: i32 = 0;
}

class B extends A {
  static a: i32 = 0;
}
```

HerrCai0907:
I think merge namespace and class will not be a good idea. But I need to think about how to handle this case.

HerrCai0907:
Cannot detect this case

namespace A {
  export function foo(): i32 {
    return 1;
  }
}

class A {
  static foo(): i32 {
    return 2;
  }
}

  trace("", 1, A.foo());
}

CountBleck:
> Cannot detect this case

That's not good at all.

PR #2813, opened 2024-01-15
fix: clear garbage bits before cast short integer to float

### LINKED ISSUE #2812: The results of arithmetic operations with u8 are not truncated to 8 bytes, when assigned to f64
### Bug description

```ts
export function add(a: u8): f64 {
  let x:u8 = 128
  x = x + a
  return x;
}
```
This function compiles to
```wat
(module
 (type $0 (func (param i32) (result f64)))
 (memory $0 0)
 (export "add" (func $assembly/index/add))
 (export "memory" (memory $0))
 (func $assembly/index/add (param $0 i32) (result f64)
  local.get $0
  i32.const 128
  i32.add
  ;; i32.const 255   missing cast to u8
  ;; i32.and             
  f64.convert_i32_u
 )
)

```

### Steps to reproduce

```ts
export function add(a: u8): f64 {
  let x:u8 = 128
  x = x + a
  return x;
}
```
From js
```js
add(129) // actual 257; expected 1
```

### AssemblyScript version

0.27.22

HerrCai0907:
Fixes #2812.


src/compiler.ts
```diff
@@ -3648,42 +3648,28 @@ export class Compiler extends DiagnosticEmitter {
     // int to float
     } else if (fromType.isIntegerValue && toType.isFloatValue) {
 
+      if (fromType.isShortIntegerValue) expr = this.ensureSmallIntegerWrap(expr, fromType); // must clear garbage bits
```
CountBleck, line 3651:
```suggestion
      // Clear extra bits.
      expr = this.ensureSmallIntegerWrap(expr, fromType);
```

CountBleck (approved):

src/compiler.ts
HerrCai0907: Why we need to remove condition here?

CountBleck: `ensureSmallIntegerWrap` does the check for you.

CountBleck (approved):

PR #2817, opened 2024-01-29
fix: asinit support --noColors opinion

### LINKED ISSUE #2816: npx asinit .
### Bug description

OS : Windows 10
Terminal : cmd

When executing npx asinit .
everything is good except the colors.
Some part of the the text on the console is white. This does not work if the user has a white background.
white color on white background is not good for user experience.

The text color should not be changed so it should be better for everyone.

### Steps to reproduce

1) Open a custom cmd terminal with white background.
2) Run the command npx asinit .

image

### AssemblyScript version

v0.27.23

HerrCai0907:
Detection terminal background technically is possible, but I don't want to introduce deps to resolve UI/UX issue.
like `asc` did, this PR provides `--noColors` opinion in `asinit` to disable colorful output .

Fixes #2816.

### DIFF
--- a/bin/asinit.js
+++ b/bin/asinit.js
@@ -55,10 +55,19 @@ const asinitOptions = {
     ],
     "type": "b",
     "alias": "y"
-  }
+  },
+  "noColors": {
+    "description": "Disables terminal colors.",
+    "type": "b",
+    "default": false
+  },
 };
 
 const cliOptions = optionsUtil.parse(process.argv.slice(2), asinitOptions);
+
+if (cliOptions.options.noColors) {
+  stdoutColors.enabled = false;
+}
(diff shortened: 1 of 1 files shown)


CountBleck (approved):
I wonder, for #2816, if we could also change the color from white to something else (gray maybe).

HerrCai0907:
> I wonder, for #2816, if we could also change the color from white to something else (gray maybe).

Some terminal only support basic 8 colors, including black, red, green, yellow, blue, magenta, cyan, while.
We should avoid to use black, while, blue since some terminal will use those three color as background.
Red, green, yellow should be used for error, warning, success.
Then we still have cyan and magenta to select.

PR #2821, opened 2024-03-01
fix: select correct override function

### LINKED ISSUE #2819: Incorrect overriden function call
### Bug description

Incorrect function will be called if a derived class does not have a function override.

### Steps to reproduce

Paste this code into the AssemblyScript playground window at www.assemblyscript.org

```
export function fib(n: i32): i32 {
  return x.f(n);
}

class A {
  f(a:i32):i32 { return a+1; }
}
class B extends A {
  f(a:i32):i32 { return super.f(a)+10; }
}
class C extends B {
  f(a:i32):i32 { return super.f(a)+100; }
}
class D extends C {
}

let x:A = new D();

```

expected output:
```
fib(0) = 111
fib(1) = 112
fib(2) = 113
...

```

actual output:
```
fib(0) = 11
fib(1) = 12
fib(2) = 13
...

```

Note: It will work correctly, if you change the variable definition to this:

`let x:D = new D();`

or this:

`let x:A = new C();`

### AssemblyScript version

v0.27.24

HerrCai0907:
Fixes #2819.

⯈
⯈
⯈


HerrCai0907:
It is weird that I cannot update this PR. Is there any issue from github side?

PR #2831, opened 2024-03-25
Do not allocate MemorySegments for memory.data(i32, i32?)

### LINKED ISSUE #2827: as crash when try to alloc large data
### Bug description

AS compiler crash after hang up a few second.

### Steps to reproduce

```ts
memory.data(0x7fffffff);
```

`npm run asbuild:debug`

### AssemblyScript version

v0.27.x

CountBleck:
Fixes #2827.

⯈ Bump `Compiler#memoryOffset` instead of allocating zero-filled segments.

There is an issue with the `initialPages` calculation in `initDefaultMemory`, however. The existing condition to run that calculation depends on `memorySegments.length`, which is now zero if only `memory.data(n: i32)` is called, as in the memory-config-errors test.
Making that calculation occur regardless of the number of memory segments would cause a separate issue: due to the default memory offsets of `options.memoryBase`, `1024`, or `8`, modules that previously compiled with an initial memory of 0 pages would compile with a minimum of 1 page.


src/compiler.ts
```diff
@@ -765,6 +765,7 @@ export class Compiler extends DiagnosticEmitter {
     let maximumPages = Module.UNLIMITED_MEMORY;
     let isSharedMemory = false;
 
+    // FIXME: Make the memory-config-errors test not fail
```
CountBleck, line 768: @dcodeIO any idea on what to do about this?

dcodeIO:
Would this perhaps become simpler if we'd keep tracking such segments as `MemorySegment`s but allowed the buffer to be `null`, indicating all zeroes?

More generally, when not indicating the presence of the segment to Binaryen, we'd always have to assume that memory is zeroed, as the zeroes won't be in the `(memory ...)`. Might well be a non-issue in practice, though.

CountBleck:
> Would this perhaps become simpler[...]?

Oh, that's...a much better idea. Thanks and sorry about that.

CountBleck:
> Might well be a non-issue in practice, though.
@dcodeIO hmm, would this conflict when `--importMemory` is specified without `--zeroFilledMemory`?

src/module.ts
```diff
@@ -2409,22 +2409,31 @@ export class Module {
...
+        empty++;
+        continue;
+      }
+
       let offset = segment.offset;
-      unchecked(segs[i] = allocU8Array(buffer));
-      unchecked(psvs[i] = 0); // no passive segments currently
-      unchecked(offs[i] = target == Target.Wasm64
+      segs.push(allocU8Array(buffer));
+      psvs[i - empty] = 0; // no passive segments currently
```
HerrCai0907, line 2428: `psvs` is always 0, then we only need to create a U8Array at the end and set it to zero.

HerrCai0907 (approved):

src/module.ts
CountBleck: I kept it there in case we do something different in the future. The `unchecked(psvs[i] = 0);` line didn't *need* to be there anyway, just as `psvs[i - empty] = 0;` doesn't *need* to be here now.

HerrCai0907 (approved):
LGTM.
A tip: Please avoid to force push after someone has already reviewed PR. This makes it difficult for reviewers to track.

CountBleck:
@HerrCai0907 I changed the commit message; this is technically a breaking change.

HerrCai0907:
> @HerrCai0907 I changed the commit message; this is technically a breaking change.

You can modify it when click squash and merge😆 . Actually the message in commit is meaningless.

PR #2826, opened 2024-03-10
Compile incrementors within the initializer's flow in for loops

CountBleck:
Fixes #2825.

Changes proposed in this pull request:
⯈ Compile for-loop incrementors within the initializer's flow

- [x] I've read the contributing guidelines
- [x] I've added my name and email to the NOTICE file

PR #2834, opened 2024-04-03
fix: ignore this

CountBleck:
This is a dummy commit to create a new release.


HerrCai0907:
Do you know why our release pipeline does not work?

HerrCai0907:
I found the reason, [aspublish](https://github.com/AssemblyScript/aspublish?tab=readme-ov-file#commit-format) requires `breaking / feat / fix` prefix to identifier the version upgrade. Maybe we need to create an action and check the PR title to avoid this weird empty PR. WDYT?

CountBleck:
@HerrCai0907 That would be nice, but if a PR only has one commit, the default commit title in the merge commit is based off of that commit's title, not the PR title (see #2826).

HerrCai0907:
> @HerrCai0907 That would be nice, but if a PR only has one commit, the default commit title in the merge commit is based off of that commit's title, not the PR title (see #2826).

Yes. But we can switch the merge strategy in configuration page. I am not sure who has the access to setup it. Maybe it needs @dcodeIO or @MaxGraey

CountBleck:
@HerrCai0907 that doesn't sound ideal...

HerrCai0907:
Configuration like this:
image

Why is it not ideal?

dcodeIO:
Changed the setting to "Pull request title" as suggested

CountBleck:
> Why is it not ideal?
> 

@HerrCai0907 Oh I misunderstood. That's perfect!

PR #2835, opened 2024-04-03
fix: addDebugInfo before convertExpression

### LINKED ISSUE #2833: Emit incorrect debugLocation while convertExpression()
### Bug description

```typescript
function main(): void {
  add(1, 2);
}
```
For above code snippet, assemblyscript wraps a `drop` for the `call add` instruction. But there is no debuglocation for `call` instruction.

### Steps to reproduce

The root cause is that, in the following code in assemblyscript. When asc wrap a drop epxr for the call epxr by execute convertExpression(), the expr variable(call expr) will be overwritten with (drop expr). And then addDebugLocation(drop).  When returning to `compileStatement` after `compileExpression`, asc will addDebugLocation(drop) again. So we missed the debugLocation for call expr.

In `src/compiler.ts` line 3453-3468,
```typescript
    if (currentType != contextualType.nonNullableType) { // allow assigning non-nullable to nullable
      if (constraints & Constraints.ConvExplicit) {
        expr = this.convertExpression(expr, currentType, contextualType, true, expression);
        this.currentType = currentType = contextualType;
      } else if (constraints & Constraints.ConvImplicit) {
        expr = this.convertExpression(expr, currentType, contextualType, false, expression);
        this.currentType = currentType = contextualType;
      }
    }
    if (wrap) expr = this.ensureSmallIntegerWrap(expr, currentType);
    // debug location is added here so the caller doesn't have to. means: compilation of an expression
    // must go through this function, with the respective per-kind functions not being used directly.
    if (this.options.sourceMap) this.addDebugLocation(expr, expression.range);
```

### AssemblyScript version

0.27.25

XMadrid:
Fixes #2833  .

add debugLocation before convertExpression.
For code snippet: 
```typescript
function add(a: i32, b: i32): i32 {
  return a + b;
}

add(1, 2);
```

Before this PR, we emit: 
```typescript
addDebugLocation(drop, line = 5, column = 2);       // In there, we should emit `call` instruction
addDebugLocation(drop, line = 5, column = 2); 
```

With this PR, we emit:
```typescript
addDebugLocation(call, line = 5, column = 2);      
addDebugLocation(drop, line = 5, column = 2); 
```

### DIFF
--- a/src/compiler.ts
+++ b/src/compiler.ts
@@ -3449,22 +3449,28 @@ export class Compiler extends DiagnosticEmitter {
         expr = this.module.unreachable();
       }
     }
-    // debug location is added here so the caller doesn't have to. means: compilation of an expression
-    // must go through this function, with the respective per-kind functions not being used directly.
-    if (this.options.sourceMap) this.addDebugLocation(expr, expression.range);
     // ensure conversion and wrapping in case the respective function doesn't on its own
     let currentType = this.currentType;
     let wrap = (constraints & Constraints.MustWrap) != 0;
     if (currentType != contextualType.nonNullableType) { // allow assigning non-nullable to nullable
       if (constraints & Constraints.ConvExplicit) {
+        // emit debug location for inner expression before conversion
+        if (this.options.sourceMap) this.addDebugLocation(expr, expression.range);
         expr = this.convertExpression(expr, currentType, contextualType, true, expression);
         this.currentType = currentType = contextualType;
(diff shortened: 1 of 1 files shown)


HerrCai0907 (changes requested):
I don't find any reason to move this debug location from outside expr to inside expr.

XMadrid:
I think we can emit debug location for both outside expr and inside expr.

PR #2837, opened 2024-04-07
chore: remove comment in PR title check pipeline

HerrCai0907:


CountBleck:
@HerrCai0907 Is a comment on the PR even necessary? Maintainers can just edit the PR title as needed, so we can avoid the extra dependency.

HerrCai0907:
> @HerrCai0907 Is a comment on the PR even necessary? Maintainers can just edit the PR title as needed, so we can avoid the extra dependency.

I think it can help developer easy to find what happened and why my PR is rejected.
If you think we should keep less dependence even in CI pipeline, it is fine to remove related logic

HerrCai0907:
I will remove related logic since I cannot find a solution to write comments now. It is weird because in last PR, it can work😕.

CountBleck:
Oh, that's strange...

CountBleck:
@HerrCai0907 I figured it out:
> Note that, if the PR comes from a fork, it will have only read permission despite the permissions given in the action for the pull_request event. In this case, you may use the pull_request_target event. With this event, permissions can be given without issue (the difference is that it will execute the action from the target branch and not from the origin PR).

In #2836, the PR was from a branch on this repo, but this PR is from a branch on your fork.

HerrCai0907:
Let's remove comments simply and avoid permission problem. WDYT?

CountBleck:
That's fine by me. I can't speak for anyone else though.

Issue #2838, opened 2024-04-09
ERROR TS2454: Variable is used before being assigned.


ultraviolet-jordan:
### Bug description

I have updated my package version from `0.27.23` to `0.27.26`.

There is now a compiling issue where:
`ERROR TS2454: Variable 'zj' is used before being assigned.`

### Steps to reproduce

Here is the code block:
```ts
let zn: i32 = gMinlen;
let zvec: i32;
let zj: u8;
for (zvec = getBits(gMinlen, state); zvec > unchecked(gLimit[zn]); zvec = (zvec << 1) | zj) {
    zn++;
    zj = getBit(state);
}
```

The fix is simple like:
```ts
let zn: i32 = gMinlen;
let zvec: i32;
let zj: u8 = 0; // <=== FIX
for (zvec = getBits(gMinlen, state); zvec > unchecked(gLimit[zn]); zvec = (zvec << 1) | zj) {
    zn++;
    zj = getBit(state);
}
```

I do not know if this is intended or not, here is valid Java code for example:
```cpp
int zn = gMinlen;
int zvec;
byte zj;
for (zvec = getBits(gMinlen, s); zvec > gLimit[zn]; zvec = zvec << 1 | zj) {
    zn++;
    zj = getBit(s);
}
```

Just wanted to report just in case. ty

### AssemblyScript version

0.27.26

CountBleck:
This looks like a regression from #2826...sorry!

PR #2839, opened 2024-04-10
fix: for loop statement incrementor will be processed in correct context

HerrCai0907:
Fixes: #2838
incrementor in for loop statement should be processed after executing body.


CountBleck (approved):
LGTM, as long as nothing else breaks

PR #2855, opened 2024-06-20
fix: use `typeof self === object && self.constructor` to identifier execution environment

HerrCai0907:
Use a better way to identifier exection environemt becasue `try` and `catch` will cause misleaded error message

### DIFF
--- a/src/bindings/js.ts
+++ b/src/bindings/js.ts
@@ -976,8 +976,8 @@ export class JSBuilder extends ExportsWalker {
       }
       sb.push(`} = await (async url => instantiate(
   await (async () => {
-    try { return await globalThis.WebAssembly.compileStreaming(globalThis.fetch(url)); }
-    catch { return globalThis.WebAssembly.compile(await (await import("node:fs/promises")).readFile(url)); }
+    if (typeof self === "object" && self.constructor) { return await globalThis.WebAssembly.compileStreaming(globalThis.fetch(url)); }
+    else { return globalThis.WebAssembly.compile(await (await import("node:fs/promises")).readFile(url)); }
   })(), {
 `);
       let needsMaybeDefault = false;
--- a/src/bindings/js.ts
+++ b/src/bindings/js.ts
@@ -976,8 +976,9 @@ export class JSBuilder extends ExportsWalker {
       }
       sb.push(`} = await (async url => instantiate(
   await (async () => {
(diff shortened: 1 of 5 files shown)


src/bindings/js.ts
```diff
@@ -976,8 +976,8 @@ export class JSBuilder extends ExportsWalker {
       }
       sb.push(`} = await (async url => instantiate(
   await (async () => {
-    try { return await globalThis.WebAssembly.compileStreaming(globalThis.fetch(url)); }
-    catch { return globalThis.WebAssembly.compile(await (await import("node:fs/promises")).readFile(url)); }
+    if (typeof self === "object" && self.constructor) { return await globalThis.WebAssembly.compileStreaming(globalThis.fetch(url)); }
```
CountBleck, line 979: I'm wondering if this could go wrong, say, in a Worker on a server-side runtime like Cloudflare Workers, Deno, or Bun.

Do you think there's a single solution that would work with most runtimes and bundlers?

HerrCai0907:
By the way, I think github's macos runner is upgraded to aarch64 architecture. `math.release` test always failed as in my PC.

src/bindings/js.ts
HerrCai0907: I found a repo to do this thing. https://github.com/flexdinesh/browser-or-node/blob/master/src/index.ts
Maybe we can get inspired from it. But I don't have so many test environment.

src/bindings/js.ts
```diff
@@ -976,8 +976,9 @@ export class JSBuilder extends ExportsWalker {
       }
       sb.push(`} = await (async url => instantiate(
   await (async () => {
-    try { return await globalThis.WebAssembly.compileStreaming(globalThis.fetch(url)); }
-    catch { return globalThis.WebAssembly.compile(await (await import("node:fs/promises")).readFile(url)); }
+    const isNodeOrBun = typeof process != "undefined" && process.versions != null && (process.versions.node != null || process.versions.bun != null);
```
CountBleck, line 979:
```suggestion
    const isNodeOrBun = typeof process != "undefined" && process.versions && (process.versions.node || process.versions.bun);
```

At some point in the future, I think this check can be removed, because Node.js in recent versions has `fetch` built-in.

CountBleck (approved):

src/bindings/js.ts
HerrCai0907: agree, let us wait for the new features in nodejs.

PR #2857, opened 2024-06-24
feat: `lowMemoryLimit` can accept more than 1 page value

HerrCai0907:
The original `lowMemoryLimit` only support memory less than 64kB (1 wasm page).
This PR wants to extend the function of `lowMemoryLimit` to support more than 64kB memory limitation.

### DIFF
--- a/cli/options.json
+++ b/cli/options.json
@@ -260,7 +260,7 @@
   },
   "lowMemoryLimit": {
     "category": "Features",
-    "description": "Enforces very low (<64k) memory constraints.",
+    "description": "Enforces memory constraints, memory usage will not exceed the given value.",
     "default": 0,
     "type": "i"
   },
--- a/std/assembly/rt/tlsf.ts
+++ b/std/assembly/rt/tlsf.ts
@@ -381,8 +381,9 @@ function addMemory(root: Root, start: usize, endU64: u64): bool {
   let end = <usize>endU64;
   if (ASC_LOW_MEMORY_LIMIT) {
     end = <u64>ASC_LOW_MEMORY_LIMIT;
+    if (start > ASC_LOW_MEMORY_LIMIT) unreachable();
   }
-  if (<u64>start > endU64) unreachable();
+  if (DEBUG) assert(<u64>start <= endU64); // must be valid
   start = ((start + BLOCK_OVERHEAD + AL_MASK) & ~AL_MASK) - BLOCK_OVERHEAD;
   end &= ~AL_MASK;
 
--- a/std/assembly/rt/tlsf.ts
+++ b/std/assembly/rt/tlsf.ts
@@ -379,6 +379,9 @@ function prepareBlock(root: Root, block: Block, size: usize): void {
 /** Adds more memory to the pool. */
 function addMemory(root: Root, start: usize, endU64: u64): bool {
   let end = <usize>endU64;
+  if (ASC_LOW_MEMORY_LIMIT) {
+    end = <u64>ASC_LOW_MEMORY_LIMIT;
+  }
   if (DEBUG) assert(<u64>start <= endU64); // must be valid
   start = ((start + BLOCK_OVERHEAD + AL_MASK) & ~AL_MASK) - BLOCK_OVERHEAD;
   end &= ~AL_MASK;
@@ -427,10 +430,6 @@ function addMemory(root: Root, start: usize, endU64: u64): bool {
 
 /** Grows memory to fit at least another block of the specified size. */
 function growMemory(root: Root, size: usize): void {
-  if (ASC_LOW_MEMORY_LIMIT) {
-    unreachable();
-    return;
-  }
   // Here, both rounding performed in searchBlock ...
   if (size >= SB_SIZE) {
     size = roundSize(size);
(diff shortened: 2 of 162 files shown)


CountBleck:
What's the use case for this?

HerrCai0907:
In high performance embedded device which has larger RAM than the traditional embedded device but not too much (>1M). In this case, 64kB per page is still to much granularity.
Then it can use `lowMemoryLimit` to control the memory usage.

CountBleck:
Does this do anything that `--maximumMemory` doesn't?

HerrCai0907:
> Does this do anything that `--maximumMemory` doesn't?

maximumMemory is in 64kB size, but lowMemoryLimit is 16B size. But I agree we may can unify them.

CountBleck:
Hmm, I think I'm confused. If you have more than 64 KiB of memory on your device, wouldn't `--maximumMemory` work?

Or are you restricted to an amount of memory that's not a multiple of 64 KiB (for instance, 68 KiB, 123 KiB, etc.)?

HerrCai0907:
> Or are you restricted to an amount of memory that's not a multiple of 64 KiB (for instance, 68 KiB, 123 KiB, etc.)?

Yes. For example, I have device with 150kB memory but want to run 2 wasm job. It is impossible to split 3 x 64kB linear memory.
By this option, each job can use 75kB.

Some background: 
Lots of runtime use virtual memory which will not commit the linear memory from `memory.grow` utill it is really used. So the `memory.grow` will / can success even we don't have so many physical memory.
When WASM want to write some data in linear, it will commit it and make it writable.
But AS's allocator will write information at the end of linear memory, which will make runtime commit the whole linear memory. Then when runtime wants to commit the memory, it will trigger OOM and be killed.

std/assembly/rt/tlsf.ts
```diff
@@ -479,13 +478,7 @@ function initialize(): void {
     }
   }
   let memStart = rootOffset + ROOT_SIZE;
-  if (ASC_LOW_MEMORY_LIMIT) {
```
CountBleck, line 482: Same as the previous comment (keep with `ASC_LOW_MEMORY_LIMIT <= 65535`)

cli/options.json
```diff
@@ -260,7 +260,7 @@
   },
   "lowMemoryLimit": {
     "category": "Features",
-    "description": "Enforces very low (<64k) memory constraints.",
+    "description": "Enforces memory constraints, memory usage will not exceed the given value.",
```
CountBleck, line 263:
```suggestion
    "description": "Enforces memory constraints; memory usage will not exceed the given address.",
```

idk; @dcodeIO wdyt?

CountBleck (commented):
I don't know the intricacies of TLSF and `--lowMemoryLimit`. @dcodeIO is better suited to handle this.

std/assembly/rt/tlsf.ts
```diff
@@ -427,10 +430,6 @@ function addMemory(root: Root, start: usize, endU64: u64): bool {
 
 /** Grows memory to fit at least another block of the specified size. */
 function growMemory(root: Root, size: usize): void {
-  if (ASC_LOW_MEMORY_LIMIT) {
```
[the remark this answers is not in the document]
HerrCai0907: This condition is not needed. The behavior will be the same as before. If `ASC_LOW_MEMORY_LIMIT` less than 65536, grow will failed and trapped.

std/assembly/rt/tlsf.ts
```diff
@@ -379,6 +379,10 @@ function prepareBlock(root: Root, block: Block, size: usize): void {
 /** Adds more memory to the pool. */
 function addMemory(root: Root, start: usize, endU64: u64): bool {
   let end = <usize>endU64;
+  if (ASC_LOW_MEMORY_LIMIT) {
+    end = <u64>ASC_LOW_MEMORY_LIMIT;
+    if (start > ASC_LOW_MEMORY_LIMIT) unreachable();
+  }
```
CountBleck, line 385: @HerrCai0907 would this work? Also, what's the background behind overwriting `endU64`?
```suggestion
  if (ASC_LOW_MEMORY_LIMIT) endU64 = ASC_LOW_MEMORY_LIMIT;
  let end = <usize>endU64;
```

cli/options.json
dcodeIO: Primary challenge here is line length :)

std/assembly/rt/tlsf.ts
HerrCai0907: The original implement is u32, but huge memory will cause u32 overflow and failed the later assert.
I don't want to mix the debug assert (it should not happen) and oom (it may happen).

CountBleck: I don't see how that overflow could happen. Memories larger than 4 GiB only exist when using memory64, and in that case, `usize` would be equivalent to `u64`, not `u32` (so no overflow could take place).

Also, about separating those assertions, you can just change the `if (DEBUG)` on the next line to `if (ASC_LOW_MEMORY_LIMIT || DEBUG)`.

One last thing: **can't `ASC_LOW_MEMORY_LIMIT` be larger than the current memory size in bytes?** An instance could start with a memory of 2 pages, but `ASC_LOW_MEMORY_LIMIT` could be set to 224 KiB (3 and a half pages). In `initialize` and `computeSize`, `addMemory` is called with an `endU64` argument of `memory.size() << 16`, which is the number of bytes available to the instance (if we assume every page has been committed). However, with this change, the value of `end` is changed in `addMemory` from the given value (`memory.size() << 16`) to `ASC_LOW_MEMORY_LIMIT`. While this wouldn't have been a problem before, since previously this was only supposed to work with amounts smaller than 1 page, `ASC_LOW_MEMORY_LIMIT` could be larger than `memory.size() << 16`, and that could lead to `addMemory` adding memory that the instance doesn't have yet (and in turn, cause a crash due to an OOB memory access when someone tries to allocate there).

HerrCai0907: the key point is 4gb, it will overflow to zero

Issue #2870, opened 2024-09-23
Union types of 3 types fail silently


mimi89999:
### Bug description

When using union types of 3 types, asbuild does not report any errors, but the functions are not exported

### Steps to reproduce

1. Define a function accepting a parameter with a union type of 2 types. Notice that asc fails with an error:
```
ERROR AS100: Not implemented: union types
   :
 5 │ export function bar(a: i32 | u32): i32 {
   │                              ~~~
   └─ in assembly/index.ts(5,30)
```
2. Now use a union type of 3 types like in the following code:
```
export function foo(a: aaa | bbb | ccc): i32 {
  return 1;
}

export function bar(a: i32 | u32 | f32): i32 {
  return 1;
}
```
3. Run `npm run asbuild`
4. Notice that the build succeeds, but the functions are not present in the wasm file. When compiling the code I pasted, I got an empty `.wat` file and a minimal `.wasm` file:
```
$ xxd build/release.wasm
00000000: 0061 736d 0100 0000 0503 0100 0007 0a01  .asm............
00000010: 066d 656d 6f72 7902 0000 2410 736f 7572  .memory...$.sour
00000020: 6365 4d61 7070 696e 6755 524c 122e 2f72  ceMappingURL../r
00000030: 656c 6561 7365 2e77 6173 6d2e 6d61 70    elease.wasm.map
````

### AssemblyScript version

0.27.29

PR #2872, opened 2024-09-24
fix: report error for mulitple union type

HerrCai0907:
Fixes: #2870


mimi89999:
Thank you Crimson Knight for the fix!

PR #2879, opened 2024-10-27
chore: update the Type.toString to support output auto

HerrCai0907:


CountBleck:
LGTM

Out of curiosity, what's the use case?

HerrCai0907:
> LGTM
> 
> Out of curiosity, what's the use case?

I think it is much easier when do some debug. sometime it is confused to find a i32 output but it is definitely not i32. after this pr. auto can be print correctly.

maybe it can also change some error in special cases. e.g. if missing type field, compiler will use auto to let subexpr give the type. in theory when subexpr cannot infer the type, old diag will report i32 but new will report auto. But i cannot find this case now.

PR #2880, opened 2024-10-28
feat: support type infer for class memeber

HerrCai0907:
It introduce a new internal mechanism in resolver to infer the expression type without `Flow`. Just like `decltype()` in C++ to support more flexible class definition.


CountBleck:
Is there any existing code that can be used instead of this, such as (perhaps) whatever code is responsible for let​/const​?

Adding a new AST node just for this, and then creating a type from it based off of the NodeKind​ of the expression, makes me a bit uncomfortable, especially since no other type-related AST node contains arbitrary expressions within it (aside from IdentifierExpression​s).

Ultimately, it's your call to make, but I'm wondering if there's a better solution?

HerrCai0907:
> Adding a new AST node just for this, and then creating a type from it based off of the NodeKind​ of the expression

We definitely need a node like `CompiledExpression` in type side to support type infer (not only for this PR, it is only the first application of this kind of node) which has syntax that user does not give the type name but compiler can infer this type.

I think both solution are fine, one is `InferedType` in this PR, the other is a `TypeNode` which contains `Type` directly. Which do you think is better.

HerrCai0907:
@CountBleck WDYT?

CountBleck:
I think I've expressed my dislike for CompiledExpression before. It's most likely better to avoid that sort of thing if possible.

HerrCai0907:
I agree with you. Then the question is should we follow the current design and patch the AST / Resolver / Compiler to do type inferring or we design a new mechanism and introduce new process during compilation. It may introduce a huge change for Class and Function elements.

PR #2881, opened 2024-11-02
chore: fix `tlsfvis.html` label issue

HerrCai0907:
`<li>` cannot be included in `<p>`

see https://html.spec.whatwg.org/multipage/grouping-content.html#the-p-element


CountBleck:
That looks good to me!

PR #2884, opened 2024-11-08
Include runtime from current directory instead of lib

sasq64:
### Changes proposed in this pull request:

The runtime entry points (`index-*.ts`) all include their child modules from `lib` instead
the current directory.

This causes problems in particular when you try to base your custom runtime on the standard runtime, since it will ignore any code and just include from the standard library embedded in the compiler instead.

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -58,6 +58,7 @@ under the licensing terms detailed in LICENSE:
 * Xinquan Xu <xinquan0203@163.com>
 * Matt Johnson-Pint <mattjohnsonpint@gmail.com>
 * Fabián Heredia Montiel <fabianhjr@users.noreply.github.com>
+* Jonas Minnberg <sasq64@gmail.com>
 
 Portions of this software are derived from third-party works licensed under
 the following terms:
--- a/std/assembly/rt/index-incremental.ts
+++ b/std/assembly/rt/index-incremental.ts
@@ -1,2 +1,2 @@
-import "rt/tlsf";
-import "rt/itcms";
+import "./tlsf";
+import "./itcms";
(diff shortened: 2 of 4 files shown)


CountBleck:
Question: if you're basing your runtime off of the standard library, wouldn't you end up copying/vendoring those files into your project?

If so, wouldn't you be able to modify those imports manually?

sasq64:
> Question: if you're basing your runtime off of the standard library, wouldn't you end up copying/vendoring those files into your project?
> 
> If so, wouldn't you be able to modify those imports manually?

Yes, that's exactly what I did, but it took me quite some time to figure out why my changes had no effect.

CountBleck:
Okay, could you merge/rebase the latest changes pushed to main?

HerrCai0907:
I cannot reproduce this issue, could you explain more about it? or give a reproduce repo link. Thanks!

sasq64:
> I cannot reproduce this issue, could you explain more about it? or give a reproduce repo link. Thanks!

Copy `node_modules/assemblyscript/std/assembly/rt` to `rt` and try to use it with
`--runtime rt/index-incremental.ts`

First you notice that you need to fix imports to parent directory (`../util` and `../shared`) but then it seems to work.

However, imports like

```ts
import "rt/tlsf";
import "rt/itcms";
```

actually import from the runtime built into the compiler which makes it ignore any changes you make in your local copy, which is very confusing.

And even if you don't use a custom runtime, I think you should always import relative files with `./` etc and not rely on the lib path?

HerrCai0907:
ahhh, it is real confused to mix library and compiler builtin library. it needs more test cases if we want to use this things in wider scenario.

HerrCai0907 (approved):
lgtm. thanks!

CountBleck:
> ahhh, it is real confused to mix library and compiler builtin library. it needs more test cases if we want to use this things in wider scenario.

Yeah, it looks weird to an outside reader, and it's probably not good style either...

PR #2887, opened 2024-11-14
chore: import node process to skip checking runtime

o-az:
Fixes # .

Deno now supports node process. It requires it to be imported from `node:process` then it should work the same way. This change means it's no longer necessary to check whether the runtime is Deno


CountBleck (changes requested):
It looks like we still need `isDeno`...did you actually test this code with Deno before sending a PR?

PR #2891, opened 2024-12-16
Add support for labeled statements, breaks, and continues

### LINKED ISSUE #2889: Support labeled statements
### Feature suggestion

Labeled statements can come in handy when implementing complex control flow with multiple nested loops and if statements.

Currently parser will complain 
```
function test1():void {
  let str = '';
  loop1: for (let i = 0; i < 5; i++) {
    if (i === 1) {
      continue loop1;
    }
    str = str + i;
  }
}
```

```
ERROR TS1109: Expression expected.
    :
 11 │ loop1: for (let i = 0; i < 5; i++) {
    │      ~
    └─ in assembly/index.ts(11,8)
```

CountBleck:
Fixes #2889.

⯈ Support labeled statements in the parser/AST
⯈ Support labeled `break`/`continue`.


CountBleck:
I just need to add some tests; then, this will be ready to merge.

@JairusSW Heads up: this change definitely breaks transforms. Do you have anything to comment on this PR?

JairusSW:
You can define multiple constructors

CountBleck:
It looks like I forgot about `if`...

CountBleck:
Okay, it appears that my code breaks something related to Flow flags, so more work on this PR is needed.

CountBleck:
@HerrCai0907 I considered that before, and it might be a pretty good idea. It would be better for transform users for sure.

Still, there's the issue of setting the proper `FlowFlags` for a `continue` to an outer loop...maybe the class I use to keep track of labels should contain the flow, and I can use something like `do flow.set(...) while ((flow = flow.parent) != targetFlow)`. (Of course, the actual code would look nicer than that.)

PR #2894, opened 2025-01-12
feat: Constructor generic type inference

### LINKED ISSUE #2893: Constructor type inference
### Feature suggestion

Consider a generic class, such as:

```ts
class Foo<T> {
  constructor(public value: T) {}
}
```

This works:
```ts
const foo = new Foo<string>("abc");
```

But this does not, currently:
```ts
const foo = new Foo("abc");
```

It fails with:
```
ERROR TS2558: Expected 1 type arguments, but got 0.
   :
 5 │ const foo = new Foo("abc");
   │             ~~~~~~~~~~~~~~
   └─ in assembly/index.ts(5,13)
```

I would expect type inference to work on constructors, the same as it presently does for functions.  Consider that this _does_ work:

```ts
function createFoo<T>(value: T): Foo<T> {
  return new Foo(value);
}

const foo = createFoo("abc");
```

And type inference on static methods works too:
```ts
class Foo<T> {
  constructor(public value: T) {}

  static create<T>(value: T): Foo<T> {
    return new Foo(value);
  }
}

const foo = Foo.create("abc");
```

mattjohnsonpint:
Fixes #2893.

Enables type inference for constructors of generic classes by:
- Refactoring to extract the existing type inference logic for functions that was in `maybeInferCall`, into its own private method `inferGenericTypeArguments`.
- Calling that new method from `resolveClassInclTypeArguments` when a generic class's constructor is called without any type arguments.
- Expanding the unit tests to verify working functionality and no regressions.

### DIFF
--- a/src/resolver.ts
+++ b/src/resolver.ts
@@ -3693,11 +3693,17 @@ export class Resolver extends DiagnosticEmitter {
     // Resolve type arguments if generic
     if (prototype.is(CommonFlags.Generic)) {
 
+      // find the constructor prototype, which may be on a base class
+      let constructorPrototype: FunctionPrototype | null = null;
+      for (let p: ClassPrototype | null = prototype; p && !constructorPrototype; p = p.basePrototype) {
+        constructorPrototype = p.constructorPrototype;
+      }
+
       // if no type arguments are provided, try to infer them from the constructor call
-      if (!typeArgumentNodes && prototype.constructorPrototype && flow && ctxTypes.size == 0) {
+      if (!typeArgumentNodes && constructorPrototype && flow && ctxTypes.size == 0) {
         resolvedTypeArguments = this.inferGenericTypeArguments(
           reportNode as NewExpression,
-          prototype.constructorPrototype!,
+          constructorPrototype,
           prototype.typeParameterNodes,
           flow,
         );
--- a/src/resolver.ts
+++ b/src/resolver.ts
@@ -731,100 +731,14 @@ export class Resolver extends DiagnosticEmitter {
 
     // infer generic call if type arguments have been omitted
     if (prototype.is(CommonFlags.Generic)) {
-      let contextualTypeArguments = cloneMap(ctxFlow.contextualTypeArguments);
-
-      // fill up contextual types with auto for each generic component
-      let typeParameterNodes = assert(prototype.typeParameterNodes);
-      let numTypeParameters = typeParameterNodes.length;
-      let typeParameterNames = new Set<string>();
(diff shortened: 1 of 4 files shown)


CountBleck:
@mattjohnsonpint Thanks, that diff looks so much more manageable now.

CountBleck:
Note: it doesn't handle this case:
```ts
class Foo<T> {
    constructor(public x: T) {}
}

class Bar<T> extends Foo<T> {}

const x = new Bar("hi")
```

That can probably be figured out at a later date.

src/resolver.ts
```diff
@@ -3644,15 +3692,28 @@ export class Resolver extends DiagnosticEmitter {
...
-        typeArgumentNodes,
-        flow,
-        ctxElement,
-        ctxTypes, // update
-        reportNode,
-        reportMode
-      );
+
+      // if no type arguments are provided, try to infer them from the constructor call
+      if (!typeArgumentNodes && prototype.constructorPrototype && flow && ctxTypes.size == 0) {
```
CountBleck, line 3697: What's the `ctxTypes.size == 0` for?

CountBleck (approved):
This looks good, probably.

src/resolver.ts
mattjohnsonpint: The else case was the logic that was there previously, and it handles inference for something like `let arr: Array<T> = new Array()` where the inference is from the type it's being assigned to.  Those are in `ctxTypes`.  So if they are present, then I need to _not_ infer from the parameters, but just use the previous logic.

mattjohnsonpint:
> Note: it doesn't handle this case:
> 
> ```ts
> class Foo<T> {
>     constructor(public x: T) {}
> }
> 
> class Bar<T> extends Foo<T> {}
> 
> const x = new Bar("hi")
> ```
> 
> That can probably be figured out at a later date.

I might be able to address that.  I'll take a look tomorrow.  Good catch!

mattjohnsonpint:
It should handle that case now.  Thanks!

mattjohnsonpint:
Also, FWIW, I don't think either of the previously existing test cases marked `TODO` are valid.  They aren't in regular TypeScript anyway, so I don't think they should be in AssemblyScript.

CountBleck (approved):
This seems good.

CountBleck:
> They aren't in regular TypeScript anyway, so I don't think they should be in AssemblyScript.

I don't know what those are for either. I'm guessing it's something related to omitting `: T` and inferring the return type (I think `a` is a placeholder for some value dependent on the generic `T`).

PR #2906, opened 2025-02-05
feat: support 'this' type on methods and properties

### LINKED ISSUE #2904: `this` type should return instance type when overridden
### Bug description

TypeScript allows the `this` keyword to be used as a type.
- https://www.typescriptlang.org/docs/handbook/2/classes.html#this-types
- https://www.typescriptlang.org/docs/handbook/advanced-types.html#polymorphic-this-types

AssemblyScript seems to recognize this syntax and compiles fine:

```ts
class X {
  doSomething(): this {
      return this;
  }
}
```

However, it always returns an object of the type where it's used - which is not necessarily the type of the instance when the object is extended.

```ts
class Y extends X {
  doSomethingElse(): string {
      return "You made it!";
  }
}
```

Found via StackOverflow: https://stackoverflow.com/q/79232050/634824

### Steps to reproduce

With the above code,

```ts
console.log(new Y().doSomething().doSomethingElse());
```

fails to compile.  The error is:

```
ERROR TS2339: Property 'doSomethingElse' does not exist on type 'assembly/index/X'.
    :
 12 │ console.log(new Y().doSomething().doSomethingElse());
    │                                   ~~~~~~~~~~~~~~~
    └─ in assembly/index.ts(12,35)
```

Essentially, `y.doSomething()` is returning a type `X` where it should return a type `Y`.

Casting the value using `<Y>` or `as Y` works, but the point is that one may have full awareness of the original type when making the call.

TypeScript tooling believes the value should be `Y`

Image

### AssemblyScript version

v0.27.32

mattjohnsonpint:
Fixes #2904 

- Implements support for polymorphic `this` type on class methods and properties
- Reports a "not implemented" for _fields_ declared with the `this` type (would be nice, but is complicated)
- Adds compiler tests for this feature


mattjohnsonpint:
FYI, I considered trying to bind the prototype instance to the derived type, but I couldn't quite get that way to work.  So instead, I resolve the current "this" to use as the class instance instead.  It works, but let me know if there's a better way to approach this.  Thanks.

HerrCai0907 (commented):
Could we create an override method for inheritance class 
e.g.
for `class A extends B`, B has method `foo` return `this`, we resolve it as return `B`. A will generate override method `foo` return `A`.
Then resolve function should resolve to correct function.

mattjohnsonpint:
> Could we create an override method for inheritance class ...

I explored this a bit.  It would have side effects, even without using the `this` return type.

For example, given:

```ts
class X {
  private foo: i32 = 0;

  doSomething(): void {
    this.foo = 1;
  }
}

class Y extends X {
}
```

... we can't just transform it to this:

```ts
class X {
  private foo: i32 = 0;

  doSomething(): void {
    this.foo = 1;
  }
}

class Y extends X {
  doSomething(): void {
    this.foo = 1;
  }
}
```

... because field `foo` is declared private.

The [resolver implementation](https://github.com/AssemblyScript/assemblyscript/blob/cdd5e01b9a8bb838ed358e5ded17ac6e85795bc4/src/resolver.ts#L3434-L3460) _does_ effectively create an alias of the member, but in doing so it uses the original prototype, bound to the base instance.  It doesn't create an actual override.

snydergd:
@HerrCai0907 does @mattjohnsonpint make sense in their explanation? Is there anything else to consider? I think it would be great to get support for this feature.

snydergd:
I am wondering if there are still any outstanding concerns with this or if it could be merged.

src/resolver.ts
```diff
@@ -2851,14 +2854,35 @@ export class Resolver extends DiagnosticEmitter {
...
     // Instance method prototypes are pre-bound to their concrete class as their parent
     if (prototype.is(CommonFlags.Instance)) {
-      classInstance = assert(prototype.getBoundClassOrInterface());
+
+      // The actual class instance may be a subclass of the bound class
+      if (this.currentThisExpression && ctxFlow) {
+        // In the case of a function or property that uses the polymorphic `this` type,
+        // this is important, so the return is typed as the actual class instance.
+        // Note: It should work without this outer type check, and does when testing, but fails the "bootstrap" build.
+        // TODO: Figure out why and remove the extra check.
```
HerrCai0907, line 2872: Could you analyze more detail about bootstrap? What is the meaning of "outer type check"?

mattjohnsonpint:
@snydergd - I replied on @HerrCai0907's code review question.  I did mean the entirety of the if condition.  If you just remove the first part, then other tests fail. Keep in mind you have to `npm run build` before re-running any tests or the bootstrap.

PR #2907, opened 2025-02-06
chore: update dependencies

mattjohnsonpint:

- Update minimum Node and NPM versions (Node 16 is EOL)
- Update all dependencies to latest version that build and pass tests without significant code changes
- `npm audit fix` for two ReDoS patches (lockfile only, on `cross-spawn` and `micromatch`)
- Fixed one minor warning about the order of `types` in `package.json`, which was seen after updating `esbuild` to latest
- Updated GitHub Actions to latest versions
- Removed code that was installing npm v6 because the linked bug was fixed some time ago.

I reviewed the change logs for these packages, and I think they should be ok.

I was not able to update `binaryen` without build/test failures.  I'll look into this separately if I can find some time.

I know how to update `eslint` and `typescript-eslint` to latest, but it will be a bit involved and so it should be in a separate PR.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -11,11 +11,11 @@ jobs:
     if: github.repository == 'AssemblyScript/assemblyscript'
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v4
       with:
         ref: main
         fetch-depth: 0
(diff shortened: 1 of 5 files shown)


mattjohnsonpint:
Just noticed the GitHub workflows need updating also.  I'll add that shortly.

HerrCai0907 (approved):
lgtm

MaxGraey (approved):

PR #2908, opened 2025-02-06
fix: use single precision for dtoa with f32

### LINKED ISSUE #2873: f32 toString emits extra decimals
### Bug description

Calling `toString` on an `f64` value works as expected, but doing the same on an `f32` string introduces extra noise decimals.  I suspect it is because the `f32` value is being cast to `f64` when passed into the `dtoa` function:

https://github.com/AssemblyScript/assemblyscript/blob/f79391c91a0875e98a6e887b3353210b4125dc87/std/assembly/number.ts#L317-L319

https://github.com/AssemblyScript/assemblyscript/blob/f79391c91a0875e98a6e887b3353210b4125dc87/std/assembly/util/number.ts#L734

### Steps to reproduce

```ts
console.log((<f64>1.1).toString());  // "1.1"
console.log((<f32>1.1).toString());  // "1.100000023841858"
```

### AssemblyScript version

v0.27.29

mattjohnsonpint:
Fixes #2873

Adjusts `dtoa` and `dtoa_buffered` to accept either `f32` or `f64`.  Implementation details [borrowed from Kotlin](https://github.com/JetBrains/kotlin/blob/320bbb713ffaffcc2ad257af7f4843fa2ad8a3b6/libraries/stdlib/wasm/internal/kotlin/wasm/internal/Number2String.kt#L247).


mattjohnsonpint:
... very strange that I get a fixture mismatch error on Linux but not on macOS.
https://github.com/AssemblyScript/assemblyscript/actions/runs/13185584355/job/36806777984?pr=2908#step:7:7660

investigating...

mattjohnsonpint:
Haha, I just read [this line](https://github.com/JetBrains/kotlin/blob/4870d5214fc2cd8d324037a02e74ad6706bb9da3/libraries/stdlib/wasm/internal/kotlin/wasm/internal/Number2String.kt#L7) in the Kotlin sources for the original implementation:

```
// Based on the AssemblyScript implementation [https://github.com/AssemblyScript/assemblyscript/blob/1e0466ef94fa5cacd0984e4f31a0087de51538a8/std/assembly/util/number.ts]
```

And then it looks like [just last week](https://github.com/JetBrains/kotlin/commit/248135a5ab6377ef048581fa63a84cb2617f0cf5), @skuzmich gave it the 32-bit precision support, for the same reasons.  Seems fair to add it back here too!

(Thanks @skuzmich!)

std/assembly/util/number.ts
```diff
@@ -559,16 +559,26 @@ function getCachedPower(minExp: i32): void {
...
-  let exp = i32((uv & 0x7FF0000000000000) >>> 52);
-  let sid = uv & 0x000FFFFFFFFFFFFF;
-  let frc = (u64(exp != 0) << 52) + sid;
-  exp = select<i32>(exp, 1, exp) - (0x3FF + 52);
+  if (isSinglePrecision) {
+    let uv = reinterpret<u32>(<f32>value);
+    exp = (uv & 0x7F800000) >>> 23;
+    let sid = uv & 0x007FFFFF;
+    frc = (u64(exp != 0) << 23) + sid;
+    exp = select<i32>(exp, 1, exp) - (0x7F + 23);
```
MaxGraey, line 572: We can simplify `select<i32>(exp, 1, exp)` to `exp || 1` nowadays.
```suggestion
    exp = (exp || 1) - (0x7F + 23);
```

std/assembly/util/number.ts
```diff
@@ -559,16 +559,26 @@ function getCachedPower(minExp: i32): void {
...
+  } else {
+    let uv = reinterpret<u64>(value);
+    exp = i32((uv & 0x7FF0000000000000) >>> 52);
+    let sid = uv & 0x000FFFFFFFFFFFFF;
+    frc = (u64(exp != 0) << 52) + sid;
+    exp = select<i32>(exp, 1, exp) - (0x3FF + 52);
```
MaxGraey, line 578: same here

std/assembly/util/number.ts
```diff
@@ -559,16 +559,26 @@ function getCachedPower(minExp: i32): void {
 
 // @ts-ignore: decorator
 @inline
-function grisu2(value: f64, buffer: usize, sign: i32): i32 {
+function grisu2(value: f64, buffer: usize, sign: i32, isSinglePrecision: bool): i32 {
```
MaxGraey, line 562: I prefer to use `isSingle`
```suggestion
function grisu2(value: f64, buffer: usize, sign: i32, isSingle: bool): i32 {
```

std/assembly/util/number.ts
```diff
@@ -716,28 +726,47 @@ function prettify(buffer: usize, length: i32, k: i32): i32 {
...
+  }
+}
+
+export function dtoa32(value: f32): String {
+  return dtoa_impl(value, true);
+}
+
+export function dtoa64(value: f64): String {
+  return dtoa_impl(value, false);
+}
```
MaxGraey, line 759: Not sure if we need these two extra methods while `dtoa` is polymorphic now (`dtoa<T extends number>(value: T)`)

mattjohnsonpint: I thought it would be more optimal, but it's easily removed if you prefer.  Here's a commit on a separate branch with just that change so you can see the WAT differences.
https://github.com/mattjohnsonpint/assemblyscript/commit/55f1789935dc719ef6f5ab24a321f7cfa98a5c7a

LMK which you prefer.  Thanks.

tests/compiler/std/string.ts
```diff
@@ -742,19 +742,17 @@ assert(dtoa(+f64.EPSILON) == "2.220446049250313e-16");
...
-// assert(dtoa(-2.109808898695963e+16) == "-21098088986959630.0"); // FIXME
 assert(dtoa(4.940656E-318) == "4.940656e-318");
 assert(dtoa(9.0608011534336e+15) == "9060801153433600.0");
 assert(dtoa(4.708356024711512e+18) == "4708356024711512000.0");
 assert(dtoa(9.409340012568248e+18) == "9409340012568248000.0");
 assert(dtoa(5e-324) == "5e-324");
-// Known imprecise issue for Grisu alghoritm. Need workaround
-// Expeced: 1.2345e+21
-// Actual:  1.2344999999999999e+21
-// assert(dtoa(1.2345e+21) == "1.2345e+21");
```
MaxGraey, line 757: same here

tests/compiler/std/string.ts
```diff
@@ -778,14 +776,11 @@ assert(dtoa(1e-323) == "1e-323");
 assert(dtoa(-1e-323) == "-1e-323");
 assert(dtoa(1e-324) == "0.0");
 
-assert(dtoa(4294967272) == "4294967272.0");
 assert(dtoa(1.23121456734562345678e-8) == "1.2312145673456234e-8");
-// assert(dtoa(-0.0000010471975511965976) == "-0.0000010471975511965976"); // FIXME
```
MaxGraey, line 783: Also here

std/assembly/util/number.ts
mattjohnsonpint: interestingly the only release wat that changed was the one for this issue.  The others optimized to the same wat.

MaxGraey: I think better to remove these and keep only single dtoa

MaxGraey (approved):
LGTM

tests/compiler/std/string.ts
```diff
@@ -742,19 +742,17 @@ assert(dtoa(+f64.EPSILON) == "2.220446049250313e-16");
...
 assert(dtoa(+f64.MAX_VALUE) == "1.7976931348623157e+308");
 assert(dtoa(-f64.MAX_VALUE) == "-1.7976931348623157e+308");
+assert(dtoa(+f32.EPSILON) == "1.1920929e-7");
+assert(dtoa(-f32.EPSILON) == "-1.1920929e-7");
+assert(dtoa(+f32.MAX_VALUE) == "3.4028235e+38");
+assert(dtoa(-f32.MAX_VALUE) == "-3.4028235e+38");
 assert(dtoa(4.185580496821357e+298) == "4.185580496821357e+298");
 assert(dtoa(2.2250738585072014e-308) == "2.2250738585072014e-308");
-// assert(dtoa(2.98023223876953125e-8) == "2.9802322387695312e-8"); // FIXME
-// assert(dtoa(-2.109808898695963e+16) == "-21098088986959630.0"); // FIXME
```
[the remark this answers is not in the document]
mattjohnsonpint: they are still there, I just moved them to the bottom where I could describe the problem better.  They're all related to the same issue of how the Grisu2 algorithm does rounding.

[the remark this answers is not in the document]
MaxGraey: Yeah, I figured out afterward

CountBleck:
If @MaxGraey says it looks good, we should merge it.

tests/compiler/std/string.ts
```diff
@@ -796,6 +791,34 @@ assert(dtoa(1.1e+128) == "1.1e+128");
 assert(dtoa(1.1e-64) == "1.1e-64");
 assert(dtoa(0.000035689) == "0.000035689");
 
+// The following demonstrates behavior that may seem surprising, but is consistent with the
+// IEEE 754 standard and behaves the same way in other languages that have 32-bit floating point numbers.
+assert(dtoa(<u64>4294967272) == "4294967272.0"); // u64 -> f64, no loss of precision
+assert(dtoa(<u32>4294967272) == "4294967300.0"); // u32 -> f32, loss of precision is expected
```
CountBleck, line 797: I'm not sure I like how `dtoa<u32>(...)` is treated as if an `f32` was passed in...maybe you should do `if (!isFloat<T>()) ERROR("dtoa only supports floats")` or something? Then again, `dtoa` is internal to AS's stdlib...

mattjohnsonpint: I have a commented assert for that reason - I forgot to ask.  I wasn't sure what the best practice is.  Is it too much overhead to put it in?

mattjohnsonpint: Or is it actually better that it casts implicitly?  I mean, `dtoa(123)` worked previously, so if I assert then wouldn't that be a breaking change?

CountBleck: Regarding the assert, you should use `if (!isFloat<T>()) ERROR("some error message")`. `ERROR` is the compile-time equivalent of `assert(false)`.

`dtoa(123)` probably would've worked before by upcasting to an `f64`...hm, I guess we can't just bail on integers...you can call `dtoa` recursively with the appropriate float type then. You'd still have to make sure `T` is a float or integer (`T extends number` doesn't do anything in AS).

CountBleck: Something like:
```ts
if (isInteger<T>()) return sizeof<T>() < 4 ? dtoa<f32>(value) : dtoa<f64>(value);
```

mattjohnsonpint: Ok.  How about this?

```ts
export function dtoa<T extends number>(value: T): String {
  if (isInteger<T>()) {
    return dtoa_impl(<f64>value, false);
  }

  if (isFloat<T>()) {
    if (sizeof<T>() == 4) {
      return dtoa_impl(<f32>value, true);
    } else {
      return dtoa_impl(<f64>value, false);
    }
  }
  
  ERROR("dtoa: unsupported type");
}
```

It actually helps with the existing integer case to avoid surprising results.  I don't think it adds much overhead, since the routine is 64-bit anyway, and there's no strange decimals added like in the f32 case.

mattjohnsonpint: applied that, and similar for `dtoa_buffered`.  Updated the test case.

LMK if you want anything else. 😃

CountBleck: Oh, I might've made a mistake...`dtoa_impl` only accepts `f64` so `ERROR` isn't actually necessary...sorry about that.

```ts
export function dtoa<T extends number>(value: T): string {
    const isSingle = isFloat<T>() && sizeof<T>() == 4;
    return dtoa_impl(value, isSingle);
}
```

This might work?

mattjohnsonpint: That's brilliant!  And the compiler catches it.  Say I try to pass a string, I get:

```
ERROR TS2322: Type '~lib/string/String' is not assignable to type 'f64'.
     :
 746 │ return dtoa_impl(value, isSingle);
     │                  ~~~~~
     └─ in ~lib/util/number.ts(746,20)

FAILURE 1 compile error(s)
Build failed.
```

The location is off, but the error is correct.  And TypeScript editor tooling catches it also.
image

mattjohnsonpint: Applied that.  It reduced some debug wat blocks. 👍

MaxGraey (approved):

PR #2910, opened 2025-02-10
feat: enhance `lowMemoryLimit` to support larger than 64k memory limitation

HerrCai0907:
This PR want `lowMemoryLimit` to accept memory limitation which more than 64k but cannot be divisible by 64k.
It can help generated wasm module with fine-grained memory management.


MaxGraey:
LowMemoryLimit was originally introduced for embedded devices with very limited RAM for which even a single 64k page was not possible

HerrCai0907:
This pr want to let this limitation available for some larger ram devices.
some rumtime like v8 will not allocate memory when memory grow. instead,  they will allocate when memory is used.

for example, some embedden devices have 1MB ram, but still not enough to run multiple wasm together. it is good to use this option to limit ram usage less than e.g. 1.5 wasm pages.

PR #2913, opened 2025-03-24
feat: use second table to visit members

HerrCai0907:
visit members in large projects will generate large switch case, it can be optimized to call_indirect when reference-types enabled.

### DIFF
--- a/src/builtins.ts
+++ b/src/builtins.ts
@@ -99,6 +99,7 @@ import {
 
 import {
   CommonFlags,
+  CommonNames,
   Feature,
   featureToString,
   TypeinfoFlags
@@ -10841,16 +10842,12 @@ function ensureVisitMembersOf(compiler: Compiler, instance: Class): void {
   }
 }
 
-/** Compiles the `__visit_members` function. */
-export function compileVisitMembers(compiler: Compiler): void {
+function compileVisitMembersWithSwitchCase(compiler: Compiler): ExpressionRef {
   let program = compiler.program;
   let module = compiler.module;
   let usizeType = program.options.usizeType;
   let sizeTypeRef = usizeType.toRef();
   let managedClasses = program.managedClasses;
-  let visitInstance = assert(program.visitInstance);
-  compiler.compileFunction(visitInstance, true); // is lazy, make sure it is compiled
-
   // Prepare a mapping of class names to visitor calls. Each name corresponds to
   // the respective sequential (0..N) class id.
   let names = new Array<string>();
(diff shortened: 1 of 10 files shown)


src/builtins.ts
```diff
@@ -10907,16 +10904,67 @@ export function compileVisitMembers(compiler: Compiler): void {
...
+export function compileVisitMembers(compiler: Compiler): void {
+  let program = compiler.program;
+  let module = compiler.module;
+  let usizeType = program.options.usizeType;
+  let sizeTypeRef = usizeType.toRef();
+  let visitInstance = assert(program.visitInstance);
+  compiler.compileFunction(visitInstance, true); // is lazy, make sure it is compiled
+  const current = compiler.options.hasFeature(Feature.ReferenceTypes)
+    ? compileVisitMembersWithCallIndirect(compiler)
+    : compileVisitMembersWithSwitchCase(compiler);
```
MaxGraey, line 10960: What do you think to use `compileVisitMembersWithCallIndirect` only for large amount of members. For example `>= 16` for `optimized for speed` and use `compileVisitMembersWithCallIndirect` without a threshold for optimized for size builds?

HerrCai0907: I think we cannot get any benefit from old implement in runtime which supports new features.

MaxGraey: As far as I know, switch-case is usually faster than dynamic dispatching. Rust even has a whole package for this, https://crates.io/crates/enum_dispatch. Or have you already tried to evaluate the performance, and it turned out to be the same as jump table / switch-case?

MaxGraey: Even in the estimated model binaryen switch is cheaper by a factor of 3. In fact, the difference is much larger.
switch: https://github.com/WebAssembly/binaryen/blob/main/src/ir/cost.h#L68
call_indirect: https://github.com/WebAssembly/binaryen/blob/main/src/ir/cost.h#L80

CountBleck: I definitely want to see performance numbers on this

HerrCai0907: Here is some micro-bench result.

Test Method:
1. build bootstrap debug wasm with and without current feature
2. clean import
```bash
wasm-opt perf/as-ref.wasm --remove-imports -o perf/as-ref.wasm -all
```
3. run script
```js
import { readFileSync } from "fs";
import { performance } from "perf_hooks";
const CNT = 100000
{
  const ins = await WebAssembly.instantiate(readFileSync("perf/assemblyscript.debug-no-import.wasm"));
  ins.instance.exports["_initialize"]();
  const ptr = ins.instance.exports["__new"](100, 0);

  const start = performance.now();
  for (let i = 0; i < CNT; i++) {
    ins.instance.exports["__visit_members"](ptr, 0);
  }
  const end = performance.now();
  console.log(end - start)
}
{
  const ins = await WebAssembly.instantiate(readFileSync("perf/assemblyscript.debug-ref-no-import.wasm"));
  ins.instance.exports["_initialize"]();
  const ptr = ins.instance.exports["__new"](100, 0);

  const start = performance.now();
  for (let i = 0; i < CNT; i++) {
    ins.instance.exports["__visit_members"](ptr, 0);
  }
  const end = performance.now();
  console.log(end - start)
}
```
original  4.517125
PR.         3.3238330000000005
```

Comment (MaxGraey) [reply src/builtins.ts:10959]: > build bootstrap debug wasm with and without current feature

I don't think test perf on debug build is good idea

Comment (HerrCai0907) [reply src/builtins.ts:10959]: > In fact, the same cranelift generates quite impressive code with many checks for call_indirect. You can check it here

Actually it is not so much. br_table also need some bound check code to handle default value logic. And actually, if the table is fixed size. we can save some add operation (I'm not sure why cranelift does not do it)

> Also call_indirect has poor optimization abilities. It can't inline, it produces side effect which poison analysis and code movement during binaryen's optimization. While direct call with switch-case, even huge can be inined (if it has single usage). 

It can be analyzed since this table is only used for this indirect call. From the perspective of information content, call_indirect carries more high-level semantics compared to switch. This means that an optimizer can easily transform call_indirect into a switch-case form, but the reverse is not possible.

> Also cfg can be simplified sometimes.

I think original wasm code with 100+ blocks and br_table is harder to build and analyze CFG.

Comment (HerrCai0907) [reply src/builtins.ts:10959]: When we tall about performance, we should not keep eyes in switch and call_indirect itself.
Here is diff between old and new.

## old
image
```
block  ;; label = @228
    block  ;; label = @229
        block  ;; label = @230
        block  ;; label = @231
            block  ;; label = @232
            local.get 0
            i32.const 8
            i32.sub
            i32.load
            br_table 0 (;@232;) 1 (;@231;) 2 (;@230;) 3 (;@229;) ...... 230 (;@2;) 231 (;@1;)
            end
            return
        end
        return
        end
        return
    end
    local.get 0
    local.get 1
    call $~lib/arraybuffer/ArrayBufferView~visit
    return
end
```
image
image


## new

```
  (func $~lib/rt/__visit_members (type 1) (param i32 i32)
    local.get 0
    local.get 1
    local.get 0
    i32.const 8
    i32.sub
    i32.load
    call_indirect $~/lib/rt/visitor (type 1))
```

Comment (MaxGraey) [reply src/builtins.ts:10959]: I can't reproduce benchmark. I bootstraped compiller with PR (with enabled reference-types feature) and main branch .wasm files. Optimize via wasm-opt as you suggested but when I try run js bench file it can't find `__visit_members`. All builds in debug btw.
```
    ins.instance.exports["__visit_members"](ptr, 0);
                                           ^

TypeError: ins.instance.exports.__visit_members is not a function
```
However rest of exports is present:
```
...
  __new: [Function: 50],
  __pin: [Function: 2389],
  __unpin: [Function: 2390],
  __collect: [Function: 2391],
  __rtti_base: Global [WebAssembly.Global] {},
  memory: Memory [WebAssembly.Memory] {},
  __setArgumentsLength: [Function: 2805],
  _initialize: [Function: 2806],
  setTarget: [Function: 5291],
  setRuntime: [Function: 5292],
  setNoAssert: [Function: 5293],
  setExportMemory: [Function: 5294],
  
  ...
```

Comment (MaxGraey) [reply src/builtins.ts:10959]: > When we tall about performance, we should not keep eyes in switch and call_indirect itself.
Here is diff between old and new.
old

It's typical codegen for any switch statement in wasm. Btw size of bootstapped compiler
```
- with switch:         1,824,783 bytes
- with indirect_call:  1,822,204 bytes
```
So with indirect call we just shrink `~0.14%`

HerrCai0907: Sorry, I forgot one step, in bootstrap wat, manual export __visit_members.

HerrCai0907: > So with indirect call we just shrink ~0.14% for non-optimized build.

It is not a module level change, it is a function level change. which means we save 2.5kB wasm code by optimizing one function.

HerrCai0907:
[perf.zip](https://github.com/user-attachments/files/19457840/perf.zip)

preprocessed wat file for benchmarkting

Issue #2909, opened 2025-02-06
NaN vs -NaN (negative NaN) on macOS/arm64


mattjohnsonpint:
### Bug description

Strange issue found while working on something unrelated.  It seems that in certain cases, the compiler creates `NaN` on some platforms, and `-NaN` on others.  This can lead to test failures.

I encountered it first here:
https://github.com/AssemblyScript/assemblyscript/pull/2908#issuecomment-2640748643

But a search found that we've seen it before.
https://github.com/AssemblyScript/assemblyscript/pull/2352/files/403cae846a804c829b70bd2de388c3286cfdb899?diff=unified&w=0#r931976127

(wait about 5-10 seconds for GitHub to jump to the comment thread on that last link)

I'm not sure if this an macOS issue or an arm64 issue, but in both cases that was the platform.  It does not occur on Linux or Windows x64.

Writing this up mostly so it doesn't get lost.  I haven't found any practical implications yet.  Some opinions about `NaN` vs `-NaN` seems to think it's not very important: https://stackoverflow.com/questions/21349847/positive-vs-negative-nans

Perhaps already fixed by a newer Binaryen implementation? 🤷 

### Steps to reproduce

On a macOS/arm64 machine:

```
npm run test:compiler -- std/math --create
git diff
```

Observe a bunch of changes like this:

```diff
-  f64.const -nan:0x8000000000000
+  f64.const nan:0x8000000000000
```

### AssemblyScript version

v0.27.32

CountBleck:
I can reproduce this on my Android phone via Termux (I knew I've experienced this issue before!).
It's definitely an Arm thing, not macOS.

HerrCai0907:
Some hint, it happened during binaryen's optimization pass. It may cause by nodejs wasm / js engine.

MaxGraey:
Should be fixed with latest Binaryen update

PR #2917, opened 2025-05-13
fix: remove unused tmp local in array literal generated code

HerrCai0907:
`tempDataStart` is never used later.


CountBleck:
`std/math` strikes again?

HerrCai0907:
> `std/math` (#2909) strikes again?

I am not sure, it cannot be reproduced even in linux env

CountBleck (approved):

PR #2922, opened 2025-05-25
fix: use Object.setPrototypeOf in JS bindings

### LINKED ISSUE #2659: Compiling esm esm cannot be used directly in the bundler module
### Bug description

Compiling esm code cannot be used directly in the bundler module，.e.g:
image
The Object.assign cant work. Because bundle module is not the same as browser esm.module export variable just only have getter.
image
image

### Steps to reproduce

1. use binding `@external("../test","test")`
2. complie esm code
image
3. I hava a Compatibility method,but i think Not good enough.
image

### AssemblyScript version

^0.27.0

CountBleck:
Fixes #2659.

⯈ Use `Object.setPrototypeOf({ /* instrumented imports */ }, __moduleN)` instead of `Object.assign(Object.create(__moduleN), { /* instrumented imports */ })`

### DIFF
--- a/src/bindings/js.ts
+++ b/src/bindings/js.ts
@@ -554,18 +554,10 @@ export class JSBuilder extends ExportsWalker {
         continue;
       }
       let resetPos = sb.length;
-      sb.push(": Object.assign(Object.create(");
-      if (moduleName == "env") {
-        sb.push("globalThis");
-      } else {
-        sb.push("__module");
-        sb.push(moduleId.toString());
-      }
-      sb.push("), ");
-      if (moduleName == "env") {
-        sb.push("imports.env || {}, ");
-      }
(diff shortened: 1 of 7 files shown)


HerrCai0907 (commented):
Code looks good. But i am not familiar with JavaScript object API. please make sure it works.

HerrCai0907 (approved):

CountBleck:
> Code looks good. But i am not familiar with JavaScript object API. please make sure it works.

The issue was that using `Object.assign` assigned every value to the new object, but since the object's prototype (the argument to `Object.create`) can have readonly properties (like in module objects from Webpack), this can throw an error. If we assign the properties first and set the prototype afterward, we don't get any errors thrown, and we can override methods even if they were readonly on the prototype.

PR #2924, opened 2025-05-28
fix: builtin_call_indirect does not manage GC obj correctly

HerrCai0907:
In AS calling conversion, arguments should be managed by callee in shadowstack.
builtin_call_indirect does not follow this call conversion.


CountBleck (approved):
LGTM

Note: I did a bit of digging, and it looks like `Compiler` has a `operandsToStack` method called by `makeCallDirect` and `makeCallIndirect`. Sadly, we can't use this method (even if we make it public) since it uses `Signature` and references `thisType` which isn't exactly applicable here.

PR #2927, opened 2025-05-30
chore: update Binaryen

CountBleck:
⯈ Updating Binaryen to v123


CountBleck:
This is a draft because I need to update binaryen.js's types so that the code typechecks correctly.
Also, some of the changes to `*.release.wat` should be looked over...

tests/compiler/function-call.release.wat
```diff
@@ -57,14 +57,15 @@
...
+  nop
  )
  (func $start:function-call~anonymous|2 (param $0 i32) (param $1 i32) (result i32)
   local.get $0
   local.get $1
   i32.add
  )
  (func $start:function-call~fn2|4 (param $0 i32) (result i32)
-  i32.const 1
+  local.get $0
```
MaxGraey, line 68: Hmm, this definitely got worse.

CountBleck: I believe these two cases are caused by the same issue (the `dae`/DeadArgumentElimination pass not doing its job? I'm not fully sure).

I did a bisection to find the Binaryen commit that's responsible. (I tested on the `call-optional` WAT file in S-expr format, but the issue applies here as well.) It looks like WebAssembly/binaryen#7135 is the cause, as that PR treats any funcref as being possibly leaked to the outside world...

How should we proceed? I don't think marking the module as closed-world is accurate...

tests/compiler/function-call.release.wat
```diff
@@ -1408,9 +1409,12 @@
   local.get $0
  )
  (func $function-call/Foo#fnVoid (param $0 i32)
+  nop
  )
  (func $function-call/Foo#fnRet (param $0 i32) (param $1 i32) (param $2 i32) (result i32)
-  i32.const 3
```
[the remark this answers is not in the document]
CountBleck: See above.

tests/compiler/function-call.release.wat
CountBleck: @MaxGraey Should we just eat the cost and merge this PR? We could also mark the module as closed world if functions don't return funcrefs and tables aren't imported/exported...what do you think?

MaxGraey: > We could also mark the module as closed world if exported functions don't return funcrefs and tables aren't imported/exported...what do you think?

That would be ideal because it's a quite common scenario.

CountBleck: @MaxGraey I made that change now!

(Now that I think about it, AS currently never exports arrow functions/etc. as funcrefs, but as boxed objects instead...regardless, the change accounts for table imports/exports as well as users specifying `ref_func` exports, or exported functions returning `ref_func`.)

MaxGraey (approved):
LGTM

PR #2933, opened 2025-07-10
fix: remove non-zero check before __visit

HerrCai0907:
In the implementation of __visit, we will do non-zero check, so in visitMemberOf, we don't need to emit non-zero check.
Emit them both maybe have some performance benefit (I don't do bench for it), but at least for -Oz target, we should remove it.

```ts
export function __visit(ptr: usize, cookie: i32): void {
  if (!ptr) return;
...
```


HerrCai0907:
I am improving the runtime performance and are reviewing generated code line by line. 😅

PR #2936, opened 2025-07-15
fix: create 2 kinds of constructor for instantiation and instantiation of inherited types

HerrCai0907:
create 2 kinds constructor can help binaryen to optimize code better. since binaryen cannot optimize conditional allocation in base class. It will be obvious for `~lib/object/Object` because it is the base class for all kinds of class.

### DIFF
--- a/src/common.ts
+++ b/src/common.ts
@@ -166,6 +166,7 @@ export namespace CommonNames {
   export const this_ = "this";
   export const super_ = "super";
   export const constructor = "constructor";
+  export const raw_constructor = "raw_constructor";
   // constants
   export const ASC_TARGET = "ASC_TARGET";
   export const ASC_RUNTIME = "ASC_RUNTIME";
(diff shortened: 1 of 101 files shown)


HerrCai0907:
similar but more complex constructor in C++: https://discourse.llvm.org/t/why-two-variants-of-every-constructor-are-created-by-the-compiler/73255

HerrCai0907:
The raw constructor will not allocate memory, it is used for inherited class's constuctor.

CountBleck:
I don't think I can say much besides that it looks really strange...is there a way to have the normal constructor just call the "raw" constructor, instead of creating two distinct types of constructors?

HerrCai0907:
It introduce more complexity when `@inline` is set. Since inlined constructor should inline to the caller site. It lets inline meaningless if we just call "raw_constructor" in normal constructor.

PR #2939, opened 2025-08-15
types(String): add [key: i32]: string to type definition

pebrianz:


CountBleck (approved):
LGTM. I don't see how this could go wrong...

PR #2941, opened 2025-08-23
feat: Map support constructor initialization with MapInitialEntries

### LINKED ISSUE #2940: Allow Map to be initialized with an iterable
### Feature suggestion

Right now, to create a map and fill it with data, we have to do it in two steps: first create an empty map, then add each item one by one with `.set()`.

This is what our code currently looks like:

```
let myMap = new Map();
myMap.set('a', 1);
myMap.set('b', 2);
```

This works, but it can get a bit long and clunky, especially for a lot of data.

My suggestion is to allow the `Map` constructor to accept an array of key-value pairs. This would let us create and populate a map in a single, clean line of code.

Here's an example of what that would look like:

```
let myMap = new Map([
  ['a', 1],
  ['b', 2]
]);
```

This change would make our code much cleaner and easier to read, and it's a common pattern in other languages like JavaScript and typescript

pebrianz:
- Added new type alias `MapInitialEntries<K, V>` = { key: K, value: V }[]
- Map constructor now accepts: 
  • An array of { key, value } objects 
  • Any array typed as MapInitialEntries<K,V>
- This allows populating a Map in one step with strong typing.
- Example:
```
  let myMap = new Map<string, i32>([ 
    { key: "a", value: 1 }, 
    { key: "b", value: 2 }, 
  ]);

  let init: MapInitialEntries<string, i32> = [ 
    { key: "x", value: 10 }, 
    { key: "y", value: 20 }, 
  ]; 
  let anotherMap = new Map<string, i32>(init);
```

Fixes #2940 

### DIFF
--- a/std/assembly/map.ts
+++ b/std/assembly/map.ts
@@ -74,39 +74,40 @@ export class Map<K,V> {
   private entriesCount: i32 = 0;
 
   constructor(initialEntries: MapInitialEntries<K,V> | null = null) {
-    if (initialEntries) { 
-      if (initialEntries.length >= this.entriesCapacity) this.bucketsMask = initialEntries.length;
-      this.rehash((this.bucketsMask << 1) | 1);
+    if (!initialEntries || !initialEntries.length) return this;
+    let initialEntriesLength = initialEntries.length;
 
-      for (let i = 0; i < initialEntries.length; i++) {
-        let key = initialEntries[i].key;
-        let value = initialEntries[i].value;
-        let hashCode = HASH<K>(key);
-        let entry = this.find(key, hashCode); // unmanaged!
-        if (entry) {
-          entry.value = value;
-          if (isManaged<V>()) {
-            __link(changetype<usize>(this), changetype<usize>(value), true);
-          }
-        } else {
-          // append new entry
-          let entries = this.entries;
-          let entry = changetype<MapEntry<K,V>>(changetype<usize>(entries) + <usize>(this.entriesOffset++) * ENTRY_SIZE<K,V>());
-          // link with the map
-          entry.key = key;
-          if (isManaged<K>()) {
-            __link(changetype<usize>(this), changetype<usize>(key), true);
-          }
-          entry.value = value;
-          if (isManaged<V>()) {
-            __link(changetype<usize>(this), changetype<usize>(value), true);
-          }
-          ++this.entriesCount;
-          // link with previous entry in bucket
-          let bucketPtrBase = changetype<usize>(this.buckets) + <usize>(hashCode & this.bucketsMask) * BUCKET_SIZE;
-          entry.taggedNext = load<usize>(bucketPtrBase);
-          store<usize>(bucketPtrBase, changetype<usize>(entry));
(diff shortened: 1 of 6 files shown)


std/assembly/index.d.ts
```diff
@@ -2406,6 +2406,14 @@ declare class Map<K,V> {
   toString(): string;
 }
 
+type MapInitialEntries<K,V> = {key: K, value:V}[]
```
CountBleck, line 2409:
```suggestion
type MapInitialEntries<K,V> = {key: K, value: V}[]
```

std/assembly/map.ts
```diff
@@ -66,8 +73,44 @@ export class Map<K,V> {
   private entriesOffset: i32 = 0;
   private entriesCount: i32 = 0;
 
-  constructor() {
-    /* nop */
+  constructor(initialEntries: MapInitialEntries<K,V> = []) {
```
CountBleck, line 76: This should be nullable, and the default argument should be `null` instead of `[]`. Tip: if you wrap the entire constructor body with `if (initialEntries) {`, you can avoid doing `initialEntries!`.

(I wonder if this change could be beneficial for code that never uses this feature, by making it easier for Binaryen to optimize)

CountBleck (commented):
This seems acceptable, since we don't have tuples or iterables/iterators. (If/when we do add support for tuples, we would replace this mechanism as a breaking change.) I don't know if we should be concerned about the size increase in `symbol.release.wat`.

See additional comments below.

std/assembly/map.ts
```diff
@@ -73,14 +73,12 @@ export class Map<K,V> {
   private entriesOffset: i32 = 0;
   private entriesCount: i32 = 0;
 
-  constructor(initialEntries: MapInitialEntries<K,V> = []) {
-    let entriesLength = initialEntries.length;
```
MaxGraey: plz revert this cached entriesLength var and use instead `initialEntries.length`

std/assembly/map.ts
```diff
@@ -66,8 +73,42 @@ export class Map<K,V> {
   private entriesOffset: i32 = 0;
   private entriesCount: i32 = 0;
 
-  constructor() {
-    /* nop */
+  constructor(initialEntries: MapInitialEntries<K,V> | null = null) {
+    if (initialEntries) { 
+      if (initialEntries.length >= this.entriesCapacity) this.bucketsMask = initialEntries.length;
```
MaxGraey, line 78: I think the predicate will be always true here, so just set `this.bucketsMask to initial len`
```suggestion
      this.bucketsMask = initialEntries.length;
```

std/assembly/map.ts
```diff
@@ -66,8 +73,42 @@ export class Map<K,V> {
   private entriesOffset: i32 = 0;
   private entriesCount: i32 = 0;
 
-  constructor() {
-    /* nop */
+  constructor(initialEntries: MapInitialEntries<K,V> | null = null) {
+    if (initialEntries) {
```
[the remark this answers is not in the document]
CountBleck: My bad for suggesting that

PR #2943, opened 2025-08-26
fix: avoid to visit nullptr in binaryen

HerrCai0907:
Module.getExpressionId will be used for nullptr ExpressionRef. It works because deref nullptr in binaryen's wasm is valid.
check BinaryenExprRef non-null before.

### DIFF
--- a/src/compiler.ts
+++ b/src/compiler.ts
@@ -6764,7 +6764,7 @@ export class Compiler extends DiagnosticEmitter {
       if (thisType.isManaged) {
         let operand = operands[0];
         let precomp = module.runExpression(operand, ExpressionRunnerFlags.Default);
-        if (!isConstZero(precomp)) { // otherwise unnecessary
+        if (!precomp || !isConstZero(precomp)) { // otherwise unnecessary
           operands[operandIndex] = module.tostack(operand);
         }
       }
@@ -6778,7 +6778,7 @@ export class Compiler extends DiagnosticEmitter {
       if (paramType.isManaged) {
         let operand = operands[operandIndex];
         let precomp = module.runExpression(operand, ExpressionRunnerFlags.Default);
-        if (!isConstZero(precomp)) { // otherwise unnecessary
+        if (!precomp || !isConstZero(precomp)) { // otherwise unnecessary
           operands[operandIndex] = module.tostack(operand);
         }
       }
--- a/src/module.ts
+++ b/src/module.ts
@@ -3019,7 +3019,6 @@ export function isNullableType(type: TypeRef): bool {
 // expressions
 
 export function getExpressionId(expr: ExpressionRef): ExpressionId {
-  if (expr == 0) return ExpressionId.Invalid;
   return binaryen._BinaryenExpressionGetId(expr);
 }
 


HerrCai0907:
If we replace guard with assert, the compiler will crash during testing.

> ▌ Whoops, the AssemblyScript compiler has crashed during compile :-(
▌ 
▌ 
▌     at Z.assert (/Users/caicongcong/dev/assemblyscript/std/portable/index.js:216:11)
▌     at getExpressionId (/Users/caicongcong/dev/assemblyscript/src/module.ts:3022:3)
▌     at isConstZero (/Users/caicongcong/dev/assemblyscript/src/module.ts:3063:7)
▌     at t.operandsTostack (/Users/caicongcong/dev/assemblyscript/src/compiler.ts:6767:14)
▌     at t.makeCallDirect (/Users/caicongcong/dev/assemblyscript/src/compiler.ts:6910:24)
▌     at t.ensureConstructor (/Users/caicongcong/dev/assemblyscript/src/compiler.ts:8865:18)
▌     at t.compileNewExpression (/Users/caicongcong/dev/assemblyscript/src/compiler.ts:8758:21)
▌     at t.compileExpression (/Users/caicongcong/dev/assemblyscript/src/compiler.ts:3432:21)
▌     at t.compileVariableStatement (/Users/caicongcong/dev/assemblyscript/src/compiler.ts:3011:27)
▌     at t.compileStatement (/Users/caicongcong/dev/assemblyscript/src/compiler.ts:2241:21)
▌     at t.compileStatements (/Users/caicongcong/dev/assemblyscript/src/compiler.ts:2284:23)
▌     at t.compileFunctionBody (/Users/caicongcong/dev/assemblyscript/src/compiler.ts:1733:20)
▌     at t.compileFunction (/Users/caicongcong/dev/assemblyscript/src/compiler.ts:1612:17)
▌     at t.makeCallDirect (/Users/caicongcong/dev/assemblyscript/src/compiler.ts:6841:15)
▌     at t.compileCallDirect (/Users/caicongcong/dev/assemblyscript/src/compiler.ts:6364:17)
▌ 

HerrCai0907:
some background: I am trying to revert the relationship between AS and binaryen. it will create a new c++ program integration binaryen in native and use AS bootstrapped WASM. Then we can reuse the lots of wasm infra in binaryen.

CountBleck:
Regarding that compiler crash, maybe we should find out where a null pointer is being passed, and see if that's intended? If it is intended, this change makes more sense.

Regarding your testing with the AS compiler embedded within a C++ executable...I don't think you can upstream that, but it might be interesting to open-source if possible. It would definitely be cool to see how necessary the JS glue in binaryen.js is to AssemblyScript.

HerrCai0907:
>  it might be interesting to open-source if possible

open source is in progress. it should be finished in this year

> Regarding that compiler crash, maybe we should find out where a null pointer is being passed, and see if that's intended? If it is intended, this change makes more sense.

It generated by `module.runExpression`. I would fix it in caller side.

CountBleck (approved):
LGTM. Looks like everyone else in `compiler.ts`/`flow.ts` does the same thing too!

PR #2945, opened 2025-09-05
feat: optimize string literal tostack behavior

HerrCai0907:
String literals are widely used static data, and our GC algorithm does not recycle them (see https://github.com/AssemblyScript/assemblyscript/blob/7e2a62d9615d2ae02b593af87ee4920a3c57b0bd/std/assembly/rt/itcms.ts#L244). Recording the addresses of all string literals and preventing the emission of tostack code can effectively improve performance

### DIFF
--- a/src/compiler.ts
+++ b/src/compiler.ts
@@ -59,7 +59,8 @@ import {
   isConstNegZero,
   isConstExpressionNaN,
   ensureType,
-  createType
+  createType,
+  getConstValueInteger
 } from "./module";
 
 import {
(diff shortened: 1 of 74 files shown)


HerrCai0907:
In bootstrap parse, the execution time reduced from 210ms  to 200ms.

HerrCai0907:
> My understanding of this PR is that you're trying to add the `if (address < __heap_base)` shortcut to the shadow stack generation code, and `staticGcObjectOffsets` is how you're trying to determine (for `needToStack`) whether some data is static data. Do you think it's possible to instead directly compare the address/offset to the value of `__heap_base`?

That is wrong. e.g. StaticArray<A> also in data segments but we should add it to stack otherwise all of the A in this array will be free.

CountBleck:
Ah, understood.

PR #2948, opened 2025-09-17
feat: optimize the active init operation of zero-initialized memory of `makeFieldInitializationInConstructor`

KKiiim:
- Reduce unnecessary zero-initialization memory in the constructor
- According to wasm-spec, grown memories should initialized with value 0x00


src/compiler.ts
```diff
@@ -10026,6 +10026,95 @@ export class Compiler extends DiagnosticEmitter {
...
 
   /** Makes a constant zero of the specified type. */
+  /** Checks if an expression evaluates to a zero value for the given type. */
+  shouldSkipZeroInit(expr: ExpressionRef, type: Type): bool {
+    let module = this.module;
+    // Try to evaluate the expression at compile time
+    let evaled = module.runExpression(expr, ExpressionRunnerFlags.Default);
+    if (!evaled) return false; // Can't evaluate at compile time
+
+    const evaledType = getExpressionType(evaled);
```
HerrCai0907, line 10036: use `isConstZero`

src/compiler.ts
```diff
@@ -10026,6 +10026,95 @@ export class Compiler extends DiagnosticEmitter {
   // === Specialized code generation ==============================================================
 
   /** Makes a constant zero of the specified type. */
+  /** Checks if an expression evaluates to a zero value for the given type. */
+  shouldSkipZeroInit(expr: ExpressionRef, type: Type): bool {
+    let module = this.module;
+    // Try to evaluate the expression at compile time
+    let evaled = module.runExpression(expr, ExpressionRunnerFlags.Default);
```
HerrCai0907, line 10033: `runExpression` and check result is not expected behavior in debug mode.
consider `x = A - A`. In debug mode, we should emit `(i32.sub (global.get A) (global.get A))` instead of simplify it to `(i32.const 0)` even totally remove it.

src/compiler.ts
```diff
@@ -10407,16 +10497,25 @@ export class Compiler extends DiagnosticEmitter {
...
+              valueExpr
+            ], field.identifierNode, true);
+            if (this.currentType != Type.void) { // in case
+              expr = module.drop(expr);
+            }
+            stmts.push(expr);
+          }
+        } else {
+          // No explicit initializer
+          assert(this.canUseZeroDefault(fieldType));
```
HerrCai0907, line 10517: I think we already check it before. remove this assert.

HerrCai0907 (commented):
optimization goal LGTM. But I think we should only check whether the field is straightly initialized by zero and ignore it.

src/compiler.ts
```diff
@@ -10025,6 +10026,14 @@ export class Compiler extends DiagnosticEmitter {
 
   // === Specialized code generation ==============================================================
 
+  /** Check if possible to optimize the active initialization away if it's zero */
+  canOptimizeZeroInitialization(valueExpr: ExpressionRef): bool {
+    if ((this.options.runtime != Runtime.Incremental) && (this.options.runtime != Runtime.Stub)) {
+      return false;
+    }
+    return isConstZero(valueExpr);
+  }
```
MaxGraey, line 10035:
```suggestion
  canOptimizeZeroInitialization(valueExpr: ExpressionRef): bool {
    const runtime = this.options.runtime;
    return runtime == Runtime.Incremental || runtime == Runtime.Stub
      ? isConstZero(valueExpr)
      : false;
  }
```

src/compiler.ts
```diff
@@ -10407,16 +10417,25 @@ export class Compiler extends DiagnosticEmitter {
...
-        let expr = this.makeCallDirect(setterInstance, [
-          module.local_get(thisLocalIndex, sizeTypeRef),
-          initializerNode // use initializer if present, otherwise initialize with zero
-            ? this.compileExpression(initializerNode, fieldType, Constraints.ConvImplicit)
-            : this.makeZero(fieldType)
-        ], field.identifierNode, true);
-        if (this.currentType != Type.void) { // in case
-          expr = module.drop(expr);
+
+        if (initializerNode) {
```
CountBleck, line 10421: If there isn't an `initializerNode`, this code neglects to zero-initialize the field when stub runtime is used (and the stub runtime doesn't zero-initialize allocated memory!). The previous code's `this.makeZero(fieldType)` happened to cover this case.

CountBleck (commented):
This seems mostly okay, but I'm wondering what all the removed `$~lib/arraybuffer/ArrayBufferView#set:buffer` is about...what is that setter function supposed to be doing anyway?

I'd like @MaxGraey to sign off before merging.

src/compiler.ts
KKiiim: Looks no need to zero-initialize because:
1. memory will not be collected and reused in stub runtime.
2. memory grow will default to initialized with 0 as wasm spec.

KKiiim: Okay, I understand your point.  
Perhaps the safest approach here is to simply disable this optimization in stub rt.

KKiiim:
@MaxGraey @CountBleck Thanks for review again

KKiiim:
Ping for review @MaxGraey @CountBleck @HerrCai0907

PR #2949, opened 2025-09-19
chore: refactor unary postfix code emitter to reduce nested tostack

HerrCai0907:
In original code, `getLocalSetValue` cannot get the original expression accurately. For managed objects, the original expression will be wrapped in a layer of `tostack`, and will be wrapped in another layer of `tostack` when called. Although the subsequent nested `tostack` can be eliminated, this brings some readability issues.

### DIFF
--- a/src/compiler.ts
+++ b/src/compiler.ts
@@ -48,7 +48,6 @@ import {
   getBlockChildCount,
   getBlockChildAt,
   getBlockName,
-  getLocalSetValue,
   getGlobalGetName,
   isGlobalMutable,
   getSideEffects,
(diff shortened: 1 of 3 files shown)


HerrCai0907:
The original behaviour

```ts
class Bar {
  v: i32 = 0;
  @operator.postfix("++")
  static postfix_inc(bar: Bar): string {
    bar.v++;
    return "i++";
  }
}

  let a = new Bar();
  let c = a++;
  trace(`a.v = ${a.v} ${c}`);
}
```

```wasm
(call $assembly/index/Bar.postfix_inc
 (call $~lib/rt/__tostack
  (call $~lib/rt/__tostack
   (local.get $0)
  )
 )
)
```

KKiiim (approved):

PR #2951, opened 2025-09-25
chore: refactor builtin type resolving to `builtins.ts`

HerrCai0907:


CountBleck:
I think moving it there is a good idea too

CountBleck (approved):
LGTM

PR #2955, opened 2025-10-25
chore: friendly initial test in asinit

### LINKED ISSUE #2954: Improve `asinit` tests
### Feature suggestion

This is current initial test code for AS project:
```js
// ./tests/index.js
import assert from "assert";
import { add } from "../build/debug.js";
assert.strictEqual(add(1, 2), 3);
console.log("ok");
```

Since [Node 18](https://nodejs.org/api/cli.html#--test) we have the `--test` flag: `node --test tests/**/*`. I suggest using this:
```js
// ./tests/index.test.js <-- renamed
import assert from "assert";
import { add } from "../build/debug.js";
import { it } from "node:test";

it('add', () => {
    assert.strictEqual(add(1, 2), 3);
})
```

Mopsgamer:
Fixes #2954.

### DIFF
--- a/bin/asinit.js
+++ b/bin/asinit.js
@@ -430,12 +430,12 @@ function ensureTestsIndexJs() {
   console.log("- Making sure that 'test/index.js' exists...");
   if (!fs.existsSync(testsIndexFile)) {
     fs.writeFileSync(testsIndexFile, [
-      "import assert from \"node:assert\";",
+      "import assert from \"node:assert/strict\";",
       "import { it } from \"node:test\";",
       "import { add } from \"../build/debug.js\";",
       "",
       "it(\"add\", () => {",
-      "  assert.strictEqual(add(1, 2), 3);",
+      "  assert.equal(add(1, 2), 3);",
       "});"
     ].join("\n") + "\n");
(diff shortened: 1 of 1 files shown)


bin/asinit.js
```diff
@@ -427,13 +427,16 @@ function ensureTestsDirectory() {
 }
 
 function ensureTestsIndexJs() {
-  console.log("- Making sure that 'tests/index.js' exists...");
+  console.log("- Making sure that 'tests/index.test.js' exists...");
```
CountBleck: Maybe we can have the directory be `test/` instead of `tests/`, so the file path can be `test/index.js`. This would let the script command in `package.json` be `node --test`, because [by](https://nodejs.org/docs/latest-v20.x/api/test.html#test-runner-execution-model) [default](https://nodejs.org/api/test.html#running-tests-from-the-command-line:~:text=**/test/,js}) it includes anything under the `test/` directory.

CountBleck (commented):
I left some thoughts below, but I do want to know what you think of them. On another note, you need to change the PR title prefix to either `chore:` or `feat:`.

PR #2958, opened 2025-11-01
chore: remove baseUrl from std/assembly.json

EDM115:
Fixes a crash of tsgolint when used in a project where AssemblyScript is present.

When #637 was introduced it also brought the `baseUrl` option to enable `paths` to work. However this option will be deprecated in TS 6 and removed in TS 7, so projects that already use TS 7 like [oxc-project/tsgolint](https://github.com/oxc-project/tsgolint) will crash when said option is set in any of the `tsconfig` files in a project (cf oxc-project/tsgolint#351).

According to https://github.com/microsoft/TypeScript/issues/62508#issuecomment-3348649259 it is safe to remove the option, and since the path *already* included it (`./`) it won't break any existing behavior (see also https://github.com/andrewbranch/ts5to6#overview).

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -61,6 +61,7 @@ under the licensing terms detailed in LICENSE:
 * Jonas Minnberg <sasq64@gmail.com>
 * Kam Chehresa <kaz.che@gmail.com>
 * Mopsgamer <79159094+Mopsgamer@users.noreply.github.com>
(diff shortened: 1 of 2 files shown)


HerrCai0907 (approved):
LGTM

EDM115:
that's nice to hear 😄

PR #2967, opened 2025-12-27
fix: check native source correctly

HerrCai0907:
native code's internal path is ~lib/native.ts instead of ~lib.
However, it is more easier to check current source is same as static
_native field.

Fixes # .

⯈
⯈
⯈

- [ ] I've read the contributing guidelines


CountBleck:
Silly me

PR #2971, opened 2025-12-30
resolver: support package exports for root import (".")

eldiablolives:
Closes or improves upon long-standing bare import limitations.
Allows libraries like @whatever/stuff to use standard "exports" mapping.
No breaking changes — full fallback to ascMain/assembly.


eldiablolives:
@HerrCai0907 Thank you for the feedback! I've moved all the package resolution logic to the CLI side only, keeping the core parser unchanged. The module resolving now happens exclusively in index.js during file loading

Issue #2972, opened 2026-01-09
Pre-initialised UTF string constants with `memory.data<>`


geraintluff:
### Feature suggestion

I'd love a syntax like this:

```typescript
let utf8Ptr = memory.data<u8>("Hello");
let utf16Ptr = memory.data<u16>("Hello");
```

At the moment I'm doing this at runtime instead:

```typescript
let bytes = String.UTF8.byteLength(str, true);
let utf8Ptr = heap.alloc(bytes);
String.UTF8.encodeUnsafe(changetype<usize>(str), str.length, utf8Ptr, true);
```

I could use `String.UTF8.encode()` instead, and then store the `ArrayBuffer` so it doesn't get GC'd, but as far as I can tell, that still ends up embedding the UTF-16 string in the output.

PR #2974, opened 2026-01-13
feat: Add `memory.dataUTF8()` built-in function

geraintluff:
Implements #2972, providing a way to embed UTF-8 string constants. 

### Rationale

UTF-16 strings are naturally embedded in the data section if you use a string constant (`"..."`) in your code.

If what you need is a UTF-8 string, then storing as WTF-16 wastes binary space and execution time.

### Current workaround

You can currently achieve the same result by using `memory.data<u8>([..., 0])` with the appropriate UTF-8 bytes, but reading/changing/reviewing this code is more difficult.

### Changes

* Declare `memory.dataUTF8(<string constant>)` function in standard library
* Implement the built-in in the compiler
* Add test to `tests/compiler/memory.ts`

`npm run check` shows no errors for the files I touched.  (Existing errors in `tests/import/index.ts` are unrelated.)


src/builtins.ts
```diff
@@ -118,6 +118,9 @@ import {
   isPowerOf2
 } from "./util";
 
+// Use the built-in `TextEncoder` for UTF-8 conversion
+declare let TextEncoder: any;
```
dcodeIO: Unfortunately, this won't work for two reasons. One is is that there is no `any` type when compiling the compiler itself to Wasm, and the other is that there is no TextEncoder implementation in the standard library.

What we have instead is [this utility](https://github.com/AssemblyScript/assemblyscript/blob/main/std/assembly/string.ts#L663) that can perhaps be reused when combined with a portable fallback.

geraintluff: Nice, thanks - I'll change it to use that

geraintluff:
@dcodeIO I wrote a `utf16ToUtf8()` helper, checked portability using `npm run bootstrap`.  It's a single pass, so allocates the most pessimistic buffer size (1.5x the WTF16 length in bytes).

It silently encodes unpaired surrogates (WTF16) as code-points, which is the same behaviour as `String.UTF8.encodeUnsafe()` under the default `ErrorMode.WTF8`.

src/builtins.ts
```diff
@@ -3491,6 +3492,78 @@ function builtin_memory_data(ctx: BuiltinFunctionContext): ExpressionRef {
...
+    } else {
+      result[utf8Length++] = 0xF0 + ((codePoint>>18)&0x07);
+      result[utf8Length++] = 0x80 + ((codePoint>>12)&0x3F);
+      result[utf8Length++] = 0x80 + ((codePoint>>6)&0x3F);
+      result[utf8Length++] = 0x80 + (codePoint&0x3F);
+    }
+  }
+  let trimmed = new Uint8Array(utf8Length);
+  for (let i: i32 = 0; i < utf8Length; ++i) trimmed[i] = result[i];
+  return trimmed;
```
MaxGraey:
```suggestion
  return result.subarray(0, utf8Length);
```

std/assembly/index.d.ts
```diff
@@ -1798,6 +1798,8 @@ declare namespace memory {
   export function data(size: i32, align?: i32): usize;
   /** Gets a pointer to a pre-initialized static chunk of memory. Alignment defaults to the size of `T`. Arguments must be compile-time constants. */
   export function data<T>(values: T[], align?: i32): usize;
+  /** Gets a pointer to a pre-initialized static chunk of memory containing null-terminated UTF8. Value must be a compile-time constant. */
+  export function dataUTF8(value: string): usize;
```
MaxGraey, line 1802: I think this API could be improved and instead `dataUTF8` use more generalized `text`:
```ts
  enum TextEncoding { UTF8, UTF16 }

  declare namespace memory {
    ...
    export function text(value: string, encoding?: TextEncoding, nullTerminated?: bool): usize;
  }
```

geraintluff: Seems reasonable to me - shall I take a swing at it?

MaxGraey: I still think about the best API. Not sure about nullTerminated. It makes sense asually only for ASCII/UTF8.
```ts
enum TextEncoding { 
  UTF8, // default ?
  UTF8CString, // null terminzated
  UTF16
}

declare namespace memory {
  ...
  export function text(value: string, encoding?: TextEncoding): usize;
}
```

PR #2975, opened 2026-01-14
feat: Let Binaryen encode our memory init more efficiently

geraintluff:
AssemblyScript produces a bunch of small data chunks.  Binaryen can split up large chunks to represent them in a compact way, but not combine them.

This change checks whether Binaryen will be doing any optimisation at all, and combines all the `MemorySegment`s into one big segment.

Here's a comparison of the bootstrapped AssemblyScript compiler itself, with and without the change:

| file | bootstrap size (before) | bootstrap size (with changes) | delta |
| ---- | ---- | ---- | ---- |
| assemblyscript.debug.wasm | 1783641 | 1785335 | +0.10% |
| assemblyscript.debug.wasm.gz | 379084 | 379560 | +0.13% |
| assemblyscript.release.wasm | 918047 | 916063 | -0.22% |
| assemblyscript.release.wasm.gz | 224686 | 223556 | -0.50% |

Since this patch adds functionality, it increases the code size of the debug build, but the release build is still smaller due to the repacking.

It's a small improvement, and on my other projects I'm only seeing a ~1% change but 🤷.

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -62,6 +62,7 @@ under the licensing terms detailed in LICENSE:
 * Kam Chehresa <kaz.che@gmail.com>
 * Mopsgamer <79159094+Mopsgamer@users.noreply.github.com>
 * EDM115 <github@edm115.dev>
+* Geraint Luff <luffgd@gmail.com>
 
 Portions of this software are derived from third-party works licensed under
 the following terms:


geraintluff:
Oh, and obviously it fails a bunch of release-build fixture tests. 😅

HerrCai0907 (commented):
My biggest concern is this feature may OOM the building machine when the wasm is large. I remember there are a PR (maybe not merged) to don't create a real JS buffer during compilation when the data is all-zero to reduce the memory cost.

e.g. `memory.data(large_number)`

geraintluff:
> My biggest concern is this feature may OOM the building machine when the wasm is large.

@HerrCai0907 Excellent point.  I'll convert this back to draft, and then re-open when I have something incremental which avoids this problem.

PR #2994, opened 2026-03-21
fix: evaluate call_indirect callee before arguments

### LINKED ISSUE #2989: Function reference and parameter are resolved in wrong order.
### Bug description

Root cause.
When  `makeCallIndirect`, parameters are executed before function address calculation, it leads to parameters overwrite the value of `fn`.

### Steps to reproduce

Code

```typescirpt

function logA(x: i32):void {
  trace(`A: ${x}`);
}

function logB(x: i32):void {
  trace(`B: ${x}`);
}

let fn = logA;

function foo(x: i32):i32 {
  fn = logB;
  return x;
}

fn(foo(42));
```
Expected behavior
In typescript, the output is `A: 42`
Observed behavior
In assemblyscript, the output is `B: 42`

### AssemblyScript version

v0.28.9

Changqing-JING:
Fixes #2989

Fix wrong order of resolving function expression and operands expression for call_indirect


Changqing-JING:
Can some maintainer help to review this PR?

HerrCai0907:
It makes the semantics of as and ts more consistent, with a slight performance loss, yet I believe this trade-off is worthwhile. @MaxGraey  WDYT?

MaxGraey:
I'm not sure. I think AssemblyScript provides much more obvious and logical approach. JavaScript has historically had a lot of quirks, and I think this is one of them.  Also, it's breaking change

Changqing-JING:
@MaxGraey In my opinion, the assemblyscript need to follow typescript/javascript design
1. It's a javascript spec compliance. ECMAScript [§13.3.6.1](https://tc39.es/ecma262/multipage/ecmascript-language-expressions.html#sec-function-calls-runtime-semantics-evaluation) mandates callee evaluation before arguments. AS claims to be a TS subset — deviating here silently produces wrong results.

2. It's not a quirk, evaluating the callee before arguments is the common convention across major languages, aligning with most programmers' mental model.

   **Python** — guaranteed left-to-right evaluation:
   ```python
   def log_a(x):
       global value
       value = 100

   def log_b(x):
       global value
       value = 200

   value = 0
   fn = log_a

   def foo(x):
       global fn
       fn = log_b
       return x

   fn(foo(42))
   print(f"value = {value}")  # → 100
   ```

   **Java** — same behavior:
   ```java
    import java.util.function.IntConsumer;

   public class EvalOrderTest {
       static int value = 0;
       static IntConsumer fn;

       static void logA(int x) { value = 100; }
       static void logB(int x) { value = 200; }

       static int foo(int x) {
           fn = EvalOrderTest::logB;
           return x;
       }

       public static void main(String[] args) {
           fn = EvalOrderTest::logA;
           fn.accept(foo(42));
           System.out.println("value = " + value);  // → 100
       }
   }
   ```

   Both Python and Java guarantee `value = 100` — AS is the outlier among languages that claim defined semantics.

2. Negligible perf cost. The allTrivial fast path skips the extra temp local when no side effects exist, which is the common case.

3. Not a real breaking change. Code relying on the wrong order is already buggy. Fixing correctness doesn't break correct programs.

MaxGraey:
Oh, now I see. Сallee should be saved before args run. Otherwise, side effects can swap the function under your feet, which is pretty footgun-y.

MaxGraey:
It will be great to add smoke test for compiler which check correct order in output.

Something like:
```ts
function origCallback(x: i32): void {
  trace(`called origCallback with ${x}`);
}

function replacedCallback(x: i32): void {
  trace(`called replacedCallback with ${x}`);
}

let cb = origCallback;

function newCallback(): i32 {
  trace("inside newCallback, replacing cb");
  cb = replacedCallback;
  return 42;
}

trace("before call");
cb(newCallback());
trace("after call");
```

Changqing-JING:
@MaxGraey https://github.com/AssemblyScript/assemblyscript/pull/2994/changes#diff-a823605e23478a27f21c0a82211c9c1614f976458d0cdc68effdeabd2937b6ee
This pr already added similar test case

src/compiler.ts
```diff
@@ -7080,36 +7079,48 @@ export class Compiler extends DiagnosticEmitter {
       }
     }
 
+    let allTrivial = (getSideEffects(functionArg, module.ref) & SideEffects.WritesGlobal) == 0;
+    if(operands && allTrivial) {
```
MaxGraey:
```suggestion
    if (operands && allTrivial) {
```

src/compiler.ts
```diff
@@ -7080,36 +7079,48 @@ export class Compiler extends DiagnosticEmitter {
...
+          allTrivial = false;
+          break;
+        }
+      }
+    }
+
+    let stmts = new Array<ExpressionRef>();
+    let sizeTypeRef = this.options.sizeTypeRef;
+
+    if(!allTrivial){
```
MaxGraey:
```suggestion
    if (!allTrivial){
```

src/compiler.ts
```diff
@@ -7080,36 +7079,48 @@ export class Compiler extends DiagnosticEmitter {
...
+      let functionArgLocal = this.currentFlow.getTempLocal(this.options.usizeType);
+      let functionArgSetExpr = module.local_set(functionArgLocal.index, functionArg, true);
+      stmts.push(functionArgSetExpr);
+      functionArg =  module.local_get(functionArgLocal.index, sizeTypeRef);
```
MaxGraey:
```suggestion
      functionArg = module.local_get(functionArgLocal.index, sizeTypeRef);
```

Changqing-JING: Done

src/compiler.ts
Changqing-JING: Done

src/compiler.ts
Changqing-JING: Done

MaxGraey:
plz sync with main and update fixtutes

src/compiler.ts
```diff
@@ -7080,36 +7079,48 @@ export class Compiler extends DiagnosticEmitter {
...
+          allTrivial = false;
+          break;
+        }
+      }
+    }
+
+    let stmts = new Array<ExpressionRef>();
+    let sizeTypeRef = this.options.sizeTypeRef;
+
+    if (!allTrivial){
```
MaxGraey:
```suggestion
    if (!allTrivial) {
```

HerrCai0907 (approved):
LGTM. I prefer it is a break changing. This implicitly changes the behavior of the code; `break` makes this change more visible.

Changqing-JING:
@MaxGraey @dcodeIO 
What's the follow up step of this PR. It's open here more a long while.

MaxGraey (approved):
LGTM

Just one note: I don't think these changes are frequent enough to be labeled as "breaking changes" It would be better to update the title with "fix:"

MaxGraey:
Also, what about a regular direct call? Is order arguments and the called function are evaluated the same as in the PR?

Changqing-JING:
Direct call doesn't have this problem. Because the target of direct call is a fixed value at compile time. There is no expression compile order problem

MaxGraey:
Thanks!

PR #3007, opened 2026-04-03
fix: report error when generic method attempts to implement non-generic interface method

### LINKED ISSUE #3006: Compiler crashes when a generic method is used to implement a non-generic interface method
### Bug description

The compiler crashes with an AssertionError: assertion failed instead of producing a diagnostic when a class implements an interface but uses a generic method where the interface declares a non-generic one.

### Steps to reproduce

```typescript
interface I {
  foo(x: i32): i32;
}

class CC implements I {
  foo<T>(x: i32): i32 {
    return x;
  }
}

let c:I = new CC();
c.foo(1);
```
run `node ./bin/asc.js ./demo/test.ts`
```
Whoops, the AssemblyScript compiler has crashed during compile :-(
▌
▌
▌     at assert2 (D:\code\workspace\github\assemblyscript\std\portable\index.js:216:11)
▌     at Resolver.resolveFunction (D:\code\workspace\github\assemblyscript\src\resolver.ts:2798:7)
▌     at Resolver.resolveOverrides (D:\code\workspace\gi42)
▌     at Module.compile (D:\code\workspace\github\assemblyscript\src\index-wasm.ts:358:32)
▌     at Module.main (D:\code\workspace\github\assemblyscript\cli\index.js:732:31)
▌     at async file:///D:/code/workspace/github/assemblyscript/bin/asc.js:33:22
▌
▌
▌ Thank you!
```

### AssemblyScript version

v0.28.13

Changqing-JING:
Fixes #3006  .

TypeScript allows this because it uses type erasure — generics disappear at runtime, so `foo<T>()` and `foo()` are the same function. AssemblyScript cannot do the same because it uses monomorphization — each type argument produces a distinct function, and the vtable needs a single fixed entry. This is the same reason C++ forbids template functions from overriding virtual functions.

The fix treats a generic method as not satisfying a non-generic interface method, matching C++ semantics.

### DIFF
--- a/src/resolver.ts
+++ b/src/resolver.ts
@@ -3449,13 +3449,12 @@ export class Resolver extends DiagnosticEmitter {
           // A generic method cannot implement a non-generic interface method
           // because monomorphization requires a concrete type to generate code,
           // but virtual dispatch through the interface has no type arguments.
-          let ifaceMember = unimplemented.get(memberName);
-          if (ifaceMember
+          if (unimplemented.has(memberName)
             && member.kind == ElementKind.FunctionPrototype
-            && ifaceMember.kind == ElementKind.FunctionPrototype
+            && unimplemented.get(memberName)!.kind == ElementKind.FunctionPrototype
           ) {
             let memberTypeParams = (<FunctionPrototype>member).typeParameterNodes;
-            let ifaceTypeParams = (<FunctionPrototype>ifaceMember).typeParameterNodes;
+            let ifaceTypeParams = (<FunctionPrototype>unimplemented.get(memberName)).typeParameterNodes;
             let numMemberTypeParams = memberTypeParams ? memberTypeParams.length : 0;
             let numIfaceTypeParams = ifaceTypeParams ? ifaceTypeParams.length : 0;
             if (numMemberTypeParams != numIfaceTypeParams) continue;
--- a/src/resolver.ts
+++ b/src/resolver.ts
@@ -3063,7 +3063,14 @@ export class Resolver extends DiagnosticEmitter {
           let boundPrototype = classInstance.getMember(unboundOverridePrototype.name);
           if (boundPrototype) { // might have errored earlier and wasn't added
             assert(boundPrototype.kind == ElementKind.FunctionPrototype);
-            overrideInstance = this.resolveFunction(<FunctionPrototype>boundPrototype, instance.typeArguments);
+            let boundFuncPrototype = <FunctionPrototype>boundPrototype;
+            let overrideTypeParams = boundFuncPrototype.typeParameterNodes;
+            let numOverrideTypeParams = overrideTypeParams ? overrideTypeParams.length : 0;
+            let baseTypeArgs = instance.typeArguments;
+            let numBaseTypeArgs = baseTypeArgs ? baseTypeArgs.length : 0;
+            if (numOverrideTypeParams == numBaseTypeArgs) {
+              overrideInstance = this.resolveFunction(boundFuncPrototype, baseTypeArgs);
+            }
           }
         }
         if (overrideInstance) overrides.add(overrideInstance);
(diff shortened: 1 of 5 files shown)


tests/compiler/override-typeparam-mismatch.ts
```diff
@@ -0,0 +1,14 @@
+interface I {
+  foo(x: i32): i32;
+}
+
+class CC implements I {
+  foo<T>(x: i32): i32 {
+    return x;
+  }
+}
```
MaxGraey, line 9: Can you also add this test case?
```ts
interface I2 {
  foo<T, U>(x: i32): i32;
}

class C2 implements I2 {
  foo<T>(x: i32): i32 {
    return x;
  }
}
```

Changqing-JING: Added this testcase in new commit

MaxGraey (approved):
LGTM

MaxGraey:
The only thing that could be improved is to add a new error specific to AS. Such errors:

```
“TS2515: Non-abstract class ‘override-typeparam-mismatch/CC’ does not implement inherited 
abstract member ‘foo’ from ‘override-typeparam-mismatch/I’.”,
```

 isn’t very informative, since the method is actually defined and has the correct signature. The only issue is incorrect parametric polymorphism.
 
 Perhaps may sense to add new [AS241 error](https://github.com/AssemblyScript/assemblyscript/blob/main/src/diagnosticMessages.json#L54) for this

src/resolver.ts
```diff
@@ -3063,7 +3063,14 @@ export class Resolver extends DiagnosticEmitter {
...
             assert(boundPrototype.kind == ElementKind.FunctionPrototype);
-            overrideInstance = this.resolveFunction(<FunctionPrototype>boundPrototype, instance.typeArguments);
+            let boundFuncPrototype = <FunctionPrototype>boundPrototype;
+            let overrideTypeParams = boundFuncPrototype.typeParameterNodes;
+            let numOverrideTypeParams = overrideTypeParams ? overrideTypeParams.length : 0;
+            let baseTypeArgs = instance.typeArguments;
+            let numBaseTypeArgs = baseTypeArgs ? baseTypeArgs.length : 0;
+            if (numOverrideTypeParams == numBaseTypeArgs) {
+              overrideInstance = this.resolveFunction(boundFuncPrototype, baseTypeArgs);
+            }
```
HerrCai0907: should we allow this implement when the generic number are the same? I think it will crash also if the typing mismatched.
IMO, maybe we should forbidden all generic function in interface?

MaxGraey: Hmm, I see only one edge case in this scenario:
```ts
interface I {
  add<T, U>(x: T, y: U): U;
}

class C2 implements I {
  add<U, T>(x: T, y: U): U {
    return x;
  }
}
```

Changqing-JING: Good point. If we follow C++ design:
1. template function aka. generic function can't be member in interface
2. template function also can't override virtual function in interface.

Shall we do the same?

I would prefer to follow C++ design because generic function in interface is a confused behaviour 
But it may break some existing code on user side.
What do maintainers prefer?

Changqing-JING:
> The only thing that could be improved is to add a new error specific to AS. Such errors:
> 
> ```
> “TS2515: Non-abstract class ‘override-typeparam-mismatch/CC’ does not implement inherited 
> abstract member ‘foo’ from ‘override-typeparam-mismatch/I’.”,
> ```
> 
> isn’t very informative, since the method is actually defined and has the correct signature. The only issue is incorrect parametric polymorphism.
> 
> Perhaps may sense to add new [AS241 error](https://github.com/AssemblyScript/assemblyscript/blob/main/src/diagnosticMessages.json#L54) for this

There I just referred the hehavior of clang. In clang, a generic(template) function isn't regarded as overriding of virtual interface function, so the interface function is regarded as "unimplemented"

src/resolver.ts
Changqing-JING: @HerrCai0907  @MaxGraey 

Dear maintainers,

I have pushed a new commit to forbidden interface function and abstract function to be generic.
Could you check this design concept.

By the way, I just awared that if base class and child class one is generic, the other one is non-generic, then there is another crash. But I would prefer to fix it in another PR. I just left a fix me in current PR.
```typescript
var called: string = "";

class A {
  foo<T>(x: T): void {
    called = "A";
  }
}

class B extends A {
  foo(x: i32): void { // non-generic overrides generic
    called = "B";
  }
}

// Static dispatch: B.foo called directly
var b = new B();
b.foo(1);
assert(called == "B");
called = "";

// Virtual dispatch through A reference: which gets called?
var a: A = new B();
a.foo<i32>(1);
assert(called == "B"); // if vtable works, "B"; if fallback to base, "A"
```

HerrCai0907 (approved):
LGTM

HerrCai0907 (commented):
should we mark it as break changing?

Changqing-JING:
I think it doesn't need to be marked as break change, because calling generic function interface is not assemblyscript designed to be. Even if it worked in some cases, it's just working by accident. 

I would regard generic function in interface as a bug which isn't detected in the past, not a feature break change.

MaxGraey (approved):

Changqing-JING:
@HerrCai0907 @MaxGraey 
Thank you for review! Can this PR be merged? I need to fix another bug base on this PR

Issue #20, opened 2018-01-30
Custom transformers


TomMarius:
Is AssemblyScript planning to support TypeScript-like custom transformers?

dcodeIO:
If I understand this correctly, custom transformers do not look AOT-compatible. More like something that requires the language to be highly dynamic.

dcodeIO:
On a second thought, something similar might be possible if there was an API to plug in custom built-ins that have full access to the compiler's type information and transform such code at compile time.

dcodeIO:
See also: https://github.com/AssemblyScript/assemblyscript/wiki/Limitations

TomMarius:
Custom transformer is a function plugged into the compilation process. It takes the AST of the program and returns a new (or the same) AST. It has nothing to do with anything dynamic; other statically typed languages have this sort of plugin functionality as well.

dcodeIO:
I see, thanks. AssemblyScript has an [AST representation](https://github.com/AssemblyScript/assemblyscript/blob/master/src/ast.ts) as well, so transforming it prior to the final compilation step shouldn't be a big deal. There is also the [Program that wraps the higher level IR](https://github.com/AssemblyScript/assemblyscript/blob/master/src/program.ts) created from the AST. That's where [built-in functionality](https://github.com/AssemblyScript/assemblyscript/blob/master/src/builtins.ts) is combined with the processed source information, which is the last step before compilation is finally performed.

dcodeIO:
Regarding the commit above, there is now a way to provide transforms with the `--transform` option to `asc`. This basically loads a node module that then exports hooks the compiler will call on specific occasions. The only hook for now is `afterParse` which calls the transform with the parser instance that contains the parsed sources, prior to instantiating and initializing a program from it. It's relatively easy to add more hooks, of course, but I didn't want to go overboard with that so let me know if there's something specific you'd need.

dcodeIO:
Closing this for now as the infrastructure is in place. Just let me know if more hooks are necessary :)

PR #3015, opened 2026-04-15
feat: Parse parameter decorators

PaperPrototype:
I noticed Josh Tenner, did a final attempt to fix [his PR](https://github.com/AssemblyScript/assemblyscript/pull/3001)... I am not sure if it was actually correct or not but I am making a new PR based on that one. Mainly wanting to try and see if I can do my own review on these changes and introduce myself into contributing to the AS repository for the coming years. The final goal of all this is for me to write an AST [transform](https://www.assemblyscript.org/compiler.html#transforms) for outputting WGSL code which `@decorators` on parameters will be useful for.

> [!NOTE]
> I am not requesting a PR review or expecting this to be merged any time soon (unless the maintainer's want to do that). Making this PR is helpful for me to keep up to date with changes on main branch and seeing the live diff.

## Changes I would like to eventually arrive at in this pull request:
Allow checking for `@decorators` on parameters and return types using a helper (with least modifications to source code of AssemblyScript). So that a transform plugin can access them and output WGSL code with the same `@decorators`. 

Something akin to this:

(AssemblyScript)
```ts
function addNumbers(a: f32, b: f32): f32 {
    return a + b;
}

// Vertex Shader
@wgsl.vertex
function vertexShader(@wgsl.builtin("vertex_index") vid: u32): @wgsl.builtin("position") vec4f {
    let pos: StaticArray<vec2f> = [
      vec2f( 0.0,  0.5),
      vec2f(-0.5, -0.5),
      vec2f( 0.5, -0.5)
    ];

    return vec4f(pos[vid], 0.0, 1.0);
}

// Fragment Shader
@wgsl.fragment
function fragmentShader(): @wgsl.location(0) vec4f {
    return vec4f(0.0, 1.0, 0.0, 1.0); // Green
}
```

(WGSL)
```wgsl
fn addNumbers(a: f32, b: f32) -> f32 {
    return a + b;
}

// Vertex Shader
@vertex
fn vertex_shader(@builtin(vertex_index) vid: u32) -> @builtin(position) vec4f {
    let pos = array(
        vec2f( 0.0,  0.5),
        vec2f(-0.5, -0.5),
        vec2f( 0.5, -0.5)
    );
    return vec4f(pos[vid], 0.0, 1.0);
}

// Fragment Shader
@fragment
fn fragment_shader() -> @location(0) vec4f {
    return vec4f(0.0, 1.0, 0.0, 1.0); // Green
}
```

Afaik this code allows reading `@decorators` from parameters using a helper function. I'll be doing my own research into this and trying to acquaint myself with AssemblyScript internals. I mainly wanted the PR so I can keep it up to date with main and see a live diff of what is different from main.

NOTE: I don't use Codex or ClaudeCode, nor am I interested in using them (I don't even have them installed). I do make use of Claude chat (claude.ai/new) quite often though for research and smaller pieces of code of which I often heavily modify, as well as occasional Copilot inline editing provided by VS Code.

Related Issues:
https://github.com/AssemblyScript/assemblyscript/issues/2497
https://github.com/AssemblyScript/assemblyscript/issues/20

Testing commands:
```
npm run build
npm run test:parser -- parameter-decorators
npm run test:compiler -- parameter-decorators --noDiff
npm run test:transform
```


src/ast.ts
```diff
@@ -2399,6 +2406,383 @@ export function findDecorator(kind: DecoratorKind, decorators: DecoratorNode[] |
   return null;
 }
 
+/** Generic AST walker. */
+export abstract class NodeWalker {
```
PaperPrototype: Why is there a whole new AST node walker...

MaxGraey: Looks like you decided to pick up the baton. Not sure why `NodeWalker` is needed here. Feels like this could’ve been done without it. 

If you want to keep working on this, the first thing to do is try to understand the internal architecture of at least the AssemblyScript parser. Claude can probably explain it better. Then try giving it a task to review the implementation in this PR and rewrite everything (except the tests), sticking to the existing code style and paradigms, with as little “creative liberty” as possible, basically minimize the diff as much as you can.

Also, I recently came across an interesting [technical prompt for Claude by Andrej Karpathy](https://github.com/forrestchang/andrej-karpathy-skills/blob/main/CLAUDE.md) I haven’t tried it yet. LLMs are pretty much useless for the kind of tasks I’m dealing with right now, but it seems like a decent starting point. Of course, ideally, you’d want to figure everything out yourself.

dcodeIO: I think the NodeWalker here - if needed at all - can be delegated into the transform itself, with only the bits and pieces added to the AS parser and compiler that are strictly necessary to support parameter decorators. By default, if no transform consumes these decorators, the compiler would reject them as unsupported as it does with other decorators. That's not the feature though. The feature is that now a transform *can* consume them.

src/ast.ts
```diff
@@ -2399,6 +2406,383 @@ export function findDecorator(kind: DecoratorKind, decorators: DecoratorNode[] |
...
+
+  /** Indicates whether walking has been stopped. */
+  stopped: bool = false;
```
MaxGraey: The problem is that this property is never set to true during the tree traversal. It is always `false`. No one outside is set it either. I just don’t understand what the point of all this is. The comments don’t offer any clear explanation either. It looks like some sort of AI artifact / hallucination. NodeWalker is just used as base Visitor-like class for ParameterDecoratorValidator which by itself looks like a highly questionable decision

src/ast.ts
PaperPrototype: This is definitely going to take me some time (especially because of how new I am). Also do you recommend I install claude code, or just paste the relevant files into a claude chat and analyze from there feeding it more context as it needs? I am somewhat unfamiliar with claude code and i've heard that it will modify a lot of stuff you may not want it to.

EDIT: this command worked well for getting the entire diff and pasting it into Claude `git diff main...HEAD | pbcopy`, it also copies the diff to your clipboard on MacOS

PaperPrototype: Main goal right now is to just understand everything.

PaperPrototype:
Result of running command `npm run test:parser -- parameter-decorators` is:

```sh
> assemblyscript@0.0.0 test:parser
> node --enable-source-maps tests/parser parameter-decorators

Testing parser/parameter-decorators.ts
diff OK

[ SUCCESS ]
```

Result of running command `npm run test:compiler -- parameter-decorators --noDiff` is:
```sh
> assemblyscript@0.0.0 test:compiler
> node --enable-source-maps --no-warnings tests/compiler parameter-decorators --noDiff

# compiler/parameter-decorators

- compile debug
  ERROR TS1206: Decorators are not valid here.
     :
   1 │ function regular(@first value: i32): void {}
     │                  ~~~~~~
     └─ in parameter-decorators.ts(1,18)

  ERROR TS1206: Decorators are not valid here.
     :
   2 │ function rest(@rest ...values: i32[]): void {}
     │               ~~~~~
     └─ in parameter-decorators.ts(2,15)

  ERROR TS1206: Decorators are not valid here.
     :
   3 │ function withthis(@self this: i32, value: i32): i32 { return this; }
     │                   ~~~~~
     └─ in parameter-decorators.ts(3,19)

  ERROR TS1206: Decorators are not valid here.
     :
   6 │ constructor(@field public value: i32) {}
     │             ~~~~~~
     └─ in parameter-decorators.ts(6,15)

  ERROR TS1206: Decorators are not valid here.
     :
   7 │ method(@arg value: i32): void {}
     │        ~~~~
     └─ in parameter-decorators.ts(7,10)

  ERROR TS1206: Decorators are not valid here.
      :
   10 │ type Callback = (@arg value: i32) => void;
      │                  ~~~~
      └─ in parameter-decorators.ts(10,18)

  ERROR TS1206: Decorators are not valid here.
      :
   11 │ const expression = function(@arg value: i32): void {};
      │                             ~~~~
      └─ in parameter-decorators.ts(11,29)

  ERROR TS1206: Decorators are not valid here.
      :
   12 │ const arrow = (@arg value: i32): void => {};
      │                ~~~~
      └─ in parameter-decorators.ts(12,16)

  ERROR TS1206: Decorators are not valid here.
      :
   14 │ export function nested(@arg value: i32): void {}
      │                        ~~~~
      └─ in parameter-decorators.ts(14,26)

  ERROR AS102: User-defined: "EOF"
      :
   17 │ ERROR("EOF");
      │ ~~~~~~~~~~~~
      └─ in parameter-decorators.ts(17,1)

  FAILURE 10 compile error(s)
  ---
  10 compile error(s)
  ---
  SKIPPED (104.258 ms)

- compare stderr
  SUCCESS (0.195 ms)

Time: 115 ms

[ SUCCESS ]
```

Result of running `npm run test:transform` is:
```sh
> assemblyscript@0.0.0 test:transform
> npm run test:transform:esm && npm run test:transform:cjs


> assemblyscript@0.0.0 test:transform:esm
> node bin/asc tests/compiler/empty --transform ./tests/transform/index.js --noEmit && node bin/asc tests/compiler/empty --transform ./tests/transform/simple.js --noEmit && node bin/asc tests/transform/parameter-decorators.ts --transform ./tests/transform/remove-parameter-decorators.js --noEmit

Transform loaded
- constructor
- afterParse
  defer
  resolve
  defer
  resolve
  defer
  resolve
  complete
- afterInitialize
  defer
  resolve
- afterCompile
Simple transform loaded
- afterParse
- afterInitialize
- afterCompile
Parameter decorator removal transform loaded
- afterInitialize strip parameter decorators

> assemblyscript@0.0.0 test:transform:cjs
> node bin/asc tests/compiler/empty --transform ./tests/transform/cjs/index.js --noEmit && node bin/asc tests/compiler/empty --transform ./tests/transform/cjs/simple.js --noEmit && node bin/asc tests/transform/parameter-decorators.ts --transform ./tests/transform/cjs/remove-parameter-decorators.js --noEmit

CommonJS transform loaded
- constructor
- afterParse
- afterInitialize
- afterCompile
Simple CommonJS transform loaded
- afterParse
- afterInitialize
- afterCompile
CommonJS parameter decorator removal transform loaded
- afterInitialize strip parameter decorators
```

src/ast.ts
PaperPrototype: > I think the NodeWalker here - if needed at all - can be delegated into the transform itself, with only the bits and pieces added to the AS parser and compiler that are strictly necessary to support parameter decorators. By default, if no transform consumes these decorators, the compiler would reject them as unsupported as it does with other decorators. That's not the feature though. The feature is that now a transform _can_ consume them.

It would be ideal, at least in my mind, if your code could remain portable such that a math function written with `@wgsl` decorators could still be imported and used by AS.

PaperPrototype:
Okay so after messing about I think I understand what needs done.

Default behaviour of decorators in AS is to ignore them if they are invalid. For example I can do:

```ts
@blabla
@interesting("")
@wgsl.vertex("main")
function initBoard(): void { }
```

... and there is no error.

The problem is that doing the following:

```ts
@blabla
@interesting("")
@wgsl.vertex("main")
function initBoard(@first hello: i32): @location(0) i32 {
  return 0;
}
```

...will throw compile errors:

```sh
ERROR TS1003: Identifier expected.
    :
 45 │ function initBoard(@first hello: i32): @location(0) i32 {
    │                   ~
    └─ in sketch.ts(45,19)

FAILURE 1 parse error(s)
ERROR TS1110: Type expected.
    :
 45 │ function initBoard(hello: i32): @location(0) i32 {
    │                                 ~
    └─ in sketch.ts(45,33)

FAILURE 1 parse error(s)
```

So we only need (as dcode put it) "the bits and pieces added to the AS parser and compiler that are strictly necessary to support parameter decorators." Yet somehow this PR seems to be doing a LOT more than that.

PaperPrototype:
1. Do we want an error to be thrown when a transform does not consume the parameter decorators? eg. `function f(@a x: i32): void {}` -> `TS1206: Decorators are not valid here.` but still parse successfully so transforms can remove them. Or...

2. Can parameter decorators simply be ignored by the compiler like other decorators?

I prefer 2, the latter case. @MaxGraey @dcodeIO 

Also, after research it seems the only changes needed are in ast.ts and parser.ts (and possibly src/extra/ast.ts). Does this sound correct?

MaxGraey:
`ERROR TSXXXX` errors relate to TypeScript's compiler and semantics. We can parse parameter decorators as we like and still error our specific `ERROR ASXXX` errors for invalid decorators which not pass our requirements.

Also, we can add suppression flag for compiler to completely ignore decorator errors. But in this case transformer should validate it by itself

PaperPrototype:
@MaxGraey I was hoping for some clarification on something you mentioned
> ERROR TSXXXX errors relate to TypeScript's compiler and semantics. We can parse parameter decorators as we like and still error our specific ERROR ASXXX errors for invalid decorators which not pass our requirements.

Current behavour I have implemented outputs these types of errors:
```ts
  ERROR TS1206: Decorators are not valid here.
     :
   1 │ function regular(@first value: i32): void {}
     │                  ~~~~~~
     └─ in parameter-decorators.ts(1,18)

  ERROR TS1206: Decorators are not valid here.
     :
   2 │ function rest(@rest ...values: i32[]): void {}
     │               ~~~~~
     └─ in parameter-decorators.ts(2,15)
```

Should I implement a custom ASXXX error code or continue to use TS1206? If we want custom error codes then how should I go about doing that?

Here is the test commands if you want to check the output for yourself:
```
npm run test:parser -- parameter-decorators
npm run test:compiler -- parameter-decorators --noDiff
npm run test:transform
npm run build
```

MaxGraey:
If there's a suitable error in TS, it's best to use it but TS transpiler is fairly conservative and generally only throws errors that are still in tc39/proposals. And I don't recall what the status is there for parameter decorators.

src/compiler.ts
```diff
@@ -10626,6 +10626,37 @@ export class Compiler extends DiagnosticEmitter {
...
+          if (decorators && decorators.length > 0) {
+            this.error(
+              DiagnosticCode.Decorators_are_not_valid_here,
+              Range.join(decorators[0].range, decorators[decorators.length - 1].range)
+            );
+          }
+        }
+      }
+    }
+  }
```
MaxGraey: I don’t like that transformer validation happens in the compiler. If it’s not valid for upstream AS semantics, it’s better to move it to transform utils or add it to their pipeline automatically.

TS will not merge decorators soon, and even if it does, parameters are still unclear:
[https://github.com/tc39/proposal-decorators/blob/master/EXTENSIONS.md#parameter-decorators-and-annotations](https://github.com/tc39/proposal-decorators/blob/master/EXTENSIONS.md#parameter-decorators-and-annotations)

If param decorator semantics are not standardized there either, it feels wrong to validate this inside our compiler.

MaxGraey: Actually, I’d just pass the bare minimum of what needs to be fed into the transformer, and let the transformer itself parse, canon and validate the decorators via the AST. Especially considering that this feature is pretty niche in terms of general use.

PaperPrototype: To be honest, I have no idea what I am doing 😅

Good to know about the decorators proposal for TS, I wasn't aware of it before 👍

tests/transform/cjs/remove-parameter-decorators.js
```diff
@@ -0,0 +1,329 @@
...
+  TypeDeclaration: 60,
+  VariableDeclaration: 61
+};
+
+const LiteralKind = {
+  Template: 3,
+  Array: 5,
+  Object: 6
+};
+
```
MaxGraey: I'd use the TS version of the transformer to handle all of this

src/compiler.ts
PaperPrototype: What do you mean by "add it to their pipeline automatically"?

MaxGraey: I mean extend transform's API by adding new validation entry point [here](https://github.com/AssemblyScript/assemblyscript/blob/f93a9415bcde40660305a32414f2cb2ccfccbfc4/cli/index.d.ts#L273C1-L284C2) between listFiles and afterParse and add validation hook to this entry point instead validate by itself.  This would allow users to validate the code in the transformer themselves before compilation but after parsing

MaxGraey: Or don't do this and just use afterParse hook only.

Simply put, in this case we have ambiguity. Should we fail because we can’t handle such decorators, or not fail and hope that the Transformer will take on this role? An additional validator interface's method for Transformer API would answer this question clearly.

PaperPrototype: > Actually, I’d just pass the bare minimum of what needs to be fed into the transformer, and let the transformer itself parse, canon and validate the decorators via the AST. Especially considering that this feature is pretty niche in terms of general use.

Sorry, I needed to reload the page before I saw this. When you say "Transformer" what are you referring to? I'm assuming you mean the implementors who will want to make use of these decorators? Eg. the example transformers in /tests/transform? Still very new to all the verbiage being thrown around here, sorry about the extreme noobiness

PaperPrototype: Oh unless you are referring to the AssemblyScript AST transformer...

MaxGraey: > I needed to reload the page before I saw this. When you say "Transformer" what are you referring to? I'm assuming you mean the implementors who will want to make use of these decorators?

yes

tests/transform/cjs/remove-parameter-decorators.js
PaperPrototype: TS version of the transformer?

src/compiler.ts
PaperPrototype: > Or don't do this and just use afterParse hook only.
> 
> Simply put, in this case we have ambiguity. Should we fail because we can’t handle such decorators, or not fail and hope that the Transformer will take on this role? An additional validator interface's method for Transformer API would answer this question clearly.

I think I finally get what you meant. I made some notes:

- AST can be extended without breaking changes, this is perfectly fine and anything that is added will just be ignored by the compiler but at least transformer plugins can then make use of it. No need to validate and throw errors unless it is an AST parse error.
    - Need to add a transformer hook so transformers can add their own validation. This way AssemblyScript can delegate validation to transformers instead of having to handle it internally, which doesn't make sense because AS shouldn't be in charge of worring about that in the first place.
    
Is this correct? Just a "👍" is fine for me, unless I am completely wrong 🤣

MaxGraey: Yeah that's pretty accurate

tests/transform/cjs/remove-parameter-decorators.js
PaperPrototype: Oh I see, write in in typescript so i can import these from assemblyscript directly, yep

MaxGraey: yes

PaperPrototype:
@MaxGraey @dcodeIO I think... everything is working?

This compiles with no errors:
```ts
function regular(@first value: i32): void {}
function multiParam(@first huh: string, @second(123, 456) works: i32, @third("can pass a string?") hello: string): void {}
function rest(@rest ...values: i32[]): void {}
function withthis(@self this: i32, value: i32): i32 { return this; }

class Box {
  constructor(@field public value: i32) {}
  method(@arg value: i32): void {}
}

type Callback = (@arg value: i32) => void;
const expression = function(@arg value: i32): void {};
const arrow = (@arg value: i32): void => {};
namespace ns {
  export function nested(@arg value: i32): void {}
}
```

I also added a `beforeCompile` hook for transformers to be able to validate decorators if they want to per Max's suggestion. Though maybe that would be better done in a separate PR? What are your thoughts?

I'm happy to continue if it's not where it should be 🙏

MaxGraey:
Sync this branch with main and try to rebuild fixtures

PaperPrototype:
Hmm, I uh, based my fork on JTenner, I'm unsure how to sync with assemblyscript/assemblyscript main 😅 

Screenshot 2026-05-04 at 7 31 27 PM

PaperPrototype:
I'll figure it out

PaperPrototype:
Closed because I am afraid of force pushing, see new PR here: #3023

PR #3016, opened 2026-04-22
feat: multi-value support (and tuples too i suppose)

### LINKED ISSUE #1633: Multi-value support?
The WebAssembly [multi-value proposal](https://github.com/WebAssembly/multi-value) has already been apart of the Wasm [finished-proposals](https://github.com/WebAssembly/proposals/blob/master/finished-proposals.md) for some time, yet I can't find any mention of how to use it in ASC, although Binaryen supports it, and AS is supposed to allow low-level control over the output Wasm, if I'm not mistaken.

I had edited a simple [Fibonnaci function implementation](https://github.com/AssemblyScript/assemblyscript/issues/162#issuecomment-403110513), into the following:

```ts
const SQRT5 = 2.23606797749979;
const C1 = 0.48121182505960347;
const C2 = -0.48121182505960336;

export function fib(n: f64): f64 {
  if (n <= 0) {
    return 0;
  }

  const temp = f64x2.mul(f64x2(C1, C2), f64x2.splat(n));

  const a = Math.exp( f64x2.extract_lane(temp, 0) );
  const b = Math.exp( f64x2.extract_lane(temp, 1) );

  return Math.round(
    (i32(n) & 1 ? a + b : a - b) * (1.0 / SQRT5)
  );
}
```

Regarding optimization, I checked the wat and noticed that the compiler was emitting two f64 locals, when it could've left the values on the stack, something more like the following:
```
local.get temp
f64x2.extract_lane 0
call Math.exp
local.get temp
f64x2.extract_lane 1
call Math.exp
local.get n
i32/f64
i32.const 1
i32.and
if (param f64 f64) (result f64)
    f64.add
else
    f64.sub
end
f64.const 1.0
global.get SQRT5
f64.div
f64.mul
call Math.round
return 
```

which should provide superior codegen in runtime VMs, while yielding size benefits.

There has been mention of waiting for Multi-value support for multiple things already, such as adding [nullable types](https://github.com/AssemblyScript/assemblyscript/issues/1410), among other things.

Is multi-value support is the plans already?

JairusSW:
Fixes #1633

This PR adds basic support for the [Multi Value Proposal](https://github.com/WebAssembly/multi-value/blob/master/proposals/multi-value/Overview.md). Some of the major changes are:
- Adds `TupleTypeNode` to the AST
- Parser now parses `TupleTypeNode`
- Supports `--enable multi-value` to cli and features
- Instead of a single `returnType: Type`, we now generally accept `returnTypes: Type[]` across the board
- Added an explicit `resolveReturnTypes()` method that essentially walks all types within the tuple and delegates it to `resolveType()`

I need to add some more tests and probably should wait until binaryen is updated before taking this off draft

I did notice that running `wat2wasm` on generated `.wat` *will not* compile because binaryen uses a `tuple` type that is not in the WAT spec. I'm assuming that the output of `--textFile` shouldn't necessarily *need* to be valid WAT?

> I *did* utilize Claude to make an outline plan for implementation which I referenced. I had it trace other features and document what files need to be modified. Really helped with such a large codebase

I intentionally tried to modify as little as possible in this PR while still making it somewhat complete. If you have time while it's still in draft phase, I'd appreciate some basic feedback to make sure I'm going in the right direction here rather than correcting something later on which just requires more time for everyone.

Here's a sort of roadmap for this PR:
- [x] Allow functions to return correct multi-values through `binaryen`
- [ ] Allow tuples to be indexed properly
  - [x] Index with compile time constant
  - [ ] Index at runtime with bounds checking
  - [ ] Support `unchecked()`??
- [ ] Allow tuples to be assigned to locals
  - [x] Function locals
  - [ ] Global locals
- [ ] Allow tuples to be ingested as parameters
- [ ] Allow tuples to be destructured properly

Let me know if there's anything i missed here

### DIFF
--- a/src/diagnosticMessages.json
+++ b/src/diagnosticMessages.json
@@ -52,6 +52,7 @@
   "Initializer, definitive assignment or nullable type expected.": 238,
   "Definitive assignment has no effect on local variables.": 239,
   "Ambiguous operator overload '{0}' (conflicting overloads '{1}' and '{2}').": 240,
+  "An interface or abstract method '{0}' cannot have type parameters.": 241,
 
   "Importing the table disables some indirect call optimizations.": 901,
   "Exporting the table disables some indirect call optimizations.": 902,
--- a/src/program.ts
+++ b/src/program.ts
@@ -2079,6 +2079,13 @@ export class Program extends DiagnosticEmitter {
         }
         case NodeKind.MethodDeclaration: {
           let methodDeclaration = <MethodDeclaration>memberDeclaration;
+          if (methodDeclaration.is(CommonFlags.Abstract) && methodDeclaration.is(CommonFlags.Generic)) {
+            this.error(
+              DiagnosticCode.An_interface_or_abstract_method_0_cannot_have_type_parameters,
+              methodDeclaration.name.range,
+              methodDeclaration.name.text
+            );
+          }
           if (memberDeclaration.isAny(CommonFlags.Get | CommonFlags.Set)) {
             this.initializeProperty(methodDeclaration, element);
           } else {
@@ -2653,6 +2660,13 @@ export class Program extends DiagnosticEmitter {
         }
         case NodeKind.MethodDeclaration: {
           let methodDeclaration = <MethodDeclaration>memberDeclaration;
+          if (methodDeclaration.is(CommonFlags.Generic)) {
+            this.error(
+              DiagnosticCode.An_interface_or_abstract_method_0_cannot_have_type_parameters,
+              methodDeclaration.name.range,
+              methodDeclaration.name.text
+            );
+          }
           if (memberDeclaration.isAny(CommonFlags.Get | CommonFlags.Set)) {
             this.initializeProperty(methodDeclaration, element);
           } else {
--- a/src/resolver.ts
+++ b/src/resolver.ts
@@ -3063,7 +3063,19 @@ export class Resolver extends DiagnosticEmitter {
           let boundPrototype = classInstance.getMember(unboundOverridePrototype.name);
           if (boundPrototype) { // might have errored earlier and wasn't added
             assert(boundPrototype.kind == ElementKind.FunctionPrototype);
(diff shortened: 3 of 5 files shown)


src/types.ts
```diff
@@ -946,6 +963,38 @@ export class Signature {
...
     hasRest: bool = false,
+  ): Signature {
+    if (returnType != Type.void) {
+      return Signature.createInternal(program, parameterTypes, [ returnType ], thisType, requiredParameters, hasRest);
+    }
+    return Signature.createInternal(program, parameterTypes, [], thisType, requiredParameters, hasRest);
+  }
+
+  /** Construct a new signature with one or more return types. */
+  public static createMulti(
```
JairusSW: Is this the right approach? I went with it to avoid breaking changes.
Alternatively, we can switch to a `Type[]` by default with non-multi return types just being `[type]`

MaxGraey:
The PR is quite large and tries to do everything at once. I would break it down into several PRs with thorough test coverage.

1. Support parsing and compiles (without multi-values) for tuples with a bunch of test cases and failure testes
```ts
function pair(a: i32, b: i32): [i32, i32];
function nested(a: i32, b: i32): [i32, [u32, string]];
function arg(pair: [a: i32, b: f64]): void;
function argReadonly(pair: readonly [a: i32, b: f64]): void;
function srgWithoutNames(pair: [i32, f64]): void; // legacy style
```
> ~~codegen should simply fallback to array literals (if all types the same) or to statically generated  for specific arity and signature class `Tuple` which is mimic to Array (ArrayLike) and at least contains "length" property and has read access by index.~~

Actually, since multi-value is now widely available, it might be worth enabling it by default, so we wouldn't have to worry about the tuple fallback. So, at this first stage, we can hide all this under the compiler's experimental flag.

2. Support multi-values infra for backend (Binaryen). Usually it relates to module.ts and compiler.ts. With own tests.
3. Finally, put everything together, enable multi-value support, and start by supporting very simple cases that don’t require complex AA & def-use analysis.

PR #3018, opened 2026-04-22
feat: properly parse and resolve tuples

JairusSW:
This PR adds basic support for the [Tuples](https://www.typescriptlang.org/docs/handbook/basic-types.html#tuple). Roadmap wise, this, along with a few more PRs, will pave the way for the [Multi Value Proposal](https://github.com/WebAssembly/multi-value/blob/master/proposals/multi-value/Overview.md) to be properly implemented.

The goal of this PR is not to add full-fledged support for `Tuples`. Instead, it supports the basic functionality for parsing and resolving them since optimal `Tuple` implementation relies on Multi Value support.

The order of the PRs to implement are:

1. Parse and resolve tuples
2. Hook up Multi Value support to `binaryen`
3. Clean up, test, and enable `multi-value` by default

Speaking of that, all the features here hide behind `--enable multi-value`

So far, this implements:

- [x] Parser
  - [x] Parameters: `function foo(x: [i32, i32]) {}`
  - [x] Return types: `function foo(): [i32, i32] {}`
  - [x] Return values syntax: `return [0, 1]`
  - [x] Type aliases: `type Foo = [i32, i32]`
  - [x] Type names: `[x: f64, y: f64]`
- [x] Compiler
  - [x] Tuple usages error with `not implemented` (for now)


src/program.ts
```diff
@@ -442,6 +442,9 @@ export class Program extends DiagnosticEmitter {
     let nativeFile = new File(this, Source.native);
     this.nativeFile = nativeFile;
     this.filesByName.set(nativeFile.internalName, nativeFile);
+
+    // temporary fix
+    this.parser.options = this.options;
```
MaxGraey: Why was this necessary?

src/parser.ts
```diff
@@ -563,6 +578,42 @@ export class Parser extends DiagnosticEmitter {
         return null;
       }
 
+    // '[' Type (',' Type)* ']'
```
MaxGraey: this should be something like this I guess:
```ts
// '[' ((Type | Identifier ':' Type) (',' (Type | Identifier ':' Type))*)? ']'
```
due to `[]`, `[x: f32, y: f32]`, `[f32, f32]`

src/parser.ts
```diff
@@ -563,6 +578,42 @@ export class Parser extends DiagnosticEmitter {
...
+    } else if (this.options && this.options!.hasFeature(Feature.MultiValue) && token == Token.OpenBracket) {
```
MaxGraey:
```suggestion
    } else if (token == Token.OpenBracket && this.options && this.options!.hasFeature(Feature.MultiValue)) {
```

src/parser.ts
```diff
@@ -563,6 +578,42 @@ export class Parser extends DiagnosticEmitter {
...
+      let elements: TypeNode[] = [];
+      let elementNames: (IdentifierExpression | null)[] = [];
+      let hasElementNames = false;
```
MaxGraey, line 582: Unnecessary var. Just compare with `elementNames.length != 0` later

tests/parser/tuple.ts
```diff
@@ -0,0 +1,25 @@
...
+function tuple8(): [head: i32, tail: [lo: i32, hi: i32]] { return [1, [2, 3]]; }
+
+function func1(a: i32, b: i32): [i32, i32] { return [a, b]; }
+function func2(x: [i32, i32]): [i32, i32] { return x; }
+function func3(x: [i32, [i32, i32]], y: i32): [i32, [i32, i32]] { return x; }
+function func4(x: readonly [i32, string]): [void] { return [void(0)]; }
+function func5(x: readonly [Array<i32>, i32[]]): readonly [i32] { return [x[1].length]; }
+function func6(x: [i32, i32] | null): [i32, i32] | null { return x; }
+function func7(x: readonly [[i32[]], [string]]): readonly [[i32[]], [string]] { return x; }
+function func8(x: [left: i32, right: i32]): [first: i32, second: i32] { return x; }
```
MaxGraey, line 19: Add 
```ts
function func9<T>(x: [T, T, i32]): [T] { return x; }
```

tests/parser/tuple.ts
```diff
@@ -0,0 +1,25 @@
+function tuple1(): [i32, i32] { return [1, 2]; }
```
MaxGraey, line 2: Add also 
```ts
function tuple0(): [] { return []; }
```

tests/parser/tuple.ts
```diff
@@ -0,0 +1,25 @@
...
+function tuple6(): [[i32[]]] { return [[[1, 2]]]; }
+function tuple7(): [x: i32, y: i32] { return [1, 2]; }
+function tuple8(): [head: i32, tail: [lo: i32, hi: i32]] { return [1, [2, 3]]; }
+
+function func1(a: i32, b: i32): [i32, i32] { return [a, b]; }
+function func2(x: [i32, i32]): [i32, i32] { return x; }
+function func3(x: [i32, [i32, i32]], y: i32): [i32, [i32, i32]] { return x; }
+function func4(x: readonly [i32, string]): [void] { return [void(0)]; }
+function func5(x: readonly [Array<i32>, i32[]]): readonly [i32] { return [x[1].length]; }
+function func6(x: [i32, i32] | null): [i32, i32] | null { return x; }
```
MaxGraey, line 17: Will be great to have:
```ts
function func6(x: [string | null, i32] | null): [string | null, i32] | null { return x; }
```

src/parser.ts
```diff
@@ -509,8 +513,19 @@ export class Parser extends DiagnosticEmitter {
...
+    // 'readonly' Type
+    if (token == Token.Readonly) {
+      let innerType = this.parseType(tn, acceptParenthesized, suppressErrors);
+      if (!innerType) return null;
+      type = Node.createNamedType(
+        Node.createSimpleTypeName("Readonly", tn.range(startPos, tn.pos)),
+        [ innerType ],
+        false,
+        tn.range(startPos, tn.pos)
+      );
```
MaxGraey: I think make sense to create helper method for this similar to [createOmittedType](https://github.com/JairusSW/assemblyscript/blob/5375b249cf4e54b8ad934e806079218535b48041/src/ast.ts#L174C10-L174C27) called `careateReadonlyType`

src/program.ts
JairusSW: To hide everything behind `--enablr multi-value`. `parser.ts` doesn't have access to `program.options` so I rigged a fix together

src/parser.ts
JairusSW: Thanks. I forgot to extend the grammar there

src/program.ts
MaxGraey: I think it makes sense to extend Parser's constructor and add `options` as 3rd arg. This might be useful later on

JairusSW: Makes sense. I was initially wary of extending core classes' signatures, but I suppose it's fine

src/parser.ts
JairusSW: I feel that using `elementNames.length != 0` is a bit more hacky...
The reason is that unless a `Map` is used, `n(elementNames) == n(elements)` (where n signifies cardinality of the set).
So, if the current element has a name, we backfill missing indexes with null so that `elementNames.length == i`, then we push the actual name.
Then, if a name has already been added (`elementNames.length != 0`), we need to push `null` for each missing name.

Essentially, this gives us something like:
```ts
if (elementName) {
  while (elementNames.length < elements.length - 1) elementNames.push(null);
  elementNames.push(elementName);
} else if (elementNames.length) {
  elementNames.push(null);
}
// ...
type = Node.createTupleType(...,elementNames.length ? elementNames : null,...)
```

MaxGraey: I see. You don't want to use `Map<{elementName}, {element}>` here. Ok

src/parser.ts
```diff
@@ -118,7 +121,8 @@ export class Parser extends DiagnosticEmitter {
   sources: Source[];
   /** Current overridden module name. */
   currentModuleName: string | null = null;
-
+  /** Compiler options. (workaround) */
```
MaxGraey:
```suggestion
  /** Compiler options. */
```

src/program.ts
JairusSW: I cannot add it to the end of `Parser`'s constructor since it cannot come after an optional argument.
```ts
/** Constructs a new parser. */
constructor(
  diagnostics: DiagnosticMessage[] | null = null,
  sources: Source[] = [],
  options: Options // cannot add required arg after optional arg
) {
  super(diagnostics);
  this.sources = sources;
}
```
We have a few options here:
1. Keep my solution as a temporary fix until `multi-value` support is enabled
2. Add `options: Options` to constructor and possibly break transforms
3. Make `options` nullable and optional

MaxGraey: Hmm, not sure why all these are optional. How critical in terms of refactoring perspective to remove these default values?

If there are many areas for refactoring, just leave it as it was originally.

dcodeIO: Iirc the idea there was that the parser can be used on its own to simply parse something, so it only uses the state it really needs.

JairusSW: Should be rather trivial, though it *may* break a few transforms out there. That being said, it's also quite simple to patch after the breaking change
I still feel like either a temporary workaround or an optional `options` argument are the best paths forward here

JairusSW: @dcodeIO, Exactly. For example, here: https://github.com/JairusSW/json-as/blob/main/transform/src/util.ts#L8
```ts
export class SimpleParser {
  private static get parser(): Parser {
    return new Parser();
  }
  ...
}
```

src/parser.ts
```diff
@@ -563,6 +573,42 @@ export class Parser extends DiagnosticEmitter {
         return null;
       }
 
+    // '[' ((Identifier ':')? Type (',' (Identifier ':')? Type)*)? ']'
+    } else if (token == Token.OpenBracket && this.options.hasFeature(Feature.MultiValue)) {
```
MaxGraey: need to revert `this.options != null` guard before:
```ts
this.options.hasFeature(Feature.MultiValue)
```

JairusSW: Will do after we decide how we want to proceed with the `Parser` constructor

MaxGraey: Yeah, stick with your original version. I didn't realize that all the values in Parser's constructor are optional. Adding a default for the `Options` isn't the best idea either. No matter how you look at it, this will all result in breaking changes.

JairusSW: Sounds good, thanks :+1:

src/program.ts
JairusSW: We're going to stick with the workaround

src/program.ts
```diff
@@ -436,12 +436,15 @@ export class Program extends DiagnosticEmitter {
...
     super(diagnostics);
-    this.module = Module.create(options.stackSize > 0, options.sizeTypeRef);    
+    this.module = Module.create(options.stackSize > 0, options.sizeTypeRef);
     this.parser = new Parser(this.diagnostics, this.sources);
     this.resolver = new Resolver(this);
     let nativeFile = new File(this, Source.native);
     this.nativeFile = nativeFile;
     this.filesByName.set(nativeFile.internalName, nativeFile);
+
+    // temporary workaround. remove after multi-value support is finished
```
MaxGraey:
```suggestion
    // TODO: temporary workaround. remove after multi-value support is finished
```

src/parser.ts
```diff
@@ -118,7 +121,8 @@ export class Parser extends DiagnosticEmitter {
   sources: Source[];
   /** Current overridden module name. */
   currentModuleName: string | null = null;
-
+  /** Compiler options. (temporary. remove after multi-value/tuple support is finished) */
```
MaxGraey:
```suggestion
// TODO: Remove when multi-value feture will enable by default.
/** Compiler options. */
```

MaxGraey (approved):
LGTM

But let someone else take a fresh look at this

tests/compiler/tuple-errors.json
```diff
@@ -0,0 +1,9 @@
+{
+  "asc_flags": [
+    "--enable", "multi-value"
+  ],
+  "stderr": [
+    "AS100: Not implemented: Tuple types",
+    "20 compile error(s)"
```
dcodeIO: Expecting a specific number of errors seems a bit brittle, as it depends on what other code paths, that might be changed independently, do. Would expect that the test instead checks that "AS100: Not implemented: Tuple types" is at the expected line index or so.

tests/parser/tuple.ts.fixture.ts
```diff
@@ -0,0 +1,68 @@
...
+}
+function func2(x: [i32, i32]): [i32, i32] {
+  return x;
+}
+function func3(x: [i32, [i32, i32]], y: i32): [i32, [i32, i32]] {
+  return x;
+}
+function func4(x: Readonly<[i32, string]>): [void] {
+  return [void(0)];
+}
```
MaxGraey: Hmm, that's interesting. @dcodeIO does this valid in AS?
In TS semantically [this is not the same](https://www.typescriptlang.org/play/?#code/GYVwdgxgLglg9mABFApgZygRgBQH0BciASigIYAmCANgJ4A8A2mCALYBGKATgDSIacwwAcwC6APgCUhBgDc4MciMQBvAFCJEnFFBCcks+eWwAGCSIDcqgL6rVoSLATJ0UAEx5CJCtXpNWHHj4oAWFxKUQGJTUNLR09CItrWygaAAcURABRAEcQUio6AEFeACExRABedURsOgAVMWwJSvK6xBQAD1QwcjREQsQAfkRMREJXZs7u3urahqaWxDaplB6+kqGRscQJ6o1h4JAUPe3gfLQUS1UU9P60C84oevau1d7kTiPyiqWrm4yiKMfiQ4mA6mkUHR-nBgM4MJgxJZ-sRXJViNpdGCIVCITC4W5EckIYgAMIACxQEAA1mjCvcuE8cnkCoDeERXGJCRogA)

But I guess it's just AST serialization issue

JairusSW: @MaxGraey, it's not implemented at least. TypeScript does allow it though, so it parses, but doesn't compile.

```ts
function foo(): void {
  return void(0);
}
foo();
```
```
ERROR AS234: Expression does not compile to a value at runtime.
    :
 49 │ return void(0);
    │        ~~~~
    └─ in assembly/test.ts(49,10)

ERROR TS2349: Cannot invoke an expression whose type lacks a call signature. Type 'auto' has no compatible call signatures.
    :
 49 │ return void(0);
    │        ~~~~~~~
    └─ in assembly/test.ts(49,10)

FAILURE 2 compile error(s)
```

dcodeIO: Iirc, the `void` type is sometimes used as a fallback type for where a type errored, and the compiler allows passing `void`s through for this reason instead of producing a bunch of follow-up errors. Without having looked in detail, my expectation would be that we diagnose a user-provided and misplaced `void` type early for this reason, and this snipped seems to be such a case.

JairusSW: Since we're on the topic, `[]` should be essentially treated as an alias for `void`, right?

dcodeIO: That would match Wasm multi-value well and is a good way of thinking about it, though I wonder how straight-forward it is to enforce it accross the existing AS codebase, as it might need updated assumptions.

MaxGraey: > Since we're on the topic, [] should be essentially treated as an alias for void, right?

No, [] is unit type. Btw I'm not sure if we actually have this type semantically. But it's somewhat similar to `null`, only for aggregates.

MaxGraey: I think we can add a TODO / FIXME for now and address this edge case in the next pull request.

src/parser.ts
JairusSW: TypeScript uses a `TypeOperatorNode` internally. Perhaps we'd want to go a similar route instead of using a named type?
I slapped the named type together as a quick solution. might not be the best choice here
```typescript
factory.createTypeOperatorNode(
  ts.SyntaxKind.ReadonlyKeyword,
  factory.createTupleTypeNode([factory.createTypeReferenceNode(
    factory.createIdentifier("i32"),
    undefined
  )])
)
```

https://ts-ast-viewer.com/#code/GYVwdgxgLglg9mABMOcAUAPAXIgTgUwEMATBAGwE9EBtGAZgCYBdAShwDc4ZjEBvAXyA

MaxGraey: Hmm, need to check for we handle `keyof Type`, `typeof Type`. `readonly Type` should follow the same way. 

Update: it seems we only tokenize this but not create in AST

tests/parser/tuple.ts.fixture.ts
JairusSW: I'll keep it in mind for the next PR then

src/parser.ts
JairusSW: We can probably ignore it for now and parse `readonly [i32, i32]` into `[i32, i32]`. In the future, it would be a good idea to provide actual support for `readonly`. That should probably be a separate PR though.

Anyways, for now I'll just ignore `readonly`

src/parser.ts
```diff
@@ -563,6 +568,49 @@ export class Parser extends DiagnosticEmitter {
...
 
+    // 'readonly' Type
+    } else if (token == Token.Readonly) {
+      let innerType = this.parseType(tn, acceptParenthesized, suppressErrors);
+      if (!innerType) return null;
+      type = innerType;
+      type.range.start = startPos;
+
+    // '[' ((Identifier ':')? Type (',' (Identifier ':')? Type)*)? ']'
+    } else if (token == Token.OpenBracket && this.options && this.options!.hasFeature(Feature.MultiValue)) {
```
HerrCai0907, line 579: Actually tuple doesn't and is impossible to rely on mutli-value.
```ts
declare function f(): [i32,i32];
function b() {
  let a = f();
  let b = a;
  b[0] = 1;
  assert(a[0] == 1);
}
```
In this cases, multiple value cannot handle this case since it treat the whole tuple as a value instead of a heap object.

HerrCai0907 (approved):
LGTM.
But I think tuple does not rely on multi-value. Here is my implement of tuple in AS. maybe  you can ref it if you are interesting to implement the whole tuple support. https://github.com/wasm-ecosystem/warpo/pull/269

src/parser.ts
JairusSW: @HerrCai0907, I'll add full support for tuples in the future. Tuples will need to exist in two forms:
1. Tuple becomes `multi value` when returned from function
2. When used as a heap type (eg. `store<[i32, i32]>()`, `Array<[i32, i32]>`), it becomes a heap type

Then, depending on usage, tuples will be lifted and lowered from multi-value to tuple

MaxGraey: We’ve been discussing this for a long time, and it’s clear that the only way to ensure that tuples are properly converted to multivalues is to place a `readonly` attribute before them. Otherwise, we’ll have to analyze the control flow for side effects which required custom pass based on binaryen infra.

yoyo837:
Heads-up: the insertion of `TupleType = 3` into the `NodeKind` enum in this PR shifted all subsequent enum values by +1. This caused a silent break in [json-as](https://github.com/JairusSW/json-as), whose compiled transform output had those values baked in as numeric literals by TypeScript's `const enum`, so `@json`-decorated classes stopped having their serialize/deserialize methods injected at build time.

Tracking issue: [json-as#188](https://github.com/JairusSW/json-as/issues/188)  
Fix: [json-as#189](https://github.com/JairusSW/json-as/pull/189) (switched to runtime `NodeKind` imports instead of hard-coded numbers)

Not a blocker — just flagging the downstream impact for visibility. 👍

yoyo837:
By the way, I just noticed that json-as is also your project! That's a nice coincidence — thank you for all the great work on both AssemblyScript and json-as, really appreciate it. 🙏

JairusSW:
If anyone hits problems with transforms since this PR changes `NodeKind`, this should fix it.

```ts
import * as asc from "assemblyscript/dist/assemblyscript.js";
export const NodeKind = (asc as unknown as { NodeKind: Record<string, number> }).NodeKind;
```

PR #3020, opened 2026-04-24
feat: add support for tail calls

JairusSW:
This PR adds basic support for the [Tail Call Proposal](https://github.com/WebAssembly/tail-call/blob/master/proposals/tail-call/Overview.md) behind the `--enable tail-calls` flag.

Currently, it only detects direct `return call()` statements. I'll go ahead and support `call_indirect` along with some other things later.

Roadmap:
- [x] `return call()` statements
- [x] `return call_indirect(...)` statements
- [ ] Tail positions inside control flow (if/else, switch, ternary) where each terminating branch returns a call
- [ ] Tail positions through expression wrappers that preserve tail semantics (e.g. block-wrapped return values)
- [ ] Indirect forms (`const x = call(...); return x;`) when no semantic changes can be proven

> Oh forgot to say, I used a LLM to scaffold initial tests. Typically I use like `foo1`, `foo2`, etc.. so this time you get nice names lol. That being said, it did miss a *lot* of tests i'd like to add, so I'll be adding them. The current tests are sufficient for now though

### DIFF
--- a/src/compiler.ts
+++ b/src/compiler.ts
@@ -6152,8 +6152,7 @@ export class Compiler extends DiagnosticEmitter {
             Constraints.ConvImplicit | Constraints.IsThis
           );
         }
-        if (!this.canTailReturn(constraints, contextualType, functionInstance.signature.returnType))
-          constraints &= ~Constraints.WillReturn;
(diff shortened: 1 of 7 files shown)


JairusSW:
~~oops forgot to fetch upstream~~

KKiiim:
Implemented in Binaryen v133.
- https://github.com/WebAssembly/binaryen/pull/9103

PR #3021, opened 2026-04-27
feat: multi-value returns

JairusSW:
This PR adds basic support for the [Multi-Value Proposal](https://github.com/WebAssembly/multi-value). For a tuple type to be considered for `multi-value`, it must be explicitly marked with `readonly` first.

Related: #1633 #1425 

So far, this adds basic compiler/binaryen support for outputting multi-value returns, but does not support things like indexing yet

I'm thinking that the next PR should add support type checking for tuples/multi-values, support indexing, and add an SROA pass to identify array literals vs tuples. After that, full support for tuples probably based off of warpo's implementation

### DIFF
--- a/src/compiler.ts
+++ b/src/compiler.ts
@@ -7104,36 +7104,48 @@ export class Compiler extends DiagnosticEmitter {
       }
     }
 
+    let allTrivial = (getSideEffects(functionArg, module.ref) & SideEffects.WritesGlobal) == 0;
+    if (operands && allTrivial) {
+      for (let i = 0; i < numOperands; ++i) {
+        if (!module.isTrivialExpression(operands[i])) {
+          allTrivial = false;
+          break;
+        }
+      }
+    }
+
+    let stmts = new Array<ExpressionRef>();
+    let sizeTypeRef = this.options.sizeTypeRef;
+
+    if (!allTrivial) {
+      let functionArgLocal = this.currentFlow.getTempLocal(this.options.usizeType);
+      let functionArgSetExpr = module.local_set(functionArgLocal.index, functionArg, true);
+      stmts.push(functionArgSetExpr);
+      functionArg = module.local_get(functionArgLocal.index, sizeTypeRef);
+    }
+
     // We might be calling a varargs stub here, even if all operands have been
     // provided, so we must set `argumentsLength` in any case. Inject setting it
     // into the index argument, which becomes executed last after any operands.
     let argumentsLength = this.ensureArgumentsLength();
-    let sizeTypeRef = this.options.sizeTypeRef;
-    if (getSideEffects(functionArg, module.ref) & SideEffects.WritesGlobal) {
-      let flow = this.currentFlow;
-      let temp = flow.getTempLocal(this.options.usizeType);
-      let tempIndex = temp.index;
-      functionArg = module.block(null, [
-        module.local_set(tempIndex, functionArg, true), // Function
-        module.global_set(argumentsLength, module.i32(numArguments)),
-        module.local_get(tempIndex, sizeTypeRef)
-      ], sizeTypeRef);
-    } else { // simplify
-      functionArg = module.block(null, [
-        module.global_set(argumentsLength, module.i32(numArguments)),
-        functionArg
-      ], sizeTypeRef);
-    }
+    
+   
+    let functionArgWithVararg = module.block(null, [
+      module.global_set(argumentsLength, module.i32(numArguments)),
+      functionArg
+    ], sizeTypeRef);
+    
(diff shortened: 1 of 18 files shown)


JairusSW:
Shall I push support for compile-time indexing of multi-values? I have it stashed and it's only around +70 loc

MaxGraey:
The more I think about tuples in AS the more I lean toward the idea that semantically, when we declare `[T, U, ...]` it should be **`readonly` by default** and maybe even implicitly so. Ofc that would be a deviate from TypeScript but we are already stricter than TS in many ways. The reason is that tuples should always be passed by value and never by reference. That is how it is done in all languages, and it greatly simplifies lowering to WebAssembly.

Unfortunately, the relevant JS proposal has been abandoned and archived: https://github.com/tc39/proposal-record-tuple

But in our case we still have the hint from the types! It would be more correct to always use `readonly [T, U...]` but having to write and read that every time and then run into compiler errors saying that tuples must be `readonly` kw would just be annoying and confusing

JairusSW:
@MaxGraey, i'd prefer that tuples are treated as `readonly` no matter how you declare them. Most languages implement them that way anyways. It would be a breaking change when compared to TypeScript though, so maybe we can get more opinions...
@dcodeIO @CountBleck thoughts?

dcodeIO:
To me it seems that `[i32, string]` is rather a tuple object on the heap, naturally with reassignable properties and the ability to be stored in a property of some other object by reference, whereas `readonly [i32, string]` is a good counterpart for multi-value, with open design space on how these behave when assigned to a property of some object.

Generally, I can roughly imagine how codegen might look when mutable `[i32, string]` was a multi-value tuple, but differences in, say, assignment to object properties makes me wonder if having a distinction here at first is the better MVP, be it only because initial implementation is cleaner and doesn't need to touch object representation and whatnot in one go.

CountBleck:
My intuition of `[T, U]` is similar to @dcodeIO's, where "a tuple object on the heap" would behave equivalently to how a WasmGC `struct` declared outside of a `rec` group would behave (structurally typed, passed by reference, mutable). Meanwhile, since pass-by-value would be indistinguishable (besides in performance) from pass-by-reference for `readonly [T, U]`, multi-value (Binaryen tuples) make sense for it.

MaxGraey:
In that case, we don’t gain any benefits from multi-values and are forced to postpone the use of tuples until full integration with the Wasm GC.

JairusSW:
Multi value would help with memory pressure. We can lift and lower to and from `readonly` aka multi-valie as needed

MaxGraey:
> Multi value would help with memory pressure. We can lift and lower to and from readonly aka multi-valie as needed

What about the ABI though? `readonly` / non-`readonly` for tuples in exported functions would have different layouts depending on the modifier, which breaks ABI stability.

Also, most users are not set `readonly` due to verbosity while computing side effects is fairly expensive and makes everything pretty complicated (even with middle IR). I can give a lot more arguments for why, with these "pass by ref" semantics we would effectively almost always end up with a heap object.

MaxGraey:
At the same time, I decided to check how often real-world TS code actually tries to mutate tuples... and I couldn’t find such code sample. It’s just not idiomatic in the typescript world. Maybe you know some examples?

JairusSW:
> What about the ABI though? readonly / non-readonly for tuples in exported functions would have different layouts depending on the modifier, which breaks ABI stability.

Dirty fix, but we could lift `multi-values` to `Tuple<T>` or vice versa. We'd just need to enforce one I suppose. But yeah, it does complicate things

> Maybe you know some examples?

I don't believe I've ever mutated a tuple and I've used TS a lot

MaxGraey:
> where "a tuple object on the heap" would behave equivalently to how a WasmGC struct declared outside of a rec group would behave (structurally typed, passed by reference, mutable)

WasmGC is hybrid of nominal subtyping + structural (hierarchical) equality. But In TS we do not need structural equality for tuples. In addition `struct` type has an RTT which I consider to be an overhead for simple tuples

JairusSW:
Should I go ahead and keep implementing this? It seems like most of us agree on `readonly [T, U]` -> `multi-value` and `[T, U]` -> `Tuple<T, U>` and we can settle on how heap tuples are actually represented later

> most users are not set readonly due to verbosity while computing side effects

I'm not too worried about that. If someone cares enough about perf to want multi-value, slapping `readonly` in front of the tuple is a small ask and it keeps things explicit rather than relying on the optimizer to guess. Besides, `readonly` makes sense that it'd be represented as a multi-value anyways

JairusSW:
oops i trashed my commits accidentally

PR #3025, opened 2026-06-09
feat: switch grisu2 float-to-string algorithm to hybrid of xjb & zmij algorithms

### LINKED ISSUE #3012: [Task] Replace Grisu2 with Zmij for float-to-string conversion
AssemblyScript currently uses a pretty old `Grisu2`-based algorithm for `f32`/`f64` to string conversion. Will be great to rewrite this using [`Zmij`](https://github.com/vitaut/zmij), which is a newer shortest-roundtrip algorithm and should be [much faster](https://github.com/vitaut/zmij?tab=readme-ov-file#performance) and probably more precise in some rare cases. SIMD should be optional.

Current implementation: https://github.com/AssemblyScript/assemblyscript/blob/main/std/assembly/util/number.ts#L729
Tests: https://github.com/AssemblyScript/assemblyscript/blob/main/tests/compiler/std/string.ts#L736

C++ Reference: https://github.com/vitaut/zmij
Rust port: https://github.com/dtolnay/zmij

JairusSW:
Fixes #3012.

* Switch to [xjb-as](https://github.com/JairusSW/xjb-as) which is an improvement over [zmij-as](https://github.com/JairusSW/zmij-as)
* Comply to the [ECMA262 Specification](https://tc39.es/ecma262/multipage/ecmascript-data-types-and-values.html#sec-numeric-types-number-tostring) with the exception of a trailing `.0`

> Note: I wrote [xjb-as](https://github.com/JairusSW/xjb-as) with claude assisting me for an initial port. I then optimized by hand. Everything is carefully checked and passes over a trillion fuzz cases compared directly against `Number::parse()` in V8. It also reaches 100% code coverage and tests every edge case both I and claude could think of.

It's good quality stuff though, and I'm quite confident in that.

Earlier main-vs-PR formatter measurements, before the exact-integer string path (10m iterations per cycle):

| Randomish input bucket | `main` | PR | Speedup |
| --- | ---: | ---: | ---: |
| `f64` buffered, ns/value | 42.2 | 20.9 | 2.02× |
| `f32` buffered, ns/value | 36.0 | 17.8 | 2.02× |
| `f64` string-returning, ns/value | 79.4 | 59.6 | 1.33× |
| `f32` string-returning, ns/value | 72.0 | 55.7 | 1.29× |

Earlier generated code size, before the exact-integer string path (a `main` function calling `dtoa/ftoa`):
| Program uses | `main` | PR | Δ |
| --- | ---: | ---: | ---: |
| `f64` only | 4,120 B | 6,357 B | +2,237 B |
| `f32` only | 4,118 B | 5,192 B | +1,074 B |
| Both | 4,342 B | 9,431 B | +5,089 B |

Most of this is the formatter which is now ECMA262 compliant, so, it's more complicated.
The actual tables are `910 bytes` (before) vs `872 bytes` (this pr)

> P.S. Sorry about the diff. Most of it is generated WAT fixtures.


PaperPrototype:
Just curious here, why the exception of a trailing .0?

JairusSW:
> Just curious here, why the exception of a trailing .0?

Max and Dan made that decision so that floats are easily identifiable when converting them to a string. For example, you know that the string `"1.0"` is a result of `f64.toString()` while the string `"1"` is likely a result of `i32.toString()`. Slight deviation from ECMA262-spec, but it's not awful

MaxGraey:
Can you rename xjb.ts to dtoa.ts and in comments mentioned about implementation based on xjb and zmij? `xjb` filename is very exotic and doesn't explains exactly what this file does

std/assembly/util/dtoa.ts
```diff
@@ -0,0 +1,907 @@
...
+@inline export function mulhi64(a: u64, b: u64): u64 {
+  const a0 = a & 0xffffffff, a1 = a >> 32;
+  const b0 = b & 0xffffffff, b1 = b >> 32;
+  const w0 = a0 * b0;
+  const t = a1 * b0 + (w0 >> 32);
+  let w1 = t & 0xffffffff;
+  const w2 = t >> 32;
+  w1 = a0 * b1 + w1;
+  return a1 * b1 + w2 + (w1 >> 32);
+}
```
MaxGraey, line 21: Also please use `const` only for leterals, for the rest locals which binded with expr use `let`

std/assembly/util/xjb.ts
```diff
@@ -0,0 +1,907 @@
+import { POWERS10 } from "./string";
+import { DIGITS } from "./number";
+
+// @ts-expect-error: may exist
+export const HAS_SIMD: bool = isDefined(XJB_SIMD) ? <bool>XJB_SIMD : (isDefined(ASC_FEATURE_SIMD) && ASC_FEATURE_SIMD);
```
MaxGraey: What is `XJB_SIMD` env? We should check only ASC_FEATURE_SIMD here. If simd is slow just ged rid of that vector-based implemantaion

std/assembly/util/dtoa.ts
```diff
@@ -0,0 +1,901 @@
...
+  return (binExp * log10_2_sig) >> log10_2_exp;
+}
+
+// Shift that keeps a fixed 128-bit fractional part after scaling by 10**dec_exp.
+// @ts-expect-error: decorator
+@inline export function computeExpShift(binExp: i32, decExp: i32): i32 {
+  const log2_pow10_sig = 217707, log2_pow10_exp = 16;
+  const pow10BinExp = (-decExp * log2_pow10_sig) >> log2_pow10_exp;
+  return binExp + pow10BinExp + 1;
+}
```
MaxGraey, line 49: Also about style. Here how it should be:
```suggestion
@inline export function computeExpShift(binExp: i32, decExp: i32): i32 {
  const LOG2_POW10_EXP = 16;
  const LOG2_POW10_SIGNIFICAND = 0x3526B;
  let pow10BinExp = (-decExp * LOG2_POW10_SIGNIFICAND) >> LOG2_POW10_EXP;
  return binExp + pow10BinExp + 1;
}
```

Also `0x3526B` (217707) is binary represenataion. Add comment what does it mean. What fp represantaion of this magic number?

std/assembly/util/dtoa.ts
```diff
@@ -0,0 +1,901 @@
...
+
+  // 192-bit product major * minor as c2:c1:c0, keep the top 128 bits (c2:c1).
+  const h1 = mulhi64(hLo, m);
+  const c0 = hLo * m;
+  const c1 = h1 + hHi * m;
+  const c2 = u64(c1 < h1) + mulhi64(hHi, m);
+
+  let hi: u64, lo: u64;
+  if ((c2 >> 63) != 0) {
+    hi = c2; lo = c1;
```
MaxGraey:
```suggestion
    lo = c1;
    hi = c2;
```

std/assembly/util/dtoa.ts
```diff
@@ -0,0 +1,901 @@
...
+  } else {
+    hi = (c2 << 1) | (c1 >> 63);
+    lo = (c1 << 1) | (c0 >> 63);
```
MaxGraey:
```suggestion
    lo = (c1 << 1) | (c0 >> 63);
    hi = (c2 << 1) | (c1 >> 63);
```

std/assembly/util/dtoa.ts
```diff
@@ -0,0 +1,901 @@
...
+const DIV10K_EXP = 40;
+const DIV10K_SIG: u64 = ((<u64>1) << DIV10K_EXP) / 10000 + 1;
+const NEG10K: u64 = ((<u64>1) << 32) - 10000;
+const DIV100_EXP = 19;
+const DIV100_SIG: u64 = (1 << DIV100_EXP) / 100 + 1;
+const NEG100: u64 = (1 << 16) - 100;
+const DIV10_EXP = 10;
+const DIV10_SIG: u64 = (1 << DIV10_EXP) / 10 + 1;
+const NEG10: u64 = (1 << 8) - 10;
+export const ZEROS: u64 = 0x3030303030303030;
```
MaxGraey:
```suggestion
const DIV10_EXP = 10;
const DIV10_SIG: u64 = (1 << DIV10_EXP) / 10 + 1;
const NEG10: u64 = (1 << 8) - 10;

const DIV100_EXP = 19;
const DIV100_SIG: u64 = (1 << DIV100_EXP) / 100 + 1;
const NEG100: u64 = (1 << 16) - 100;

const DIV10K_EXP = 40;
const DIV10K_SIG: u64 = ((<u64>1) << DIV10K_EXP) / 10000 + 1;
const NEG10K: u64 = ((<u64>1) << 32) - 10000;

export const ZEROS: u64 = 0x3030303030303030;
```

std/assembly/util/dtoa.ts
```diff
@@ -0,0 +1,901 @@
...
+
+let gBcd: u64 = 0;
+let gBcdLen: i32 = 0;
```
MaxGraey, line 153: gBcd -> gBcdVal
?

std/assembly/util/dtoa.ts
```diff
@@ -0,0 +1,901 @@
...
+
+// value < 1e8 -> 8 packed BCD digits (SWAR: divide-by-const reciprocals).
+function toBcd8(abcdefgh: u64): void {
+  const abcd_efgh = abcdefgh + NEG10K * ((abcdefgh * DIV10K_SIG) >> DIV10K_EXP);
+  const ab_cd_ef_gh =
+    abcd_efgh + NEG100 * (((abcd_efgh * DIV100_SIG) >> DIV100_EXP) & 0x7f0000007f);
+  const a_b_c_d_e_f_g_h =
+    ab_cd_ef_gh + NEG10 * (((ab_cd_ef_gh * DIV10_SIG) >> DIV10_EXP) & 0xf000f000f000f);
+  const bcd = bswap<u64>(a_b_c_d_e_f_g_h);
```
MaxGraey: All this names is quite verbose and strange. You can use comments for display byte/num relations

std/assembly/util/dtoa.ts
```diff
@@ -0,0 +1,901 @@
...
+  gBcd = bcd;
+  gBcdLen = <i32>((70 - clz<u64>((bcd << 1) | 1)) / 8);
+}
+
+// to_digits<64> result: two u64 of ASCII digits + significant digit count.
+export let gDigHi: u64 = 0;
+export let gDigLo: u64 = 0;
+export let gDigNum: i32 = 0;
```
MaxGraey: You have a lot of global vars. Does we need all of them? And can we declare all of necessary vars in one place?

std/assembly/util/dtoa.ts
```diff
@@ -0,0 +1,901 @@
...
+  const t = i32x4.shr_u(mulhiU16(y, div100), 3);
+  const z = i32x4.add(y, i32x4.mul(neg100v, t));
+  return i16x8.add(z, i16x8.mul(neg10v, mulhiU16(z, div10v)));
+}
+
+// Pack the low 32 bits of each i64 lane into adjacent i32 lanes 0,1 (zero the
+// rest), so a single-input swizzle replaces a two-input shuffle.
+// @ts-expect-error: decorator
+@inline function packLo32(v: v128): v128 {
+  return v128.swizzle(v, i8x16(0, 1, 2, 3, 8, 9, 10, 11, -128, -128, -128, -128, -128, -128, -128, -128));
```
MaxGraey:
```suggestion
  return v128.swizzle(v, i8x16(
    0, 1, 2, 3, 8, 9, 10, 11, 
    -128, -128, -128, -128, -128, -128, -128, -128
  ));
```

tests/compiler/bindings/raw.release.wat
```diff
@@ -54,36 +54,41 @@
...
+ (data $24 (i32.const 3580) ",")
+ (data $24.1 (i32.const 3592) "\02\00\00\00\1a\00\00\00~\00l\00i\00b\00/\00a\00r\00r\00a\00y\00.\00t\00s")
+ (data $25 (i32.const 3628) "|")
+ (data $25.1 (i32.const 3640) "\02\00\00\00^\00\00\00E\00l\00e\00m\00e\00n\00t\00 \00t\00y\00p\00e\00 \00m\00u\00s\00t\00 \00b\00e\00 \00n\00u\00l\00l\00a\00b\00l\00e\00 \00i\00f\00 \00a\00r\00r\00a\00y\00 \00i\00s\00 \00h\00o\00l\00e\00y")
+ (data $26 (i32.const 3756) "<")
+ (data $26.1 (i32.const 3768) "\02\00\00\00*\00\00\00O\00b\00j\00e\00c\00t\00 \00a\00l\00r\00e\00a\00d\00y\00 \00p\00i\00n\00n\00e\00d")
+ (data $27 (i32.const 3820) "<")
+ (data $27.1 (i32.const 3832) "\02\00\00\00(\00\00\00O\00b\00j\00e\00c\00t\00 \00i\00s\00 \00n\00o\00t\00 \00p\00i\00n\00n\00e\00d")
+ (data $28 (i32.const 3888) "\10\00\00\00 \00\00\00 \00\00\00 ")
+ (data $28.1 (i32.const 3912) "\81\08\00\00\01\19\00\00\01\02\00\00$\t\00\00\a4\00\00\00$\n\00\00\02\t\00\00\02A\00\00\00\00\00\00A\00\00\00 ")
```
dcodeIO: Wondering if the baseline amount of static memory can be reduced here by computing the parts of it that aren't strictly necessary to have as LUTs?

std/assembly/util/dtoa.ts
```diff
@@ -0,0 +1,901 @@
...
+  const y = i64x2.add(x, sub);
+
+  const bcd = toBcd4x4(y);
+
+  const mask = i8x16.bitmask(i8x16.gt_s(bcd, i8x16.splat(0)));
+  gDigNum = 16 - ctz(mask); // mask is never 0 (significand >= 1)
+
+  const ascii = v128.or(
+    v128.swizzle(bcd, i8x16(15, 14, 13, 12, 11, 10, 9, 8, 7, 6, 5, 4, 3, 2, 1, 0)),
+    i8x16.splat(0x30));
```
MaxGraey:
```suggestion
  const REV_ORDER = i8x16(15, 14, 13, 12, 11, 10, 9, 8, 7, 6, 5, 4, 3, 2, 1, 0);
  const ascii = v128.or(
    v128.swizzle(bcd, REV_ORDER),
    i8x16.splat(0x30)
  );
```

std/assembly/util/dtoa.ts
```diff
@@ -0,0 +1,901 @@
...
+  if (lo == 0) {
+    gDigHi = hiBcd + ZEROS;
+    gDigLo = ZEROS;
+    gDigNum = hiLen;
+    return;
+  }
+  toBcd8(lo);
+  gDigHi = hiBcd + ZEROS;
+  gDigLo = gBcd + ZEROS;
+  gDigNum = 8 + gBcdLen;
```
MaxGraey: rename gDigNum -> gDigits

std/assembly/util/dtoa.ts
```diff
@@ -0,0 +1,901 @@
...
+}
+
+// @ts-expect-error: decorator
+@inline export function toDigits64(value: u64): void {
+  if (ASC_FEATURE_SIMD) return toDigits64Simd(value);
+  toDigits64Swar(value);
+}
```
MaxGraey, line 270:
```suggestion
@inline export function toDigits64(value: u64): void {
  if (ASC_FEATURE_SIMD) {
    toDigits64Simd(value);
  } else {
    toDigits64Swar(value);
  }
}
```

std/assembly/util/dtoa.ts
```diff
@@ -0,0 +1,901 @@
...
+    store<u16>(base, <u16>(ascii >> 56), 14);
+  }
+}
+
+// ECMAScript spellings for the non-finite cases.
+// @ts-expect-error: decorator
+@inline export function writeNaN(buf: usize): usize {
+  store<u16>(buf, 0x4e); store<u16>(buf, 0x61, 2); store<u16>(buf, 0x4e, 4); // "NaN"
+  return buf + 6;
+}
```
MaxGraey, line 314:
```suggestion
@inline export function writeNaN(buf: usize): usize {
  store<u16>(buf, 0x4e, 0); // 'N'
  store<u16>(buf, 0x61, 2); // 'a'
  store<u16>(buf, 0x4e, 4); // 'N'
  return buf + 6;
}
```

std/assembly/util/dtoa.ts
```diff
@@ -0,0 +1,901 @@
...
+
+// @ts-expect-error: decorator
+@inline export function writeInfinity(buf: usize, neg: bool): usize {
+  if (neg) { store<u16>(buf, 0x2d); buf += 2; }
```
MaxGraey: Don't use single lines like that
```suggestion
  if (neg) { 
    store<u16>(buf, 0x2d); 
    buf += 2; 
  }
```

std/assembly/util/dtoa.ts
```diff
@@ -0,0 +1,913 @@
...
+// SIMD version of toDigits64: builds all 16 ASCII digits in one pass.
+// @ts-expect-error: decorator
+@inline function toDigits64Simd(value: u64): void {
+  const hi = value / 100000000;
+  const lo = value - hi * 100000000;
+
+  const x = i64x2.replace_lane(i64x2.splat(lo), 1, hi);
+
+  const q = i64x2.shr_u(
+    i64x2.extmul_low_i32x4_u(packLo32(x), i32x4.splat(<i32>DIV10K_SIG)), DIV10K_EXP);
```
MaxGraey:
```suggestion
  const q = i64x2.shr_u(
    i64x2.extmul_low_i32x4_u(packLo32(x), i32x4.splat(<i32>DIV10K_SIG)), 
    DIV10K_EXP
  );
```

std/assembly/util/dtoa.ts
```diff
@@ -0,0 +1,920 @@
...
+    return;
+  }
+
+  const decExp = ((rawExp - DOUBLE_EXP_OFFSET) * 78913) >> 18;
+  const powExp = -decExp - 1;
+  const h = q + ((powExp * 217707) >> 16);
+  const shift = h + 1 + EXTRA_SHIFT;
+
+  loadPow10Xjb64(powExp);
+  const pHi = gPow10Hi, pLo = gPow10Lo;
```
MaxGraey:
```suggestion
  let pHi = gPow10Hi;
  let pLo = gPow10Lo;
```

std/assembly/util/dtoa.ts
```diff
@@ -0,0 +1,920 @@
...
+  const y = c << shift;
+
+  const a = mulhi64(pHi, y);
+  const plo64 = pHi * y;
+  const lo = plo64 + mulhi64(pLo, y);
+  const p_hi = a + u64(lo < plo64);
```
MaxGraey: Very inconsistent names `plo64` instead `pLo64`, `p_hi` instead `pHi64`

std/assembly/util/dtoa.ts
```diff
@@ -0,0 +1,920 @@
...
+
+    const integral = pow10Hi >> (11 - h);
+    const halfUlp = pow10Hi >> (-h);
+    const dotOne = pow10Hi << (53 + h);
+
+    let one = ((((dotOne >> (53 + h)) * 5) + (((<u64>1) << (9 - h)))) >> (10 - h));
+    one = ((((dotOne >> 54) * 5) & 0x1ff) > ((halfUlp >> 55) * 5))
+      ? ((((dotOne >> 54) * 5) >> 9) + 1)
+      : one;
+    one = dotOne == ((<u64>1) << 62) ? 2 : one;
```
MaxGraey: A lot of exotic constants. Most of them exp or significand mask + 1, or - 1. So better to hoit it to local consts and define some meaningful names

std/assembly/util/dtoa.ts
```diff
@@ -0,0 +1,920 @@
...
+@inline function writeUInt16(buf: usize, value: u64): usize {
+  const len = decimalLen16(value);
+  let p = buf + (<usize>len << 1);
+  let v = value;
+  while (v >= 100) {
+    const q = v / 100;
+    p -= 4;
+    store<u32>(p, load<u32>(DIGITS + (<usize>(v - q * 100) << alignof<u32>())));
+    v = q;
+  }
```
MaxGraey, line 584: this loop can be unrolled

std/assembly/util/dtoa.ts
```diff
@@ -0,0 +1,920 @@
...
+    if (v < 1000000) return v < 100000 ? 5 : 6;
+    return v < 10000000 ? 7 : 8;
+  }
+  if (v < 1000000000000) {
+    if (v < 10000000000) return v < 1000000000 ? 9 : 10;
+    return v < 100000000000 ? 11 : 12;
+  }
+  if (v < 100000000000000) return v < 10000000000000 ? 13 : 14;
+  return 15;
+}
```
MaxGraey, line 567: See similar apreach which slightly more optimal for perf for branch leafs:

https://github.com/AssemblyScript/assemblyscript/blob/main/std/assembly/util/number.ts#L133

std/assembly/util/dtoa.ts
```diff
@@ -0,0 +1,920 @@
...
+    store<u32>(p, load<u32>(DIGITS + (<usize>(v - q * 100) << alignof<u32>())));
+    v = q;
+  }
+  if (v >= 10) {
+    store<u32>(buf, load<u32>(DIGITS + (<usize>v << alignof<u32>())));
+  } else {
+    store<u16>(buf, 0x30 + <u32>v);
+  }
+  return buf + (<usize>len << 1);
+}
```
MaxGraey, line 591: Also did you compared this approach with original?
https://github.com/AssemblyScript/assemblyscript/blob/main/std/assembly/util/number.ts#L221

I think original whould be faster

std/assembly/util/dtoa.ts
```diff
@@ -0,0 +1,920 @@
...
+  const threshold: u64 = 1000000000000000;
+
+  // is_normal: 1 <= bin_exp <= 2046
+  const isNormal = <u32>(binExp - 1) < <u32>2046;
+  if (!isNormal) {
+    if (binExp != 0) {
+      if (binSig != 0) return writeNaN(buf);
+      return writeInfinity(buf, neg);
+    }
+    if (binSig == 0) { store<u16>(buf, 0x30); return buf + 2; } // +/-0 -> "0"
```
MaxGraey:
```suggestion
    // +/-0 -> "0"
    if (binSig == 0) { 
      store<u16>(buf, 0x30); 
      return buf + 2; 
    }
```

Btw I guess it should be "0.0"?

std/assembly/util/dtoa.ts
```diff
@@ -0,0 +1,920 @@
...
+}
+
+// to_digits<32>: a single u64 of 8 ASCII digits (value < 1e8).
+// @ts-expect-error: decorator
+@inline export function toDigits32(value: u64): void {
+  if (ASC_FEATURE_SIMD) return toDigits32Simd(value);
+  toBcd8(value);
+  gDigHi = gBcd + ZEROS;
+  gDigNum = gBcdLen;
+}
```
MaxGraey, line 771:
```suggestion
@inline export function toDigits32(value: u64): void {
  if (ASC_FEATURE_SIMD) {
    toDigits32Simd(value);
  } else {
    toBcd8(value);
    gDigHi = gBcd + ZEROS;
    gDigNum = gBcdLen;
  }
}
```

std/assembly/util/dtoa.ts
```diff
@@ -0,0 +1,920 @@
...
+    dHi = (dHi >> 8) | (lastDigitChar << 56);
+  }
+
+  // decExp >= 8: integer rendered as significant digits then trailing zeros.
+  if (decExp >= 8) {
+    putBlock8(buf, dHi);
+    if (hasExtraDigit) store<u16>(buf + 16, <u32>lastDigitChar);
+    const sig = 8 + i32(hasExtraDigit);
+    const endByte = buf + ((decExp + 1) << 1);
+    for (let z = buf + (sig << 1); z < endByte; z += 16) putBlock8(z, ZEROS);
```
MaxGraey:
```suggestion
    for (let z = buf + (sig << 1); z < endByte; z += 16) {
      putBlock8(z, ZEROS);
    }
```

std/assembly/util/dtoa.ts
```diff
@@ -0,0 +1,920 @@
...
+    return endByte;
+  }
+
+  const n = numDigits + i32(hasExtraDigit);
+  const endPos = decExp >= 0 ? (n > decExp + 1 ? n + 1 : decExp + 1) : n;
```
MaxGraey: nit: double ternary

MaxGraey:
I think I’ve roughly point style direction. There are still quite a lot of tweaks & nits to make. Ping me when you’re done, and I’ll give it another pass.

std/assembly/util/dtoa.ts
```diff
@@ -0,0 +1,942 @@
...
+  const LOG_10_2_EXP = 20;
+  return (binExp * LOG10_2_SIG) >> LOG_10_2_EXP;
+}
+
+// Shift that keeps a fixed 128-bit fractional part after scaling by 10**dec_exp.
+// @ts-ignore: decorator
+@inline export function computeExpShift(binExp: i32, decExp: i32): i32 {
+  const LOG2_POW10_SIG = 217707;
+  const LOG2_POW10_EXP = 16;
+  const pow10BinExp = (-decExp * LOG2_POW10_SIG) >> LOG2_POW10_EXP;
```
MaxGraey:
```suggestion
  let pow10BinExp = (-decExp * LOG2_POW10_SIG) >> LOG2_POW10_EXP;
```

std/assembly/util/dtoa.ts
```diff
@@ -0,0 +1,942 @@
...
+  // 192-bit product major * minor as c2:c1:c0, keep the top 128 bits (c2:c1).
+  let h1 = mulhi64(hLo, m);
+  let c0 = hLo * m;
+  let c1 = h1 + hHi * m;
+  let c2 = u64(c1 < h1) + mulhi64(hHi, m);
+
+  let hi: u64, lo: u64;
+  if ((c2 >> 63) != 0) {
+    hi = c2;
+    lo = c1;
```
MaxGraey:
```suggestion
    lo = c1;
    hi = c2;
```

std/assembly/util/number.ts
```diff
@@ -460,301 +409,37 @@ export function itoa64(value: i64, radix: i32): String {
...
+  let end = buffer + (<usize>len << 1);
+  while (p < end) {
+    let c = <i32>load<u16>(p);
+    if ((c < CharCode._0 || c > CharCode._9) && c != CharCode.MINUS) return len;
+    p += 2;
   }
+  store<u16>(end, CharCode.DOT);
+  store<u16>(end, CharCode._0, 2);
+  return len + 2;
 }
```
MaxGraey, line 430: Looks like post-process stage. Can we optimize this and do in-place on main processing stage?

std/assembly/util/number.ts
```diff
@@ -460,301 +409,37 @@ export function itoa64(value: i64, radix: i32): String {
...
-  if (!isFinite(value)) {
-    if (isNaN(value)) return "NaN";
-    return select<String>("-Infinity", "Infinity", value < 0);
+  let len: u32;
+  if (isFloat<T>() && sizeof<T>() == 4) {
+    // @ts-ignore: type
+    len = ftoa_buffered_single(dtoa_buf, <f32>value);
+  } else {
+    // @ts-ignore: type
+    len = dtoa_buffered_double(dtoa_buf, <f64>value);
```
MaxGraey, line 439: ftoa & dtoa already have hint about type so simplify names to:

`ftoa_buffered` and `dtoa_buffered`

std/assembly/util/dtoa.ts
```diff
@@ -0,0 +1,976 @@
...
+
+export let gPow10Hi: u64 = 0;
+export let gPow10Lo: u64 = 0;
+
+// Compact pow10 (Dougall Johnson's method, ported from vitaut/zmij)
+// 10**i = top-128-bits(major[(i+10)/28] * minor[(i+10)%28]) minus a per-power round-down bit.
+
+// 28 normalized exact powers 10**0..10**27 - the within-stride minor factors.
+// @ts-ignore: decorator
+@lazy const POW10_MINOR = memory.data<u64>([
```
MaxGraey, line 59: Plz move all constans and lut tables on top

std/assembly/util/dtoa.ts
```diff
@@ -0,0 +1,976 @@
...
+
+const DIV100_EXP = 19;
+const DIV100_SIG: u64 = (1 << DIV100_EXP) / 100 + 1;
+const NEG100: u64 = (1 << 16) - 100;
+
+const DIV10K_EXP = 40;
+const DIV10K_SIG: u64 = ((<u64>1) << DIV10K_EXP) / 10000 + 1;
+const NEG10K: u64 = ((<u64>1) << 32) - 10000;
+
+export const ZEROS: u64 = 0x3030303030303030;
```
MaxGraey, line 150:
```suggestion
export const BCD_ZEROS: u64 = 0x3030303030303030;
```

std/assembly/util/dtoa.ts
```diff
@@ -0,0 +1,976 @@
...
+  // four 2-digit groups, one per 16-bit lane: [12][34][56][78]
+  let pairs =
+    quads + NEG100 * (((quads * DIV100_SIG) >> DIV100_EXP) & 0x7f0000007f);
+  // eight 1-digit groups, one per byte: [1][2][3][4][5][6][7][8]
+  let singles =
+    pairs + NEG10 * (((pairs * DIV10_SIG) >> DIV10_EXP) & 0xf000f000f000f);
+  // bswap to big-endian so the most-significant digit lands in the high byte
+  let bcd = bswap<u64>(singles);
+  gBcd = bcd;
+  gBcdLen = <i32>((70 - clz<u64>((bcd << 1) | 1)) / 8);
```
MaxGraey, line 169: What `70` meaning here? Can you make local const with meaningful name or comment?

std/assembly/util/dtoa.ts
```diff
@@ -0,0 +1,976 @@
...
+  
+  let hi = value / 100000000;
+  let lo = value - hi * 100000000;
+
+  let x = i64x2.replace_lane(i64x2.splat(lo), 1, hi);
+
+  let q = i64x2.shr_u(
+    i64x2.extmul_low_i32x4_u(packLo32(x),
+    i32x4.splat(<i32>DIV10K_SIG)), DIV10K_EXP
+  );
```
MaxGraey, line 221:
```suggestion
  let q = i64x2.shr_u(
    i64x2.extmul_low_i32x4_u(packLo32(x), i32x4.splat(<i32>DIV10K_SIG)),
    DIV10K_EXP
  );
```

std/assembly/util/dtoa.ts
```diff
@@ -0,0 +1,976 @@
+import { POWERS10 } from "./string";
+import { DIGITS, MAX_DOUBLE_LENGTH } from "./number";
+
+// High 64 bits of the 128-bit product x * y. Matches umul128.
+// @ts-ignore: decorator
+@inline export function mulhi64(a: u64, b: u64): u64 {
```
MaxGraey, line 6: mulhi64 -> umul64hi

std/assembly/util/dtoa.ts
```diff
@@ -0,0 +1,976 @@
...
+  let w2 = t >> 32
+  
+  w1 = a0 * b1 + w1
+  
+  return a1 * b1 + w2 + (w1 >> 32)
+}
+
+// Returns (x * y + c) >> 64.
+// @ts-ignore: decorator
+@inline export function umul128AddHi64(x: u64, y: u64, c: u64): u64 {
```
MaxGraey, line 25: umul64hiCarry(a: u64, b: u64, carry: u64): u64

std/assembly/util/dtoa.ts
```diff
@@ -0,0 +1,976 @@
...
+// floor(log10(2**bin_exp)). (The f64 path only ever needs the regular form; the
+// irregular 3/4 variant lives in ftoa.ts's own copy.)
+// @ts-ignore: decorator
+@inline export function computeDecExp(binExp: i32): i32 {
+  return (binExp * LOG10_2_SIGNIFICAND) >> LOG10_2_EXP;
+}
+
+// Shift that keeps a fixed 128-bit fractional part after scaling by 10**dec_exp.
+// @ts-ignore: decorator
+@inline export function computeExpShift(binExp: i32, decExp: i32): i32 {
```
MaxGraey, line 46:
```suggestion
@inline export function exponentShift(binExp: i32, decExp: i32): i32 {
```

std/assembly/util/dtoa.ts
```diff
@@ -0,0 +1,976 @@
...
+// Fixed-point log significands shared by the dec-exp / binary-exp estimates below.
+const LOG10_2_SIGNIFICAND = 0x4D105; // ~(log10(2) * 2**20)
+const LOG10_2_EXP = 20;
+const LOG2_POW10_SIGNIFICAND = 0x3526B; // ~(log2(10) * 2**16)
+const LOG2_POW10_EXP = 16;
+
+// floor(log10(2**bin_exp)). (The f64 path only ever needs the regular form; the
+// irregular 3/4 variant lives in ftoa.ts's own copy.)
+// @ts-ignore: decorator
+@inline export function computeDecExp(binExp: i32): i32 {
```
MaxGraey, line 40:
```suggestion
@inline export function toDecExponent(binExp: i32): i32 {
```

std/assembly/util/dtoa.ts
```diff
@@ -0,0 +1,976 @@
...
+  0x0c000000, 0x82c81380, 0x5e4ce01f, 0xd730f60f, 0x0000001b,
+  0x00000000, 0xcdf7fffc, 0x6e8201d8, 0x40cd3fd1, 0xdb642501,
+  0x00000d0d, 0x14042400, 0x53713840, 0x11781db4, 0x00000000,
+]);
+
+// Rebuild gPow10Hi/gPow10Lo for table index i in [0,617]: the top 128 bits of
+// major[(i+10)/28] * minor[(i+10)%28], normalized left if the top bit is clear,
+// then the per-power fixup bit subtracted off the low limb.
+// @ts-ignore: decorator
+@inline function computePow10(i: i32): void {
```
MaxGraey, line 100:
```suggestion
@inline function pow10(exp: i32): u64 {
   ...
   return lo;
}
```

This also reduce one global var. Also do the same for the rest similar functions which store low u64 part in glob instead return it

std/assembly/util/dtoa.ts
```diff
@@ -0,0 +1,976 @@
...
+@inline export function loadPow10Xjb64(power: i32): void {
+  computePow10(power + 293);
+  gPow10Lo += u64(power < 0);
+}
+
+// @ts-ignore: decorator
+@inline export function loadPow10HiXjb64(power: i32): u64 {
+  computePow10(power + 293);
+  return gPow10Hi;
+}
```
MaxGraey, line 136: These two functions are too small and trivial. Inline into call sites and remove them

std/assembly/util/dtoa.ts
```diff
@@ -0,0 +1,976 @@
...
+    v128.store(base, i16x8.extend_low_i8x16_u(i64x2.splat(ascii)));
+  } else {
+    store<u16>(base, <u16>(ascii & 0xff));
+    store<u16>(base, <u16>((ascii >> 8) & 0xff), 2);
+    store<u16>(base, <u16>((ascii >> 16) & 0xff), 4);
+    store<u16>(base, <u16>((ascii >> 24) & 0xff), 6);
+    store<u16>(base, <u16>((ascii >> 32) & 0xff), 8);
+    store<u16>(base, <u16>((ascii >> 40) & 0xff), 10);
+    store<u16>(base, <u16>((ascii >> 48) & 0xff), 12);
+    store<u16>(base, <u16>(ascii >> 56), 14);
```
MaxGraey, line 303: It's ASCII -> UTF16. It can be more efficient. It's typical interleave with zero idiom:
```suggestion
    let lo = ascii & 0xFFFFFFFF;
    let hi = ascii >> 32;
    
    lo = (lo | (lo << 16)) & 0x0000FFFF0000FFFF;
    hi = (hi | (hi << 16)) & 0x0000FFFF0000FFFF;
    lo = (lo | (lo <<  8)) & 0x00FF00FF00FF00FF;
    hi = (hi | (hi <<  8)) & 0x00FF00FF00FF00FF;
    
    store<u64>(base, lo, 0);
    store<u64>(base, hi, 8);
```

std/assembly/util/dtoa.ts
```diff
@@ -0,0 +1,976 @@
...
+export const BIASED_HALF: u64 = ((<u64>1) << 63) + 6;
+export const DOUBLE_MAX_DIGITS10 = 17;
+// Fixed notation when decExp (= decimal-point position - 1) is in [-6, 20].
+export const MIN_FIXED_DEC_EXP = -6;
+export const MAX_FIXED_DEC_EXP = 20;
+
+// Eight packed ASCII digits in a u64 -> 8 UTF-16 code units (16 bytes) at
+// `p + off`. SIMD zero-extends the bytes to u16 lanes in one store.
+// @ts-ignore: decorator
+@inline export function putBlock8(p: usize, ascii: u64, off: usize = 0): void {
```
MaxGraey, line 291: Also `putBlock8` is not perfect name. I guess better use `writeUnpacked8` or `writeDeinterleved64toUtf16`  or something like this

std/assembly/util/dtoa.ts
```diff
@@ -0,0 +1,976 @@
...
+@inline export function writeFixed(
+  buf: usize,
+  start: usize,
+  decExp: i32,
+  hasLastDigit: bool,
+): usize {
+  if (decExp < 0) putBlock8(start, ZEROS);
+  let lastDigitChar = <u64>(0x30 + (hasLastDigit ? gLastDigit : 0));
+  let numDigits = hasLastDigit ? 16 : gDigits - 1;
+  let dHi = gDigHi, dLo = gDigLo;
```
MaxGraey, line 338:
```suggestion
  let dHi = gDigHi;
  let dLo = gDigLo;
```

std/assembly/util/dtoa.ts
```diff
@@ -0,0 +1,976 @@
...
+    // shifted right by k chars (fractional part at [k+1,...]), then drop '.' at k.
+    let k = decExp + 1;
+    let d16: u64 = lastDigitChar;
+    let s = k << 3;
+    let fHi: u64, fLo: u64;
+    if (s < 64) {
+      fHi = (dHi >> s) | (dLo << (64 - s));
+      fLo = (dLo >> s) | (d16 << (64 - s));
+    } else if (s == 64) {
+      fHi = dLo; fLo = d16;
```
MaxGraey, line 376:
```suggestion
      fHi = dLo; 
      fLo = d16;
```

std/assembly/util/dtoa.ts
```diff
@@ -0,0 +1,976 @@
...
+  buf -= usize(buf - 2 == start + 2) << 1; // drop a trailing point
+  return writeExponent(buf, decExp);
+}
+
+// Writes "e +/- d" / "e +/- dd" / "e +/- ddd" exponent.
+// @ts-ignore: decorator
+@inline export function writeExponent(buf: usize, decExp: i32): usize {
+  let m = decExp >> 31; // all-ones if decExp < 0
+  store<u16>(buf, 0x65); // 'e'
+  store<u16>(buf, 0x2b + (m & 2), 2); // '+' / '-' branchlessly
```
MaxGraey, line 424: We have `CharCode` enum for such char literals
```ts
import { CharCode } from "./string";
```

std/assembly/util/dtoa.ts
```diff
@@ -0,0 +1,976 @@
...
+    gSig = <i64>integral;
+    gLastDigit = 0;
+    gHasLastDigit = false;
+  } else {
+    gSig = <i64>integral;
+    gLastDigit = <i32>one;
+    gHasLastDigit = true;
+  }
+  gExp = decExp;
+}
```
MaxGraey, line 467: What if we inline this function on all 3 call sites? Does this allow us to eliminate or reduce the number of such glob vars? It really bothers me that there are so many of glob vars even though most of them could be stored locally and passed as arguments. Also at least one could be returned from the function. And use extra glob vars as fallback apprach when we need to return a multi-value

std/assembly/util/dtoa.ts
```diff
@@ -0,0 +1,976 @@
...
+
+    let integral = pow10Hi >> (11 - h);
+    let halfUlp = pow10Hi >> (-h);
+    let dotOne = pow10Hi << (53 + h);
+
+    let one = ((((dotOne >> (53 + h)) * 5) + (((<u64>1) << (9 - h)))) >> (10 - h));
+    one = ((((dotOne >> 54) * 5) & 0x1ff) > ((halfUlp >> 55) * 5))
+      ? ((((dotOne >> 54) * 5) >> 9) + 1)
+      : one;
+    one = dotOne == ((<u64>1) << 62) ? 2 : one;
```
MaxGraey, line 490: How about invert sign for powExp at first place then is will be much more optimal:

```suggestion
    let decExp = (q * LOG10_2_SIGNIFICAND - 131072) >> LOG10_2_EXP; // 131072 = 2**17 rounding bias
    let powExp = decExp + 1;
    let h = ((powExp * LOG2_POW10_SIGNIFICAND) >> LOG2_POW10_EXP) - q;
    
    let pow10Hi = computePow10(293 - powExp);
    gPow10Hi = pow10Hi;

    let integral = pow10Hi >> 11 + h;
    let halfUlp = pow10Hi >> h;
    let dotOne = pow10Hi << 53 - h;

    let one = ((((dotOne >> (53 - h)) * 5) + (((<u64>1) << (9 + h)))) >> (10 - h));
    one = ((((dotOne >> 54) * 5) & 0x1ff) > ((halfUlp >> 55) * 5))
      ? ((((dotOne >> 54) * 5) >> 9) + 1)
      : one;
    one = dotOne == ((<u64>1) << 62) ? 2 : one;
```

MaxGraey: Also I recommend `((dotOne >> 54) * 5)` and `((dotOne >> (53 - h)) * 5)` move out from expr as locals

std/assembly/util/dtoa.ts
MaxGraey: Also as I mentioned we can remove `gBcdLen` and return it as laocal var from here function and leave only `gBcd` for glob var.

MaxGraey, line 169: Also I found a bug. It should be:

```ts
  gBcdLen = <i32>((70 - clz<u64>((singles << 1) | 1)) / 8);
```

| value | fixed gBcdLen | wrong (current) gBcdLen |
|------:|----------:|-----------------:|
| 1 | 1 | 8 |
| 78 | 2 | 8 |
| 1234 | 4 | 8 |
| 10 | 2 | 7 |
| 100 | 3 | 6 |
| 10000 | 5 | 4 |

MaxGraey:
Also plz add extra tests which cover new implementation-specific edge cases (I anready found a bug) and uncomment existing test cases which failed with grisu2

PR #3029, opened 2026-06-13
Fix CLI metadata startup loading in WebContainers

### LINKED ISSUE #3027: TypeError: WebAssembly.compile expects a Response object in WebContainer environments
### Bug description

`asc --version` no longer works inside web containers (e.g. StackBlitz).

### Steps to reproduce

I upgraded the package.json for AssemblyScript to 0.28.17 using the playground found here https://stackblitz.com/edit/learning-assemblyscript-01-t6meir?file=package.json

Then, in the terminal,
`npm upgrade`
`asc --version`

And I got the following error:

```
[TypeError: An argument must be provided, which must be a Response or Promise<Response> object] {
  message: "Failed to execute 'compile' on 'WebAssembly': An argument must be provided, which must be a Response or Promise<Response> object"
}
```

0.27.x is fine, but as soon as you go 0.28.x, you hit this error.

### AssemblyScript version

0.28.x

culpen90:
Fixes #3027.

## Summary
- defer compiler and Binaryen initialization until after metadata-only CLI exits
- avoid a static Binaryen import in the built asc bundle
- install an `instantiateStreaming` fallback while loading compiler startup dependencies, so environments with broken streaming support can still compile
- add browser test coverage for the version path, bundle shape, and first compile with broken `instantiateStreaming`

## Validation
- corepack npm run build
- corepack npm run lint
- corepack npm run test:browser
- corepack npm run test:cli
- corepack npm run test:transform:esm
- corepack npm run test:transform:cjs

### DIFF
--- a/cli/index.js
+++ b/cli/index.js
@@ -33,18 +33,20 @@ import { utf8 } from "../util/text.js";
 import * as optionsUtil from "../util/options.js";
 import * as generated from "./index.generated.js";
 
-import binaryen from "../lib/binaryen.js";
-import * as assemblyscriptJS from "assemblyscript";
+let assemblyscript = null;
+let binaryen = null;
+let binaryenModule = "../lib/binaryen.js";
 
-// Use the TS->JS variant by default
-let assemblyscript = assemblyscriptJS;
+async function loadAssemblyScript(wasmPath) {
+  if (assemblyscript) return assemblyscript;
+  return assemblyscript = wasmPath == null
+    ? await import("assemblyscript")
+    : await import(new URL(wasmPath, url.pathToFileURL(process.cwd() + "/")));
+}
 
-// Use the AS->Wasm variant as an option (experimental)
-const wasmPos = process.argv.indexOf("--wasm");
-if (~wasmPos) {
-  const wasmPath = String(process.argv[wasmPos + 1]);
-  process.argv.splice(wasmPos, 2);
-  assemblyscript = await import(new URL(wasmPath, url.pathToFileURL(process.cwd() + "/")));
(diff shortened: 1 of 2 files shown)


CountBleck:
Wouldn't the underlying problem still be present when you try to compile something?

Also, from my testing, I think the issue is that StackBlitz doesn't properly support `WebAssembly.instantiateStreaming()`, at least in the form `WebAssembly.instantiateStreaming(fetch("data:application/wasm;base64,..."))`.

culpen90:
@CountBleck The first version only avoided loading the compiler/Binaryen for metadata-only exits, so yes, an actual compile could still hit the same broken startup wasm path.

I pushed a follow-up in `0ab3f456` that covers the compile path too. The CLI now installs a temporary `WebAssembly.instantiateStreaming` fallback while loading the compiler startup dependencies and Binaryen: it still tries native streaming first, but if that fails it falls back to `WebAssembly.instantiate(await response.arrayBuffer(), imports)`.

I also added browser regression coverage that patches `WebAssembly.instantiateStreaming` to throw and verifies `asc.compileString(...)` still succeeds, in addition to the existing `asc --version` check.

MaxGraey:
> install an instantiateStreaming fallback while loading compiler startup dependencies, so environments with broken streaming support can still compile

Hmm, I'm not sure we should provide compile streaming fallback from our side. `importWithInstantiateStreamingFallback` add extra async abstructions for compile streaming for most of browsers which support streaming.

It's much better to have the environment load the polyfill if it really required this by itself. For example use this one: https://www.npmjs.com/package/wasm-instantiate-streaming

PR #3033, opened 2026-07-08
fix: optimize dayOfWeek & dateFromEpoch routines in date

MaxGraey:
```ts
floorDiv(year, 400)  ->  floorDiv(floorDiv(year, 100), 4) 
```

Since arithmetic shift `year >> 2` is equal to `floorDiv(year, 4)`:

```ts
century = floorDiv(year, 100)
floorDiv(century, 4)  ->  century >> 2
```

And full simplifcation:

```ts
floorDiv(year, 4) - floorDiv(year, 100) + floorDiv(year, 400)
```
 -> 
 ```ts
(year >> 2) - century + (century >> 2)
```

Also added couple of tests

### DIFF
--- a/std/assembly/date.ts
+++ b/std/assembly/date.ts
@@ -62,7 +62,7 @@ export class Date {
       let timeString: string;
       dateString = dateTimeString.substring(0, posT);
       timeString = dateTimeString.substring(posT + 1);
-      
+
       // might end with an offset ("Z", "+05:30", "-08:00", etc.)
       for (let i = timeString.length - 1; i >= 0; i--) {
         let c = timeString.charCodeAt(i);
@@ -82,8 +82,8 @@ export class Date {
           } else {
             let offsetHours = i32.parse(timeString.substring(i + 1));
             offsetMs = offsetHours * MILLIS_PER_HOUR;
-          }    
-    
+          }
+
           if (c == 45) offsetMs = -offsetMs; // negative offset
           timeString = timeString.substring(0, i);
           break;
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: std/assembly/date.ts [hunk 1]
```diff
@@ -361,7 +361,8 @@ function dayOfWeek(year: i32, month: i32, day: i32): i32 {
   const tab = memory.data<u8>([0, 3, 2, 5, 0, 3, 5, 1, 4, 6, 2, 4]);
 
   year -= i32(month < 3);
-  year += floorDiv(year, 4) - floorDiv(year, 100) + floorDiv(year, YEARS_PER_EPOCH);
+  const century = floorDiv(year >> 2, 100 >> 2);
+  year += (year >> 2) + (century >> 2) - century;
   month = <i32>load<u8>(tab + month - 1);
```
CountBleck: Maybe re-order to match what the original algorithm looked like. Also, you can use the offset to shave a few stray bytes (not like it really matters).

```suggestion
  year += (year >> 2) - century + (century >> 2);
  month = <i32>load<u8>(month, tab - 1);
```

CountBleck (approved):

std/assembly/date.ts
MaxGraey: Yeah, but `<i32>load<u8>(month, tab - 1);` may be dangerous. Better to leave this opt to binaryen

CountBleck: There wasn't any validation on `month` in this function before, so this isn't any less unsafe. Also, every caller of `dayOfWeek` passes in `Date#month`, which is private and calculated via `dateFromEpoch`.

That being said, it's a microoptimization so I don't particularly care either way.

MaxGraey: I mean `<i32>load<u8>(month, tab - 1)` instead `<i32>load<u8>(tab + month - 1)` can catch oob in certain conditions due to `tab` choosed by compiler and `imm-offset` in wasm by spec doesn't take into account duting bounds checking. For example:
```wasm
;; pre-alloc 1 page

i32.const 65535
i32.load8_u offset=0   ;; ok

i32.const 65535
i32.load8_u offset=1   ;; trap
```

CountBleck: That doesn't seem worse...

PR #3039, opened 2026-08-20
fix: template literal OOM error caused by child elements not visible to GC in staticArray

KKiiim:
**Root cause:**

When AssemblyScript compiles a template literal for string concatenation, the compiler generates a static `StaticArray<string>` as an intermediate buffer. This static array resides in the data segment with its ITCMS GC header `gcInfo` set to `3` (transparent/pinned color), but the object is **neither linked into the pinSpace linked list nor visited by `__visit_globals`** (which explicitly skips static GC objects). As a result, the GC has no way to trace the dynamic child strings written into the array via `__uset`, and may incorrectly collect strings that are still in use — leading to `rtSize` corruption, memory corruption, and eventual trap crashes.


KKiiim:
@MaxGraey Hi, could you help to take a look. Thanks

